PR #1, opened 2018-01-02
Write all disk image contents before swapping disk images

hbekel:
Fixes data loss when a disk image was swapped via the disk swap feature before all contents have been written back to the filesystem. 

Before mounting the new disk image, poll() is called repeatedly until the disk is no longer marked dirty. Also uses proper SubsysCommands to perform the mount.


GideonZ:
Thanks, sounds good!  Will look into this, as soon as the horror with the
new website has passed.. :)

Best regards,
Gideon

PR #3, opened 2018-01-13
Version 2 of the Ultimate Sidplayer implemented

WilfredC64:
A new version of the Sidplayer has been implemented.

Changes:

-	rewritten player
-	added an advanced player for displaying clock, song length, time bar and system info
-	added keyboard support for sub tune selection and fast forward
-	player and advanced player are made as separate modules so they can be used by external software
-	song length is now shown and can be retrieved when SID collection contains SSL files
-	the player now plays all SIDs
-	many improvements and fixes

sidcrt.tas is renamed to sidcrt.asm. The location is also changed. The sidcrt.asm file should be assembled with the latest 64tass and cannot be assembled with the currently used 64tass.

The sidcrt has now its own makefile. The sidcrt.bin will be created in the target folder of the sidcrt. The firmware should make use of the makefile in the sidcrt folder and get the sidcrt.bin from the target folder.


WilfredC64:
I've added one more change so that the system info is now always shown.

PR #4, opened 2018-01-15
fixed song length display and added runstop key support

WilfredC64:
Fixed songlength display. Songlengths are now read from $b000.
Added runstop-key support to make it possible to return to ultimate menu.


WilfredC64:
I've compacted the cartridge size a bit to make it possible to add code for handling run-stop key.

PR #6, opened 2018-01-18
Fix display glitch in C64 and cartridge settings menu

hbekel:
When adding the new REU preload options, a piece of code was overlooked that
adjusts the PHI timing values at runtime for either u2plus or the u64.

This caused the option for the ultimate audio to use 16ns and 32ns as values
rather than enable/disable, and the phi timing options to use the wrong
values as well.

This commits corrects this and similar bugs in the future by looking up
CFG_C64_TIMING by id instead of index.

### DIFF
--- a/software/io/c64/c64.cc
+++ b/software/io/c64/c64.cc
@@ -145,8 +145,15 @@ C64::C64()
     flash = get_flash();
 
     if (getFpgaCapabilities() & CAPAB_ULTIMATE2PLUS) {
-        c64_config[7].items = timing2; // hack!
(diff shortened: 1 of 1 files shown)


GideonZ:
Hi Henning!
Thanks for this important fix. However, when I see my original code, I realize this is not the way to do it. I need to fix it using a compile switch rather.
Regards,
Gideon

hbekel:
Yes, I figured that the "hack!" comments were still applicable, so I left them there ;) At least this will prevent others from making a similar mistake until it is properly fixed.

PR #10, opened 2018-02-06
V3.1d 180119 markus c64

markusC64:
I've announced the changes to you, so there is no need to describe them here a 2nd time.


software/io/c64/c64_subsys.cc
```diff
@@ -192,6 +204,83 @@ int C64_Subsys :: executeCommand(SubsysCommand *cmd)
             }
         }
         break;
+	case MENU_C64_SAVEMODULE:
```
thomasf: I think these are accidental CR's added?

markusC64:
Not really. According to khexedit, that file has MSDOS File Endings on several lines (have not checked all of them), so my editor decided to use such line endings for the new lines, too.

PR #21, opened 2018-03-18
Fix Reboot C64: Kernal Replacement

markusC64:
The Code we decided to enable again did not handle Kernal Replacement correctly... Here is the fix.


GideonZ:
Thanks, Markus!  I have pulled this request.

GideonZ:
Hi Markus,

Thanks for the update. I have pulled this request.

I have looked into rewriting the piece of code in filetype_crt.cc to handle
kernal images. When I saw that you introduced two possible kernal base
addresses, I realized that 16K kernals will actually never work correctly.
The cassette "sense" line will *also* switch between two 8K banks of mirror
RAM, which is not intended.  I don't see an easy fix for this.

Best regards,
Gideon

PR #23, opened 2018-04-02
Fast Reset for C64 & Improved D64 filesystem

markusC64:
This patch adds a fast reset option to the Ultimate 64 (internal and custom Kernals) and for the 1541 U2(+) (custom kernals only).

Note that for some patched kernals fast reset is not compatible. This should be detected automatically and fast reset should get disabled.

The secons patch in this pull request changes the d64 file system so copy files from a d64 to outside of a d64 is working much better - and supportings geos files, too.


markusC64:
Rebased to lateest U64 firmware and tested succesfully on U64 with firmware 1.02.

markusC64:
Well, you have done the 8-voices-change in branch "issues/usbtiming". Since we need it in master for COMAL 80 PAKMA to fir in FPGA, I cherry-picked it. And integreated it fully with another commit.

xlar54:
Was this only for the ftp filesystem or is this the code that runs the vfs ?  Ive made some local changes to this on my machine but they only appears in an ftp session, not when using the virtual drive and CDing into a D64 file.

PR #46, opened 2018-11-24
Multiple inprovements

markusC64:
* Fix large commands over port 64
* Fix menu entry
* Access contents of D64 via ftp
* Loading of kernal/dos on U64
* Hopefully this improves compatibility for U2(+) connected to U64
* 3rd alt. ROM /  2nd alt. ROM
* Reading of flash
* Fix writing to kernal via TCP socket 64
* Switching video mode via keyboard when turning on U64


Grrrolf:
Hey Markus!

Quick question:
Will the "Fix large commands over port 64" (also) fix the SOCKET_CMD_MOUNT_IMG and SOCKET_CMD_RUN_IMG commands? These are broken in U2+ firmware 3.3.

Rolf

markusC64:
Have you looked at the source? https://github.com/GideonZ/1541ultimate/pull/46/commits/9331013073b4456c75f8e338b34d5e3f4fe8dc80

The patch is supposed to fix that commands: 
[...]
if ( (cmd == SOCKET_CMD_MOUNT_IMG) || (cmd == SOCKET_CMD_RUN_IMG))
[...]

markusC64:
The defaukt is to list each file twice. But with special user names, that can be changed to other behavior: IIRRC:
D64:  See file as directory
D642:  See file as one image file
aby other name: Listing twice

that is ment for clients incompatible with listing tice.

markusC64:
Well, if the keypress at boot is not recognised, nothing bad happens. All is working asbefore the merge.

Is the keypress is deteced (it did so in all test done by Marty and me), the user can change current output if the monitor is incampatibel to current settings. No bad if in rare situations he has to do it twice.

gintronic007:
the function of choosing the home directory doesn´t work with the 1541 U2 since 3.3. By Pressing the middle button you always get in the same Directory. with 3.2 it worked pretty well. 
Middle Button -> F2 -> User Interface Settings -> Home Directory = no function

Issue #50, opened 2018-12-16
Virtual drive - going into D64 images giving zero for all file sizes


xlar54:
Ive tracked this down to iec_channel.h:

`   int read_dir_entry(void)

    {

        if(dir_index >= dir_last) {

            buffer[2] = 9999 & 255;

            buffer[3] = 9999 >> 8;

            memcpy(&buffer[4], "BLOCKS FREE.             \0\0\0", 28);

            last_byte = 31;

            pointer = 0;

            prefetch_max = 31;

            prefetch = 0;

            return 0;

        }

        printf("Dir index = %d %s\n", dir_index, (*interface->iecNames)[dir_index]);

        FileInfo *info = (*interface->dirlist)[dir_index];

        //printf("info = %p\n", info);

        uint32_t size = 0;

        uint32_t size2 = 0;

        if(info) {

            size = info->size;

        }

        size /= 254;  <-----------------------------this line

        if(size > 9999)

            size = 9999;

        size2 = size;`

When that line is as-is, VFS file sizes are correct, but when you cd into a D64 image, they report zero for all files.  Commenting that line out, gives proper file sizes inside D64 images, but incorrect sizes in the regular VFS.  Catch-22 or am i reading this wrong?

xlar54:
this seems to help, but i dont want to submit a pull request because im not exactly sure how this should work:

if (size > 254)
            size /= 254;

xlar54:
also, should be able to do away with the 9999 BLOCKS FREE as well from inside a disk image.

GideonZ:
Dear xlar,
Thank you very much for your contributions. Yes, please fork and issue a pull request. This is the standard GIT way of working. It also allows me to review the changes and maybe give some pointers when needed. Some changes may go against an established design pattern, and reviews are a great mechanism to achieve maintainable code and avoid quick hacks in the main stream code. (I am not claiming that I have not added any hacks myself.. ;-) )

GideonZ:
> this seems to fix the issue, but i dont want to submit a pull request because im not exactly sure how this should work:
> 
> if (size > 254)
> size /= 254;

As Markus pointed out, this won't work. This kind of code is in the category "hacks". :-)

xlar54:
Yep, no problem.  Will try to come up with a better solution.

PR #51, opened 2018-12-16
Issue #50 - Display correct disk image file sizes inside VFS,  VSCode to gitignore

xlar54:


xlar54:
I use VSCode (excellent editor).  Added a specific file to gitignore for it.  Also, and this may be rejected, but I added DEL and REL files to the remote ftp d64 image.  I can understand why its not really relevant, so can remove if desired.   I actually didnt understand that code at the time, but now becoming clearer.

xlar54:
> Please do not commit functional changes and formatting changes together.

Yes, sorry.  I think some of my temporary changes moved some things around and I should have reverted everything.  

> As discussed, the if(size >254) size /= 254 is rejected; this is a hack that is based on assumptions.

Completely understood.  Just wanted to get the concept in front of you guys so that perhaps a better solution could be implemented.  I think I understand why it was done that way, but as markus64 said - there are bound to be files nested in d64 and d81 images larger than 255 blocks.

xlar54:
Change abandoned for now.

PR #70, opened 2019-04-07
Sidplayer bug fix and pause feature implemented

WilfredC64:


GideonZ:
Big thanks again to Wilfred Bos!

PR #77, opened 2019-05-04
Ability to turn off and on drives via dos command interface

xlar54:
Bart van Leeuwen on FB requested this.  Tested on U64, seems ok.  Let me know if anything im missing or unintended consequences.


xlar54:
test program here:  https://github.com/xlar54/ultimateii-dos-lib/tree/master/target

u-shell.prg  

driveon will power on the drive A, driveoff will turn it off.  drivestat will tell if its on or off
(I did not add a same for drive b)

GideonZ:
Simple code, nice. Because only a register is set/read directly this probably works correctly. Conceptually, it should be executed through a system command call, so that it is executed in the right thread context.

I think this doesn't belong in the DOS target, but in the CONTROL target. Maybe you could move it?

xlar54:
I was thinking about that too.  Yeah, let me see about that and will update the PR

PR #79, opened 2019-05-06
Colour virtual printer

rgc2000:
Hi Gideon,

This update brings colour to the existing virtual printer. The user is still able to run the printer in B&W mode as before. Printing with colour enabled is slower and produce larger files even if there is no colour in the page (bitmap depth is 8 in colour, and only 2 in B&W).

Typical time to generate the same test page :

- B&W on 1541 Ultimate II : 6s
- Color on 1541 Ultimate II : 20s
- B&W on 1541 Ultimate II+ : 30s
- Color on 1541 Ultimate II+ : 90s

This shows the impact of choosing colour output and the better speed of the Microblaze processor on 1541 Ultimate II (not +). I don't own a Ultimate 64, so I can't tell how fast the virtual printer is on this platform.

Please note that printer output order preference has changed in menu. Check your printer settings after a firmware upgrade.

Printer documentation has been updated.

Regards,
René


xlar54:
nice!!

GideonZ:
Thank you very much for this wonderful contribution!!

PR #82, opened 2019-05-08
Plugins

xlar54:
This contains the simple plugins architecture (and apparently my previous PR as well).  Attached is sample plugins to be save to /<root>/ultimate-plugins

[ultimate-plugins.zip](https://github.com/GideonZ/1541ultimate/files/3156469/ultimate-plugins.zip)

Note - this is still just a demonstration of what could be done.  Im totally fine if youd rather go in a different direction with this, or nix the idea completely.


xlar54:
also, if you do like the idea, code will be added to return back to the ultimate menus via the control target upon hitting the runstop key.  didnt want to get too deep in that until i know your thoughts

xlar54:
PR closed for this code due to load address issue.  Preference to make it more dynamic

PR #84, opened 2019-05-12
Implemented MUS cartridge and SID player fixes.

WilfredC64:
Implemented MUS cartridge for playing MUS and stereo MUS (STR) files. Also MUS files inside SID files can be played.
Some bugs were fixed in the SID player. Some SIDs could not be played in the previous version because the SID header could not be read properly when the SID header was allocated at the same address as the cartridge.


GideonZ:
Excellent addition!

PR #87, opened 2019-05-21
fixed getting SID info for 8580 model

### LINKED ISSUE #86: SID Player: wrong SID selected on SidFX
In configurations where SID#1 is a 8580 and SID#2 is 6581, the 6581 is selected for all tunes that specify a SID model.

The bug is caused by getRequiredSidInfo in detectionwrapper.asm: for 8580 it returns $02 instead of $00.

dabadab:
Fixes #86


GideonZ:
Wilfred might still make some changes to this.

PR #89, opened 2019-05-26
Added support for toggling screen on/off

WilfredC64:


GideonZ:
Thank you for all your wonderful contributions!

PR #101, opened 2019-07-21
soft mount d81 images via IEC

xlar54:
This is a means to allow users to "mount" D81 files in the UI.  It calls the CD command of the IEC interface to change directly to the inner D81 file.  If accepted, Ill do the same for D71 files.

Only issue is that mounting in the menus usually results in exiting the menu immediately.  Ill look and see how to do that.  Also, since its just a soft 1581, the disk header is wrong (it shows up at the first file on disk) and the blocks free is wrong. These can also be fixed as I know how to read and allocate the BAM.

### DIFF
--- a/software/application/ultimate/ultimate.cc
+++ b/software/application/ultimate/ultimate.cc
@@ -9,6 +9,7 @@
 #include "c64.h"
 #include "c64_subsys.h"
 #include "c1541.h"
+#include "c1581.h"
 #include "screen.h"
 #include "keyboard.h"
 #include "userinterface.h"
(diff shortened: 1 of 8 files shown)


xlar54:
Ive started on the capability to mount d81 disks as an emulated drive "C" with full r/w capability.  Probably by the weekend.  It will be a different PR so you can choose if this one, that one, or neither would be accepted in the official repo.

xlar54:
finally (!) got eclipse and the debugger working, useful now that im deeper into 81 mount capability.  Im utilizing the IEC code to simulate an 81 on the bus.  Not Pi1541 81 cycle exact, but it will allow mounting, r/w, etc as sd2iec does.

xlar54:
Im going to close this PR as Im working on full 1581 r/w support.  Will submit soon.

PR #114, opened 2019-09-30
Disable settings not available for firmware 09 of FPGASID

markusC64:
Works fine - just tested.

### DIFF
--- a/software/u64/sid_device_fpgasid.cc
+++ b/software/u64/sid_device_fpgasid.cc
@@ -8,6 +8,7 @@
 #include "sid_device_fpgasid.h"
 #include "config.h"
 #include <stdio.h>
+#include <string.h>
 #include "c64.h"
 #include "dump_hex.h"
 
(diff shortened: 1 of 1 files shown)


software/u64/sid_device_fpgasid.cc
```diff
@@ -129,6 +130,12 @@ SidDeviceFpgaSid :: FpgaSidConfig :: FpgaSidConfig(SidDeviceFpgaSid *parent)
 
     sprintf(name, "%b", parent->fpga_rev);
     cfg->set_string(CFG_FPGASID_FPGAREV, name);
+    
+    if (!strcmp(name, "09"))
+    {
+       cfg->disable(CFG_FPGASID_OUTPUTMODE);
+       cfg->disable(CFG_FPGASID_LEDS);
+    }
```
GideonZ: Why would you do a string compare here, if you have the value?

GideonZ: And is it really true that in version 0A these options are again gone?

markusC64:
Hm, strcmp returns 0 is equal. Therefore "if (!strcmp(...))" is executed if equal.

And "09" was the very first firmware delivered (except alpha testers). So no need to concider older firmware revisions before 09.

Having this explained, you are welcome to replace the string compare by another expression.

PR #141, opened 2020-05-11
Update VIA - timer, shift register, and CB transition detection fixes

gyurco:
These changes are verified in:
- VIC20, viavarious, via_sr and other misc tests from VICE. IFR tests
  in via_sr are failing, because it expects timing on CA2 (cassette motor
  input), which is affected by an external RC component.
- 1541 connected to C64: viavarious tests all pass
- BBC Micro: SD-Card adapter with heavy use of the shift register. It's
  using an undocumented feautre: Mode 0 actually accepts input and loads the
  shift register.

### DIFF
--- a/fpga/1541/vhdl_source/via6522.vhd
+++ b/fpga/1541/vhdl_source/via6522.vhd
@@ -97,8 +97,11 @@ architecture Gideon of via6522 is
     signal serport_en    : std_logic;
     signal ser_cb2_o     : std_logic;
     signal hs_cb2_o      : std_logic;
-    signal trigger_serial: std_logic;
-        
+    signal cb1_t_int     : std_logic;
+    signal cb1_o_int     : std_logic;
+    signal cb2_t_int     : std_logic;
+    signal cb2_o_int     : std_logic;
+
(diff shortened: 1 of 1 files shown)


GideonZ:
Dear Gyurco,

Thank you very very much for the fixes that were made to the VIA (6522). Very very much appreciated! 

Gideon

gyurco:
You're welcome, I'm glad to help!
May I request something?
As this is supposed to be a Versatile device, and was very popular, it would be good to re-use this HDL implementation (which is the best so far) in various projects, but the license text is a bit scary. Is it possible to change it to something more friendly?

GideonZ:
Sounds quite reasonable.. It's the best implementation so far, because of your additions, I think.. ;-)
I'll change the license text in the file itself and push it to the repository. From there, you can merge the license text change to your own project. Is that ok?  (In that way the license text is the same in your and my branch.)

gyurco:
Perfect, thank you!
It was the best even before :)
Actually making it the better wouldn't be possible without testing it in other cores. The 1541 works well without a cycle-exact shift register for example.

PR #169, opened 2020-07-09
add directory command to d64 menu

xlar54:
Just a quick directory viewer context menu item for D64 files.  Makes life easier dealing with disks on the machine.


tenuke:
Please add this feature =)

chrisgleissner:
Hi @Gee-64 and @GideonZ ,

it also seems to me that this can be closed now. We already have a D64 content viewer now.

Thanks
Christian

GideonZ:
Yes, close.. because it already existed; but xlar didn't know you could press CRSR-RIGHT. ;-)

PR #207, opened 2021-02-03
Add numeric response codes to modem

xlar54:
This code should provide numeric response codes via atv0

also, fixed a potential issue of showing RING from calling side.  RING should only be sent to listener


GideonZ:
A very welcome addition. Thank you, @xlar54!

Issue #216, opened 2021-03-25
FPGASID 6581 filter bias value is not signed


grue74:
6581 filter bias is using value from 0-15, but it should be signed 3-bit value.

grue74:
I don't know anything about your signed ints or whatever but this patch work a-ok.

PR #217, opened 2021-03-25
FPGASID filter bias should be handled as signed int

naali:
Fixes https://github.com/GideonZ/1541ultimate/issues/216


GideonZ:
Shouldn't this be anded with 0x0F, as it is only a 4 bit register value?

Issue #212, opened 2021-02-25
Long file name, can name be scrolled?


Mclaneinc:
I love my 1541 Ult++ but with long file names I can't always tell if I have the right disk.

I've search the docs but can't find a way to do it..Would it be something for the future?

Thanks

Paul...

Jumpman64:
...or maybe just a simple "Show filename" menu option would be sufficient, like viewing textfiles. 

You could use "Rename" as half working work-around if the name isn't longer that 38 chars but an option to display the whole name would be great indeed.

Mclaneinc:
Yup, a simple show name would be great, on my AVG cart (atari 8 bit) the author just added a key press to show the other half of the name. Scrolling looks more professional but a key press would more than enough..

Thanks for the work around..

GideonZ:
The associated pull request was merged into the development branch. Feature should appear in 3.10.

Mclaneinc:
Thanks to all involved..

PR #219, opened 2021-04-01
Squeeze file and directory names if they are too long

naali:
Adds an UI option to "User Interface Settings" -> "Filename overflow squeeze"
that allows user to select the squeeze point (None/Beginning/Middle/End)

Kind of fixes "Long file name, can name be scrolled?":
https://github.com/GideonZ/1541ultimate/issues/212

### DIFF
--- a/software/userinterface/browsable.h
+++ b/software/userinterface/browsable.h
@@ -68,6 +68,7 @@ public:
 	virtual Browsable *getParent() { return 0; }
 	virtual const char *getName() { return "Browsable"; }
 	virtual void getDisplayString(char *buffer, int width) { strncpy(buffer, getName(), width-1); }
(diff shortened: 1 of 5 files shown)


naali:
Options for setting the squeeze point:
squeeze-options


How long filenames look when squeeze point is set to middle:
filenames

This is how squeezed directory names look like when squeeze point is set to beginning:
directories

Grrrolf:
Thank you for making this configurable 👍

Mclaneinc:
Bit late to ask but will there be a shortcut key press to allow it on the fly which just cycles through the 3 styles of the squeezing

GideonZ:
Shortcut keys are a common request for many different features. Which means
that they need to be factored out into a common, preferably programmable,
mechanism.
So at this point: no.

Mclaneinc:
No problem, just glad to be getting the new feature..Thank you..

PR #222, opened 2021-05-10
enable custom cartridge rom for ultimate utility cartridge

bvl1999:
This enables a custom uuc rom entry in the cartridge selection in C64 and cartridge settings.
This entry uses the FC3 cart logic, but enables CART_REU and CART_ETH, and hence allows UCI, REU and ACIA to work


markusC64:
Have I read your patch right? "Custom UUC Rom" is actually a CRT? Well, in that case naming the entry as such would be better...

markusC64:
Gideon is about to replace the whole cartridge system anyway... You can see the filesystem intended for that already n "git master". So it's more important to define requirements.

> (currently it does for run cart but not for a flashed one)
I think it does. Flash a crt and use cartriudge type "custom crt".

markusC64:
Oh well, I'm going to test that - that would be a new bug. Given Gideon's rewrite of that part, probably not that urgent, but anyway...

bvl1999:
As to replacing the cartridge system, I am aware of that. There is a need for that also, but that is a much bigger change.

The requirements are based on https://github.com/GideonZ/final_cartridge (for which I have a whole bunch of fixes) and as a result, I'm pretty sure Gideon is aware of those. 

The motivation to not wait for a rewrite of the cartridge system...

It is a really huge limitation to not be able to use fast iec access together with a jiffydos/jaffydos/speedos/dolphindos or whatever custom kernal. This can allow for such a setup, and in fact should have been released together with the original 'hyperspeed' firmware and kernal, but due to bugs in the cartridge code, and it not working as custom crt, this was not released at that time.

If this patch should be included totally depends on when the new cartridge subsystem is going to be ready, as this functionality does not actually add any complications for such a rewrite, serves many users, and hence, it would be a shame to let it wait for too long.

bvl1999:
That sounds very cool, and could offer better compatibility with some software, but doesn't provide the ease of access/usability for the typical person who isn't going to create their own kernal, or just wants to quickly try one of the many available custom kernals, it would depend on a few who are capable of creating those 16/24k kernal images.

FeralChild64:
@markusC64 We definitely need more ROM space. I have tried to use cartridge as extra ROM blocks, but it wasn't too practical.

bvl1999:
@FeralChild64 hmm, in what way exactly? Unless you take that space from available ram, you'll still need some kind of active bank switching scheme.

markusC64:
@bvl1999 In case @FeralChild64 means ROM space for the kernal, then I can tell you that we had a discussion on forum64.de for ideas of bank switching...

Since Ultimate 64 and 1541 Ultimate II is about to reimplement hardware from last century, we need to implement the bank switching of the ROM 64er DOS V4. Please note that this is not convincing, so we should also wand a more modern bank switching available.

And at least for the 1541 Ultimate II(+), we would get 3 banks for the price of 2 (price in terms of memory usage on the Ultimate II and roughly, but not exact, for the FPGA). That's because access to internal memory can be done dword-wise.

And yes, I would appreciate supporting 16k and 25k ROMs with the upcoming rewrite of the cartridge system...

bvl1999:
@markusC64 iirc there are 2 things here:
- is it possible and usefull from the UII+/U64 side to support large kernals? That is a question that obviously is also about the technical possibilities, but also about how many people would benefit from this.

markusC64:
1st Question: Up to 3 banks are no problem at all. And I already have decompiled source code for dolphin dos 2 and 3 (c64 kernal) waiting for inclusion of a second bank to inject Gideons Fast IEC code and other... And yes, bank switching is a solved problem, but with skoe's kernal replacement logic and the wish not to "bankswitch" the ram under the rom, it's somewhat challaging. But I have a running POC with 64'er DOS V4 bank switching logic in U2+ working.
And U64 is much easier, you have direct access to CS Signal of Kernal ROM... Ok, no public source code so Gideon has to do the U64 side...
Want to improve my favrourite ROM S-Jiffydos that way, too - when there is time...

2nd Qjuestion:
Well, emulated Modules would also have the problems described for the 1st question. And if they are too simple, they do niot allow access to RAM at $e000-$fffff which is very bad. And since I know the VHDL code in the firmware, implementing not as a cartrdige is easier... the code for skoe's kernal replacement logic is already there.

bvl1999:
@markusC64 as to the 1st question, I can see why you'd want that, and I think it would be pretty cool to have and would certainly go ahead and create some custom kernal combination. I just question how usefull this is for the average user who isn't going to decompile and modify kernals and relies on others making images available. Sure, for people like you, @FeralChild64, me etc this would probably be quite usefull or at least interesting.

The 2nd question.. accessing ram below kernal rom, that is easier to do with code which isn't itself in the $e000-$ffff range, but of course that merely moves the problem to another address range. However, dealing with that has been a solved problem for almost as long as the 64 is around, with ever more advanced and capable variations appearing throughout the 1980s and early 90s. So that is really mostly a matter of picking one of the already existing implementations, re-use the glue it has for doing its bankswitching, and replacing the remainder. Kinda what Gideon started and what I'm working on now for this uuc cartridge thing. The only reason for wanting a slight firmware change for that is to support the combination of cartridge logic, REU and UCI (which is what makes it different from a regular fc3 cartridge for as far as the U64/UII+ is concerned).

FeralChild64:
@bvl1999 Yes, I have a bank switching in the ROM (in fact, I already have two switching schemes: one for RetroReplay cartridge image, one for MEGA65), and for certain builds some features are moved to external banks.

"how many people would benefit from this" - that all depends on the Kernal mods available. Take the JiffyDOS ROM, move the DOS Wedge to a separate bank, add support for DolphinDOS (protocol is really easy) + ELOAD1 + ACIA RS-232 + simple file browser + TurboTape 64 + re-add CBM format tape support + add some Ultimate64-specific features - and I bet you will get a lot of users. And all this (or nearly all) can be fit into a single 8KB bank.

Open ROMs might eventually reach similar feature level, but the compatibility will always be far worse.

markusC64:
I think FeralChild is right... From the technical standppoint, for having 2 banks we need to allocate the same amount of memory as for 3 banks on the U2+ (very technical cause) So I prefer to plan with 3 banks. That would allow a distiction with the ROM 64er DOS V4 from last century from the technical standpoint.
No one forces anyone to "use" the 3rd bank, if you do not need it, you can copy the 1st bank's content to the 3rd banjk just to make it present and a reset well defined in case the 3 bank is actived somehow.

bvl1999:
@FeralChild64 @markusC64 

I understand the possibilities you could use with more rom space, but the only argument I see in both your comments that point at something which might be fundamentally easier to do when this is done by a banking kernal, is cross platform openrom code.

I also find it rather interesting we are having this discussion at all in response to an actually already implemented scheme (just need to be able to actually select it, which is all my patch is doing) for something which lets the typical user combine 'hyperspeed' with a kernal of their liking, without the typical user having to create their own rom images, or depending on others for that, beyond that one cartridge image. 

Its a slightly different use case, and while I totally appreciate the desire for a larger kernal, imo this discussion is more about advocacy for a specific solution, which does not directly conflict with what I am proposing here. So, while interesting, I wonder why we are discussing it in relation to the patch I proposed.

markusC64:
Sorry I cannot follow - @FeralChild64 and I are speaking of a larger kernal that does not occupy any I/O register in the I/O Range making it compatible with cartridges. A cartridge like FC3 that uses IO1 resp. IO2 is not compatible because all cartridges that use that I/O range will not work properly.
So IMO that's sth completely different.

bvl1999:
@markusC64 

Yes, a custom kernal can also leave io1 and io2 free. In general, being able to combine it with some cartridge is an advantage of a kernal solution indeed (through, the mentioned fc3, in its original form, has some compatibility issues beyond just io space, it will also mess up the 'quoteless string' parsing of jiffydos and the like (fixed in the uuc code).

So we simply have 2 different situations, both with their own distinct compatibility advantages:
- large kernal solution is mostly compatible with existing cartridges but requires building a custom kernal to support one of the pre-existing alternative kernals (jiffydos, speeddos etc)
- cartridge based solution is mostly compatiblle with existing alternative kernals, but won't combine with other, similar cartridges.

That said, there is some 48k of usable rom space left for the uuc cartridge, if we use as a starting point that modifying either kernal or cartridge rom is an option, this offers a lot of room to integrate such functionality.

So.. technically different, and slightly different compatibility, and there is no reason why we can't have both (beyond available space on the cartridge, and someone having to actually implement it).

I understand you and @FeralChild64 were talking about larger kernals, but really, this is a pull request for enabling a specific combination of cartridge logic and other features. Larger kernals are an alternative for achieving the functionality this 'uuc' setup offers, but this discussion should be about what is in the pull request, and not about larger kernals.

GideonZ:
This PR is no longer required, as the cartridge logic has been rewritten, and a FC3 subtype has been defined for this.

PR #224, opened 2021-05-16
Regression fix

markusC64:
This is a regression fix for "DISK ERROR" messages in file browser since sommit a637fe9e9cf170c98f0c2f7fa2e8a0eb6baaed0c

### DIFF
--- a/software/filesystem/blockdev_flash.cc
+++ b/software/filesystem/blockdev_flash.cc
@@ -96,6 +96,8 @@ DRESULT BlockDevice_Flash::ioctl(uint8_t command, void *data)
                 return RES_ERROR;
             *dest = 1; // erase size is 1 sector
             break;
+        case CTRL_SYNC:
+            break;
         default:
             printf("IOCTL %d.\n", command);
             return RES_PARERR;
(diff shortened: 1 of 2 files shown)


GideonZ:
Merged. This only fixes the issues with /flash and /temp but not with the
usb drives (and SD card on U2).
I committed another fix for these.

markusC64:
Well, on my branch I don't see this issue on my USB device - I think it might be because filesystem/partition.cc already contains that case.

GideonZ:
partition.cc doesn't fix it for you.. because it is commented out:

DRESULT Partition::ioctl(uint8_t command, void *data)
{
    /*
     // Let's just forward the sync command, as we do not yet implement
caches.
     if(command == CTRL_SYNC) {
     return RES_OK;
     }
     */
    if (command == GET_SECTOR_COUNT) {
        *((uint32_t *)data) = length;
        return RES_OK;
    }
    if (!dev)
        return RES_NOTRDY;
    return dev->ioctl(command, data);
}

So honestly I don't know why it wouldn't occur on USB... Interesting.

GideonZ:
Just confirmed that in the previous build, the issue also occurs on USB.
Try renaming a file.

PR #225, opened 2021-05-22
Allow loading & configurationg ROM of floppies

markusC64:
This is usefull at least until redesign is finished. It allows setting of Floppy ROMs from directory "/flash" via file browser.
Can be extended to other roms whenever tehy are switched to the new method.

### DIFF
--- a/software/drive/c1541.h
+++ b/software/drive/c1541.h
@@ -164,6 +164,8 @@ public:
     // Called from user interface thread?  Is this allowed at all? -> no no, the interface thread should
     // issue a subsys command.
     void swap_disk(void);
+
+    friend class FileTypeBin; // bin file needs to access config
 };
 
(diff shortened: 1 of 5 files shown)


GideonZ:
Thanks for the patch. I'd say this may not be thread-safe... I'd have to
look into it in greater detail, but I am usually not too keen on cross
links between modules. (Passing a module config to a filetype?!)
It has to be solved in another way.

markusC64:
I see.

Still it's good to have something until that is solved more convincing. I doubt "user" will start two concurrend config updates while this is alpha.

GideonZ:
There are no users while this is alpha.

"Users" that build their own version from anything but the master branch
have no right to complain.

markusC64:
Right. That's why I have put "user" in quotation marks.

PR #248, opened 2021-08-22
Rework socket DMA for disks and cartridge files

p2mate:
Rather than downloading the whole image into a 200KB buffer and then
writing the buffer to a ramdisk file, use a 646KB buffer and append
the data to the ramdisk file as it is received. This allows for images
larger than 200KB (eg easyflash CRT images or D81 images). With some more
work this can also be used for flashing the update and the esp32 images
using socket DMA.

Signed-off-by: Peter De Schrijver <p2@psychaos.be>

### DIFF
--- a/software/network/socket_dma.cc
+++ b/software/network/socket_dma.cc
@@ -48,6 +48,9 @@
 #define SOCKET_CMD_READFLASH    0xFF75
 #define SOCKET_CMD_DEBUG_REG    0xFF76
 
+static const char tcpimage_d64[] = "tcpimage.d64";
+static const char tcpimage_crt[] = "tcpimage.crt";
+
 SocketDMA socket_dma; // global that causes the object to exist
 
 extern cart_def sid_cart;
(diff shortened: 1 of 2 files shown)


Grrrolf:
This will likely also fix the (not yet reported) issue when sending a d81 disk image with a mount-image command to the TCP port 64 command interface  hanging/crashing the Ultimate application?

GideonZ:
>
> It would be a band-aid on a gaping wound.
>
I think it is better to replace the socket dma test code with a properly
defined and documented API, e.g. through http.

>

p2mate:
I wonder if HTTP would be the right way to go. It would require a web server and a lot of parsing code because web APIs tend to be UTF-8 based. I think COAP (https://en.wikipedia.org/wiki/Constrained_Application_Protocol) would be a better idea. Opinions?

GideonZ:
Yeah, the PR doesn't need to be fixed. The whole port 64 thingie can be removed in the next firmware release, as everything that could be done through that raw tcp port can now be done through the API in a proper way.

PR #260, opened 2022-02-08
fixed issue where modem returns wrong value

xlar54:
There are more changes that need to be made to be more hayes accurate, but this one solves a problem when the modem is told to supply numeric response codes  (via ATV0 command).  It does so fine, for all responses except CONNECT responses.


xlar54:
tested and working with CNet BBS v10

xlar54:
+++ escape code returns OK when entered (or numeric 0)

ATA attempts to answer.  No response code given (except either CONNECT .. or NO CARRIER)

PR #294, opened 2022-09-27
Fix GCR decoding

markusC64:
This patch improves gcr decoding without significcantly being slower. Without it, some g64 do not decode correctly.


markusC64:
Re-pushed to solve new conflicts due to your latest commit.

Issue #306, opened 2023-03-20
Printer emulation cuts rightmost character


Schwefelholz:
Tried this in release 3.7 on Ultimate II, 3.10a on Ultimate 64 Elite and 3.10e on Ultimate II+L:

Using Commodore MPS, Epson FX-80/JX-80, IBM Graphics or IBM Proprinter emulations, once a printed line exceeds 80 characters of length, a character is cut from the right hand side. See attached printout: 1st line misses a colon, 2nd line missed an "I" on the right hand side.

![Page-001](https://user-images.githubusercontent.com/6893444/226280377-48d13788-21cf-4bd6-a2c7-38af1c1b1eb3.png)

Schwefelholz:
Now that your write this: I've seen those one-pixel column characters as well in another printout.

Is this something that can make it into the next firmware? Any chance it can get fixed in the Ultimate II?

rgc2000:
Yes, last official firmware for U2 is 3.7 and I can see that I am using a custom firmware 3.8 / FPGA 117 on my U2, not very official certainly unfinished work in progress inside. Version 3.9 does not load on U2, not enough room for secondary boot loader. Anyway, back porting the fix to 3.7 is easy.
Gideon, you are right the three targets build for U2, my local git was not up to date.

Schwefelholz:
That would be very cool!
--
Sichere Kommunikation per Kurznachricht:
Threema: https://threema.ch/download
oder
Signal: http://sgnl.link/1CYCQQN

per E-Mail:
GnuPG-Key: gpg --recv-keys 0x29AA1668
GnuPG-Fingerprint: BDB7 63C2 8A48 A6A7 25FF  6D80 C444 F8C8 29AA 1668

--
"Those who would give up essential Liberty to purchase a little temporary Safety, deserve neither Liberty nor Safety." (Benjamin Franklin)

Am 21. März 2023 00:03:16 MEZ schrieb "René" ***@***.***>:

Grrrolf:
> As far as I'm aware, the latest official firmware for U2 is 3.7.

The latest firmware can be found here: https://ultimate64.com/Firmware and it is version 3.10a. There are also U2 updaters in the zip file. Networking may be broken for the U2 in this firmware, according to reports in the FB group.

rgc2000:
[...]

PR #307, opened 2023-03-22
Virtual printer: fixed lost/cut character on the right of the page

rgc2000:
Printable area is 1 pixel too wide and too long making the virtual printer to attempt to print an extra character on the right, cut by the lack of space. This was a right margin mistake as the first pixel of the row is number 0 and not number 1. #306


rgc2000:
Something wrong with this merge request. Too many files involved. I close it and I will create a new one. The fix is only in printer source files.

PR #315, opened 2023-04-15
Fix drv info

markusC64:
CMD_GET_DRVINFO returned the hardware ID of the device (that is the one the DIP switches of the 1541 II are configured) instead of the software ID.

This patch improved the UCI call that by default the software ID is returned, but the calling programm can specify that the hardware ID is wanted.

### DIFF
--- a/software/drive/c1541.cc
+++ b/software/drive/c1541.cc
@@ -367,6 +367,13 @@ int  C1541 :: get_current_iec_address(void)
 	return iec_address;
 }
 
+int  C1541 :: get_effective_iec_address(void)
+{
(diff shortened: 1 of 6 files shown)


xahmol:
Strongly support this change, as my GeoUMount fails in many configurations as GEOS often changes device IDs.
It was also checked with and confirmed by Bart that the change at least does not impact Device Manager ROM.

markusC64:
Getting the right device for mounting, unmounting or disk swapping should not consider turned off devices.
And it should better use the effevtice device id.

Fixed.

PR #316, opened 2023-04-15
Geos RAMDISK support in UCI

markusC64:
This patch enabled UCI loading disk images into the GEOS Ramdisk.

### DIFF
--- a/software/filemanager/dos.cc
+++ b/software/filemanager/dos.cc
@@ -4,6 +4,7 @@
 #include "endianness.h"
 #include "rtc.h"
 #include <string.h>
+#include "c64.h"
 
 static const char* wdnames[7] = { "SUN", "MON", "TUE", "WED", "THU", "FRI", "SAT" };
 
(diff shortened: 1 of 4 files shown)


xahmol:
Functionality included and tested to work in latest GeoUMount builds, enabling mounting images under GEOS as ultra fast RAM drives.

GideonZ (commented):
I need more time to review this.
Does it really belong in the DOS target?
Actually I think that the DOS target is not anymore just a file interface.. It should be cleaned up.
Maybe it should be part of a GEOS target?

markusC64:
Well, loading a file into some adress space needs the concept of the current directory to which the filename is. No other target has that.

And another point: What is the main difference between loading parts of the REU (which is part of the DOS target) and loading RAMDISK (which essentially is loading to REU with some additional checks and computation of target address inside the REU)?

I only have added the two functions to the DOS target that deals with files, that is loading image into ramdisk and saving ramdisk as image.

The query "what ramdisks are there" is in the control target.

GideonZ:
Yes, I have exactly the same question about loading files into the REU and
setting the time and all that.. Doesn't really belong in DOS.  But well,
that's the way it is now.

The DOS is also limited to having just 2 files open, which is actually not
how it should be. I think it should be deprecated.

markusC64:
Perhaps right, but moving REU load to another target would certainly break C64 programs using that. It happens that I know some.

BTW: Before hyperspeed (kernal), loading the data into the REU and then use DMA of the REU was by far the fasted way to load data via UCI. The same for saving data.

xahmol:
Demo video of my tool actually using this functionality:
https://youtu.be/4MEEn1nNpJc

GideonZ:
I decided to simply merge it... :)

PR #317, opened 2023-04-15
Fix renaming files in DNP iamge

markusC64:
Renaming files in DNP images did not work in subdirectories

### DIFF
--- a/software/filesystem/filesystem_d64.cc
+++ b/software/filesystem/filesystem_d64.cc
@@ -479,12 +479,36 @@ FRESULT FileSystemCBM::file_open(const char *pathname, uint8_t flags, File **fil
 
 FRESULT FileSystemCBM::file_rename(const char *old_name, const char *new_name)
 {
+    PathInfo pi(this);
+    pi.init(old_name);
+    PathStatus_t pres = walk_path(pi);
+    if (pres == e_DirNotFound) {
+        return FR_NO_PATH;
+    }
+    if (pres != e_EntryFound) {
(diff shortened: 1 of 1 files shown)


GideonZ (commented):
Looks a lot better than what I had! Thank you!

However, how would this work:

rename dir_a/prog.prg new.prg

Would this move the file from the 'dir_a' to the root? I think the file name still only works within the same directory, while it should work within the same filesystem, I think.

markusC64:
I think moving files would not work.

Perhaps we should distinguish between renaming and moving. Moving is more complex because you have to differ between moves inside a filesystem and moves between filesystems. Seperate calls for move might therefore make sense.

GideonZ:
Agreed. To some extent. The move operation falls apart into checking
whether source and destination are in the same file system. If so, it
should call 'rename' of that particular filesystem. If not, it should
either give an error, or perform a copy & delete on success. You cannot
always do a copy & delete when the source and destination are on the same
filesystem, because there might not be enough space for a copy of the same
file. So, I stick to my original statement, that a rename should support
moving between directories. Unless all filesystems get a 'move' operation,
which is ... a different implementation of rename?

markusC64:
I think move and rename are quite different. rename operates in exactly one directory and it is sufficient to change an existing directory entry (only the name part).
In contrast to move. It operates on two directories, it has to invalidate the source directory entry and create a new directory anty in the destination directory.

So separated implementation makes sense. And a function "renameOrMove" could call what matches for the actual call.

PR #329, opened 2023-04-25
Fix writing Dxx Images

markusC64:
To reproduce the problem to be solved:

(1) Load a D64 image (or probably other image) into the Floppy Drive A
(2) F5, Drive A, Save Binary, Enter Filename
(3) Observe that the size is not the expected size, in case d64 it should be 683 × 256 ≈ 171k, but the filemanager shows 172k and its size is exactly 686 × 256.

This patch fixes that.


markusC64:
Re-Pushed due to your latest commit.

Now the problem is solved, but some code optimization is possible, see the comments in your commit on github.

Issue #342, opened 2023-05-07
Printer : unable to print large files


rgc2000:
A change in the IEC poll method causes the IEC processor to be disabled while printing large files (with lots of IEC trafic) and I have the ? DEVICE NOT PRESENT ERROR message in the middle of a print job.

In the debug console I can see the message "-> ** IEC Processor seems to have gone haywire. Stopped! **" Depending on the cartridge this happens sooner (on U2+) or later (on U2). I need to comment out the looplimit code that was added recently to be able to print again until the job is done.

GideonZ:
Hehe... Oh god.

GideonZ:
It is an active polling, but with a task delay in the loop. Many mechanisms
are derived from the event driven implementation of firmware 2.x. There are
still more instances in the firmware where a task is created that calls the
old "poll" function repeatedly.

I am surprised that the FIFO status is actually non-empty for such a long
time. You see, I would expect that the CPU is fast enough to take at most
some chars from the FIFO and then retire, since IEC is slow.

I think the problem is solved completely when the data from the FIFO is
pushed into a queue to the printer handler. Maybe there should be a special
control character to keep the IEC bus 'busy' if a queue is full, but the
data should not remain in the FIFO. That is not what it was meant for.

GideonZ:
I propose to implement the loop limit as follows:

```
        uint8_t a;
        for (int loopcnt = 0; loopcnt < 500; loopcnt++) {
            a = HW_IEC_RX_FIFO_STATUS;
            if (a & IEC_FIFO_EMPTY) {
                break;
            }
```

This would limit the number of bytes transferred in one "poll" to 500, but it will always return to the TaskDelay after that to allow other threads get time. What do you think?

rgc2000:
I'm going to test this on my two cartridges. I can see now how it works and the place where the scheduler can switch to other tasks. I was very surprised about the active polling in a multitask OS. I also understand why the cartridge is unresponsive while printing. The IEC bus can be very busy while printing, the C64 floods the printer and the whole job can be done in the IEC FIFO loop without breaking to previous infinite loop.
The PNG generation is still a very time consuming procedure and there is no task switching during this operation. That's why I added a modal message with a progress bar if it is started by the action menu "printer - flush/eject".

rgc2000:
[...]

Issue #339, opened 2023-04-28
IEC processor can continue listening after EOI


rgc2000:
On very rare occasions, the IEC program allows the IEC processor to continue listening after EOI. This may acknowledge a byte that is not intended for the software IEC.

To reproduce this :
- First talk to an existing software IEC device. You can use C64 BASIC, eg: LOAD "$",10  (default devices ID on software IEC are 10 for IEC drive or 4 for printer).  The data transmission must be followed by a EOI (should be the case).
- Then use C64 BASIC to talk to another device ID that is not on the IEC bus (eg 12).
- We expect the error "? DEVICE NOT PRESENT ERROR" but if you fall In the IEC bug you will get no error and BASIC will hang thinking that the device ID exists.

Depending which part of the code is executed, the LISTENER flag may not able cleared since last CHECK_ATN_BYTE sequence. To prevent this you can add "SET LISTENER=0" when receiving EOI. No one should be listening anymore after EOI, a new ATN sequence should select the next listener.

Here is a suggested patch or maybe we could add SET LISTENER=0 and SET TALKER=0 at _dont_print label in code (because this is where I suspect LISTENER to keep set to 1 after receiving a new ATN not intended for software IEC.

[iec_code.patch](https://github.com/GideonZ/1541ultimate/files/11357352/iec_code.patch)

radius75:
```
You can use C64 BASIC, eg: LOAD "$",10 (default devices ID on software IEC are 10 for IEC drive or 4 for printer). The data
```
Off topic:
In FW3.10f the default ID is 11.
In FW3.10a I think it was 10.
Interesting. Maybe it doesn't matter.
I'm reporting it anyway.

rgc2000:
You are right, default IEC drive ID is 11 now but in the IEC code it is still 10. It should not be a problem as the value in IEC code is replaced by the configured ID at initialization. Search/replace is still looking for ID 10 for replacement.

Issue #308, opened 2023-03-22
The virtual printer stops working and the status of the USB key changes to FAILED


rgc2000:
It seems that there is a conflict between the virtual printer process and the USB stack. Printer is very CPU intensive and is running in the IEC thread which has higher priority than USB stack. This makes the USB diver to timeout on key access, making the system think that the USB key is unresponsive.

rgc2000:
Changes in the printer code to fix this issue : 
- Interpreter procedure is now a Task created with the lowest priority (tskIDLE_PRIORITY)
- This task is normally suspended waiting for a TaskNotify event to run and process the incoming data from IEC.
- IEC task sends the TaskNotify to printer task when printer data is available.
- IEC task is suspend while the printer task is processing the data waiting for another TaskNotify from printer.
- When all the data has been processed, printer sends a TaskNotify event to IEC to go on. Printer task goes back to sleep waiting for a new TaskNotify from IEC to process new data.

Considerations :
- IEC has to be suspended while printing. This is how a real printer works, holding the IEC bus while the printer is busy.
- TaskNotify functions are used as binary semaphores here
- Call to ulTaskNotifyTake and xTaskNotifyGive. Not using "index" versions go theses functions. Same as using index 0. Should not be a problem I can't find any TaskUsage call in the rest of the code.
- Call to ulTaskNotifyTake with timeout portMAX_DELAY. Should mean no timeout if INCLUDE_vTaskSuspend is set to 1 in FreeRTOS configuration file. By security return code is tested and task is put back to sleep if return code means timeout.
- The printer flush procedure is not handled by the printer task. This means that it runs in the GUI interface task. It seems not to disturb the USB task. I have never experienced USB errors on printer flush.

Questions someone may answer :
- I thought that making the printer task to run with the lowest priority could make the cartridge to be responsive again while printing. Unfortunately not. I still can't invoke Ultimate screen while printing.
- I can hear that some new sounds have been added to 3.10 (disk inserted). Do we want to add printer noise to the cartridge ? Maybe yes if it is no CPU consuming and as easy as a trigger to call to start/stop the printer sound. I already use a LED on/off call to show printer activity.

rgc2000:
[...]

PR #345, opened 2023-05-11
Virtual printer fixes and enhancements

rgc2000:
Fixes:
- Page width calculation was wrong and the rightmost character was cut or lost. Fixes Issue #306
- Virtual printer is no longer a subroutine of IEC task. It runs in its own task and has lower priority than USB task. This prevents timeouts when accessing USB key filesystem. Fixes issue #308 
- IEC processor code will stop listening on IEC after an EOI. This prevents the software IEC to acknowledge bytes that are not intended to it. Fixes issue #339
- Disable printer target in polling routine when an ATN starts. This prevents to misroute some data for IEC drive to printer  just after using printer when no EOI is received before ATN.
- Added Gideon's workaround to prevent CPU task starving on large printer jobs without disabling IEC processor. Fixes issue #342
- Activity LED will be lit while generating PNG file also when USB drive is the file destination.
- PNG generation code has been rewritten to use much less memory. On 3.10 firmwares and later the color printer was broken on Ultimate-II (not +). Now, color output is available again on all devices. Code is also a bit faster but not very noticeable and much more smaller as large parts of unused code have been deleted.

Enhancements:
- Virtual printer has its own configuration menu and its own configuration section in config files. You will probably need to reconfigure your printer settings after firmware update.
- Virtual printer can be disabled independently of IEC drive
- Action menu to flush/eject printer page will now show a modal dialog with a progression bar.
- Printer documentation updated


rgc2000:
Maybe you can see in this merge request that the first 21 commits are not related with the printer enhancements I have done these last weeks. I don't know why git is showing them as they are not my own and they are already part of the Gigeon's master branch. If you browse the file changed list, you can see that no files are changed by those commits.
@GideonZ if this disturbs you I can try to create a new branch from your master and to commit only printer changes but I would lose the changes history.

The most noticeable changes are : 
- Sources for printer have been moved to a new directory software/io/printer and are no longer in software/io/iec. Makefiles have been updated to include this new path
- lodepng has been replaced by a thinner version. Makefiles have been updated for this file replacement.
- There are 3 versions of the IEC processor code. One for printer and drive, one for printer only, one for drive only. Code load has been moved to a new method iec_processor_configure(). It is called by two methods used to enable/disable and configure IDs on IEC bus:  iec_drive_enable() and iec_printer_enable()
- The only FPGA code evolution is to add a new printer_busy register in the ITU to stop using the usb_busy register to access the activity LED. It uses very few gates and fits in Ultimate-II.

radius75:
In my opinion, the {Files changed} tab is the most important here.
Here are shown what changes to "GideonZ/1541ultimate:master" will result from the Request #345 Merge option.
Only this should be carefully analyzed before Merge :)
https://github.com/GideonZ/1541ultimate/pull/345/files

And all those "weird" commits in your PR are due to you updating your branch by fetching new changes from GideonZ/1541ultimate:master.

Take a look at rgc2000/1541ultimate:master in Network, you can move the view with the mouse.
https://github.com/GideonZ/1541ultimate/network

radius75:
Here.
You updated your Master without doing a Fetch with Gideon/master.
That's why you have these Commits at your place.
You must analyze all changes in the {Files changed} tab to confirm that you have not accidentally included a Commit that is not yet Printer (eg from other Repos) and should not be added to Gideon:master.
obraz

rgc2000:
I have carefully reviewed all the changes in the "Files Changed" tab and everything is correct as expected. No extra or unwanted changes reported.
I guess the those mysterious commits are the consequence of manual local fetches from remote origin repository to sync my branch with Gideon's master instead of using pull requests in the web site. Now I use the pool request feature to sync with latest changes.

rgc2000:
pull request updated to solve conflicts if merged with 3.10i
tested successfully on U2_RiscV and U2+

GideonZ:
I created a merge with the latest Wifi feature branch in "printer_enh".
I think it won't work yet, because the initialization is not done in the right order.
The IEC is now initialized by means of an InitFunction, and the reference is now a pointer.
However, the constructor of the IEC already references the printer pointer, which is probably still NULL.

rgc2000:
I'm going to take a look at the printer_enh branch and check if init has to be changed. It should not be a big issue as the printer class instance is a singleton. It should be created at first access if not already done. I know that I have done something that was not dependent of the IEC object init order.

GideonZ:
I am sure you did it all correctly.. but in my merge operation, I screwed it up. I didn't have time anymore to fix it as I needed to do other things.

But... that said; I'd like to make it a bit cleaner. I'd like to separate the IecInterface from the drive completely. The drive is, just like the printer, a slave to the IecInterface. As you can see, in the IecInterface class the channels are created. These are not a property of the IecInterface but a property of the drive. Which brings me to the first question: Does a printer have channels at all? Could a printer have a status channel? (e.g. to report out of paper, out of ink, paper jam, printer type, whatever...).

Separating the IecInterface from the drive will also allow two instances of the drive to be attached to the interface. Or a third, completely different type of device.

rgc2000:
Mmm, I was too confident. I need to understand how your InitFunction class works and then found a solution to make the IEC and Printer objects work together, regardless of initialization order.

GideonZ:
InitFunction: I think there is a parameter that can be passed to give
relative order. The list is first sorted, then executed.

Clean: If the drive and printer "register" themselves to the IEC driver,
then the IEC driver could create itself with a singleton construction. In
this case, the order is always guaranteed. Plus... this way, the IEC driver
is never instantiated when there are no clients.

rgc2000:
Trying to compile your code out if the git depot : 
It compiles with no error for u64, u2+ and u2+L
But I am unable to compile for u2 :
- on u2 mb, I have a Xillinx error on PAR step for double disk version of the firmware. A timing error : TIMEGRP "usb_control" OFFSET = IN 10.5 ns BEFORE COMP "ULPI_CLOCK"
- on u2 rv, I have a ld error while creating output.out in target/u2/riscv/updater with symbols defined multiple times (_exit, _isatty, _brk, _sbrk) in neorv32/sw/lib/source/syscalls.c and in software/portable/riscv/riscv_main.c

markusC64:
I have to confirm that "printer_enh" does not compile for the U2 RiscV.

I cannot tell about microblaze because I don't care about microblaze. AFAIK microblaze has passed EOL.

GideonZ:
Just don't flash these builds as they are now, as the init will crash and
thus brick the unit.

markusC64:
BTW: Branch cli fails on the U2 with the same error.

GideonZ:
For the U64 FPGA, I think it is better to leave it out of the git
repository and use a mechanism to fetch it somewhere else.. OR maybe as an
intermediate solution use git LFS.
Do you have any ideas about this, Markus?

markusC64:
If I would have to decide, I would do it in the following way:

1) Remove the external/* file (there is only one)
2) Add "external/README" with some content, just that the folder is non empty.
3) Change the top level makefile to use "curl" to fetch the latest fpga and place it inside the external folder if and only if the fpga blob does not exist.

Note: That way, one can also uses an archived fpga.

4) For the U2+ / U2+L builds: If there is an FPGA blob within the /external directory for that device, use that instead of building one. So builds are much faster when they do not change the fpga - assuming that the user of the compiler has placed his last fpga output inside the "external" directory. If it simplifies things, the two cases could be separate make targets.

5) If possible, another top level target could fetch a prebuild fpga for U2+ / U2+L using curl - to speed things up even more.

rgc2000:
I fixed the IEC and printer object initialization dependency. IEC no longer needs to call the init_done() print method. This method has been removed because the printer now has another way to know that initialization has been completed for IEC and the printer.

I have tested this on my own master branch, not a clone of the printer_enh branch. I don't know if the whole firmware is stable enough to upgrade my U2+. Can it brick the device ? I guess that wifi is only for U64. I prefer testing it on the U2 as the ultimate.bin loading facility on boot makes the tests much easier.

@GideonZ to answer about printer channel : the printer never talks on IEC, it is only receiving data from the IEC master. There is no status channel to inform the C64 that the printer is out of paper/ink, etc...  IEC can send 2 type of information to the printer : data and commands. Data are the bytes sent to printer by the program running on the C64 they are interpreted by the printer emulation according to the EPSON/CBM/IBM instruction set selected by the user. Commands are only 4 possible : open printer, close printer, cursor up and cursor down. I am using open and close to open/close output file when printing in RAW/TXT mode. Cursor up/down are used to communicate to the printer the secondary device address used to select upper/lower case charset when printer is working with CBM emulation.

markusC64:
> I prefer testing it on the U2 as the ultimate.bin loading facility on boot makes the tests much easier.

Well, testing on the U2+ is even easier if you have an usb blaster. Then you can trasfer the new application by wire and have it started any time. Without having to powercycle the C64. So in case I expect several attempts before my change works, I prefer U2+ (or U64, it's the same on U64) over U2 and U2+L.

rgc2000:
I have just tested my changes on U2+ with 3.11⍺ and printer seem to work fine. I have noticed a small bug, disabling IEC drive from F5 action menu has no effect.
I am going to clone that branch in my git repo to be able to send you the fixes.

rgc2000:
Pull ressues #385 created to fix init sequence
(Now I remember why I prefer U2 against U2+ to test and improve the printer : because printer is much faster on U2 !)

rgc2000:
A small bug fix added to the pull request : when turning off/on the IEC or Printer from the action F5 menu, the F2 Setting page may not show the current status (Enabled/Disabled) for IEC or Printer.

Bad news : I think I will not be able to do more dev/tests for a while. I have fried my C64, I have to find a new one. I am still able to access the Ultimate setting but it is not stable and the C64 side does not work. Something wrong with VIC-II or CIA, I don't know.

rgc2000:
All code changes have already been merged to Gideon master branch. The only changes remaining have been added later for the doc directory to explain how to build the standalone virtual printer on a Linux system.

Issue #151, opened 2020-06-15
Feature request: machine code monitor


lagomorph:
A built in monitor (even if extremely basic) that could simply examine an address and change its value would be an extremely useful enhancement. While various monitors exist in cartridge and software form, they alter the system and are not always functional in all situations.

Zibri:
I agree. Pressing the freeze button with no cartridge loaded should go straight into a Monitor.
The one of retroreplay would suffice.

Jusalak:
Even memory poke functionality would be a nice addition. It probably is not too difficult to implement.

Jusalak:
A memory dump display and accompanying hex input would be a good start and far easier to implement.

GideonZ:
Indeed. For this, the 'viewer' class should start supporting editing.
INterestingly the class is called "Editor", but it never got the edit
functions.
Sub classes could implement special operations for reading and writing, so
it would be relatively easy to edit memory.
It is just a hassle to implement this with just 'move' and 'out' functions.

chrisgleissner:
@GideonZ I found some poke functionality in the context-sensitive menu of the U64 which you tied to a feature flag, disabled in prod. I enabled and extended it, and it seems exactly what we need as a starting point. I can peek values of the C64 RAM (C64 running, Ultimate UI in overlay) and poke them, and both sides (C64 and Ultimate UI) see the correct values.

Was there any reason why you decided to not expose the Poke functionality in prod, and should this feature also work for non-Ultimate-64 devices?

chrisgleissner:
Very happy to finally close this one.

What started here as a request for even a basic memory monitor has grown into a full-featured Machine Code Monitor, implemented in #358 and now part of the upcoming firmware 3.15 release.

This has been a substantial piece of work for me, and I am very happy and proud to see it becoming part of the firmware after all these years. Thank you @lagomorph for starting the discussion, and everyone who contributed feedback and testing along the way. 

Also, thank you very much to @GideonZ for his help, guidance, and for trusting me with this work.

Please find the new documentation here: https://1541u-documentation.readthedocs.io/en/latest/howto/machine_code_monitor.html

I am also still working on **debugger** support in #705. This will allow for step in/out/over, 10 breakpoints, and much more. That work is now almost feature-complete, but did not quite make the cut-off for 3.15. I plan to release in in 3.16. If anyone would like to try it and provide feedback, I would be very grateful.

Closing this original feature request as completed.

PR #357, opened 2023-07-04
Added hex viewer

chrisgleissner:
# Features
- Added hex viewer (https://github.com/GideonZ/1541ultimate/issues/320), using a format popular with various [C64 Machine Code Monitors](https://www.c64-wiki.com/wiki/Machine_Code_Monitor): 4 byte offset, 8 bytes data, first in hex, then in the Ultimate's UI character set with dots for bytes that can't be rendered. 
- Support new editor shortcuts: Home/F2: start of file. End/F8: End of file. This facilitates navigating across huge files.

# Design Notes
- The decision about which view to use for opening a file was deliberately left to the user. There may be legit use cases to open a text file in hex view, and vice versa.
- The firmware resulting from this PR was successfully flashed to a 64 Ultimate Elite, revision 1.4. All testing was done on this machine.
- The offset rolls over from FFFF to 0000 if browsing a file larger than 64KiB. This should be OK as it mostly concerns browsing D64 and similar large files.

# Examples

New "Hex View" entry in context-sensitive menu:
image

New "Hex View":
image

Same file opened via existing "View" which was left unchanged. In this example, byte 0x00 at offset 0x3b terminates the string rendering:
image

New "Hex View" using sample file containg contiguous sequence of numbers from 1 through 255:
image

PR #358, opened 2023-07-08
Add machine code monitor

chrisgleissner:
# Add C64 Machine Code Monitor

Addresses #151

This PR adds a full-featured C64 machine code monitor for Ultimate-based devices.

- **Views and editing:** Supports Hex, Binary, ASCII, Screen Code, and 6510 Assembly views, including optional undocumented opcodes, inline editing, copy/paste, bookmarks, and number conversion/calculation.
- **Assembly coding:** Convenient on-device assembly coding with built-in automatic opcode lookup that enforces valid syntax but does not hinder normal typing. Understand program flow by following/returning from jump/branching opcodes.
- **Monitor commands:** Includes jump, fill, transfer, compare, hunt, load/save, and `G` execution.
- **Banking and memory:** Provides CPU bank-aware memory access, VIC bank visibility, RAM-under-ROM handling, and memory-source annotations for `RAM`, `BASIC`, `CHAR`, `IO`, and `KERNAL`.
- **Runtime modes:** Supports overlay and freeze modes, restores session state on reopen, and allows concurrent use from multiple Telnet sessions and the device UI.

This PR has been tested on:

- Ultimate 64 Elite I
- C128 in 128 mode with UII+ (thanks @bvl1999)

## Docs

[machine_monitor.md](https://github.com/chrisgleissner/1541ultimate/blob/feature/151-machine-code-monitor/doc/machine_code_monitor.md)

## Video Walkthrough

- General: https://youtu.be/swDOzfrO0xI
- Bookmarks / Polling: https://www.youtube.com/watch?v=8Js9uDSfLco
- Number Conversion: https://youtu.be/_k8_4sl7cns

## Entry Points

* Device: 
  - Menu / `CBM+O` shortcut 
  - Menu / `F5` / `Developer / `Machine Code Monitor`
* Telnet access

## Features

* Views: Hex, Binary, ASCII, Screen Code, Assembly (incl. optional illegal opcodes)
* Edit and copy/paste in all views
* Memory operations: goto, fill, transfer, compare, hunt

### Banking

* CPU bank–aware memory access
* VIC bank shown in status line
* Memory source annotations: `RAM`, `BASIC`, `CHAR`, `IO`, `KERNAL`

### Session Handling

* Session state restored on reopen
* Concurrent use via multiple Telnet sessions and directly on the device

## Backend

* Supports overlay and freeze modes
* Handles RAM-under-ROM writes
* Uses CPU-mapped reads in freeze mode

## Testing

* Unit tests
* Many manual tests (see video walk-through above)
* E2E tests interacting with machine code monitor on Ultimate 64 Elite:

```
chris@mickey:~/dev/c64/1541ultimate/tools/developer/machine-code-monitor$ ./monitor_test.py
[01] initial CPU7/KERNAL monitor status ... OK
[02] KERNAL $E000 hex view and REST match ... OK
[03] paging away and back keeps memory view stable ... OK
[04] KERNAL disassembly formatting ... OK
[05] KERNAL $E010 REST match ... OK
[06] CPU6 RAM under BASIC write/read ... OK
[07] CPU5 RAM under KERNAL status ... OK
[08] ASCII view width and scrolling ... OK
[09] HEX edit writes both nibbles ... OK
[10] CPU bank cycling reaches CHAR and RAM mappings ... OK
[11] COMPARE reports differing address ... OK
[12] G executes finite loop and returns to monitor ... OK
[13] G repeated execution updates RAM sentinel ... OK
[14] G handoff preserves stable VIC state ... OK
monitor_test: OK (14 checks)
```

---

## Keyboard Shortcuts

For full details, please see the official [documentation](https://github.com/chrisgleissner/1541ultimate/blob/feature/151-machine-code-monitor/doc/machine_monitor.md).

### General

* `CBM+O`: **O**pen monitor
* `X` or `CBM+O`: E**x**it monitor
* `F3` or `?`: Toggle help
* Cursor keys: Navigate by row/column
* `Page Up` or `F1`: Navigate one page up
* `Page Down` or `F7`: Navigate one page down

Please note that interacting with the machine code monitor often requires switching the two user interface modes (freeze and overlay):
- Overlay allows monitoring the C64 behaviour whilst changing RAM with the monitor
- Freeze mode allows for streaming the monitor's UI over the network, e.g. using C64 Stream. In Overlay mode, only the C64 video is streamed, hiding the C64 app menu, and thus the monitor.
 
To simplify this UI switch, I added a new keyboard shortcut accessible from the menu:
* `CBM+I`: Toggles the User **I**nterface Mode and exits the menu (akin to Joystick toggle)  

### Views

* `M`: **M**emory
* `A`: **A**ssembly
* `B`: **B**inary
* `I`: ASCI**I**
* `V`: **V**IC Screen Code
* `O`: CPU bank **o**ptions
* `Shift-O`: VIC bank **o**ptions
* `P`: Toggle **p**olling; great for observing live changes in UI Overlay mode

### Edit Mode

* `E`: **E**dit
* `CBM+E`: Exit **e**dit
* `RUN/STOP` or `ESC`: Exit edit mode or monitor

### Copy / Paste

* `R`: Mark **r**egion start/end
* `CBM-C`: **C**opy
* `CBM-V`: Paste

### Bookmarks

Bookmarks allow for storing up to 10 different regions (and user-preferred view settings for them). This allows for very quick jumping to high value memory regions. 

* `CBM-0` ... `CBM-9`: Jump to bookmark
* `CBM-B`: Open **b**ookmarks list (allows for jump, edit, delete)

### Delete

The DEL key behaviour depends on the view and is always enabled, even if not in Edit mode:

- Assembly: Overwrites with `NOP`, moves cursor forward
- Memory: Overwrites with `00`, moves cursor forward
- Binary: Overwrites with `0` (shown as `.`), moves cursor forward 
- ASCII/Screen: Overwrites with space, moves cursor left (to operate more like a text editor)

#### Assembly

* Up/Down: Step one instruction
* Left/Right: Step one byte
* `U`: Toggle **u**ndocumented opcodes

##### Follow / Return

The Assembly view supports lightweight control-flow navigation with `RETURN`:

* If the cursor is on a `JMP`, `JSR`, or branch instruction, `RETURN` follows its target and saves the current position on a monitor-local stack.
* If the cursor is not on a followable instruction, `RETURN` returns to the most recent saved position, if any.
* The follow/return stack holds up to 10 entries.

#### Memory

* `W`: Toggle **w**idth, 8 or 16 bytes. In 8 byte mode, an ASCII representation of the bytes appears to the right.

#### Binary

* `W`: Toggle **w**idth, 1-4 bytes. Use 1 byte for charset editing and 3 bytes for sprites.

---

## Freeze

* `Z`: Free**z**e in overlay mode
* Always frozen in freeze mode

---

## Memory Commands

ASCII and Screen Code edit modes must be exited before using memory commands, to avoid ambiguity.

* `J`: **J**ump to address
* `F`: **F**ill memory range
* `T`: **T**ransfer memory range
* `C`: **C**ompare memory ranges
* `H`: **H**unt for byte sequence
* `N`: Evaluate **n**umber
  * Shows hex, decimal, binary, ASCII, and Screen Code representations of an 8-bit or 16-bit value.
  * `RETURN` writes the selected value to the current address in the active memory view.
  * Includes a built-in **calculator** that supports +, -, *, and / operators, as well as decimal, hexadecimal, and binary operands.
  * `ESC` closes without writing.

---

## Execution Commands

* `G`: **G**o, executing the program at the specified address.

---

## Disk Commands

* `L`: **L**oad file to RAM. Choose PRG header address or absolute target address; optional file offset and length supported.
* `S`: **S**ave RAM to file: 
  1. Overwrite existing file: use "Select File" in file menu
  2. Create new file: press Enter on synthetic "<Create File>" entry only shown if file system is used as file picker

Files can be loaded from / saved to either U64 filesystem files, or alternatively files within C64 disk files such as .d64 files.

---

### Developer Menu

The machine code monitor can be opened in one of two ways:

1. Open menu, press `CBM+O`. (**O** as in m**O**nitor; the m shortcut is already taken)
2. Open menu, press `F5` - `D` - `M` - `Enter`. In other words, via the Action Menu's "Developer" sub-menu, as seen in the screenshot below. 

Screenshot 2026-05-01 19-02-12

### Freeze Mode

The "Freeze Mode" is used if the user selects `Interface Type = Freeze` in the `User Interface Settings` of the menu.

The freeze mode is largely identical to the overlay mode discussed below, but it does not have access to the BASIC nor KERNAL regions. Instead, it views the memory beneath them. 

#### Help Menu

Screenshot 2026-05-14 10-55-21

#### Memory (i.e. Hex) Mode

Screenshot 2026-04-29 22-53-16

#### Assembly Mode

The following steps take us to this screenshot
1. Copy Kernal to `$1000` with the `T E000-FFFF,1000` command (only works in Overlay UI Mode since Kernal not visible in Freeze UI mode)
2. Enter `E`(dit) mode
5.  Type `ST`

Screenshot 2026-04-29 22-53-01

From here, the user would normally continue as follows:
1. As the user types, a popup appears with all opcodes that match the characters typed so far.
2. The user can either select one of the opcodes and populate its arguments or alternatively type out the entire line.
3. The cursor automatically advances to the next line.

#### Binary Mode

Useful for viewing/editing the character set (as shown here) or sprites. Configurable width (**W** shortcut) of 1 to 4 bytes per row, with a sprite mode (`3S`) that removes spaces between binary byte representations.

W=1
Screenshot 2026-05-02 11-11-46

W=2
Screenshot 2026-05-02 11-11-52

W=3
Screenshot 2026-05-02 11-11-55

W=3S
Screenshot 2026-05-02 11-13-44

#### ASCII Mode

Please note that the character set used by the UI does not include all characters of the standard C64 character set, so this is a best effort rendering.

Screenshot 2026-04-29 22-53-20

#### Screen Code Mode

As mentioned under "ASCII mode", this is best effort rendering.

Screenshot 2026-04-29 22-53-26

#### Evaluate Number

See the Hexadecimal / Decimal / Binary / ASCII / Screen Code representation of the value at the current cursor position.

Screenshot 2026-04-29 22-50-37

#### Bookmarks

Configure up to 10 bookmarks for different memory regions and view settings. Comes with overrideable defaults.

Screenshot 2026-05-02 11-11-37

#### Save and Load

This section demonstrates how a byte range is saved, then loaded again. This relies on the standard Ultimate 64 file picker.

From U64...

Screenshot 2026-04-29 23-10-46

Screenshot 2026-04-29 23-12-13

...To C64U...

Screenshot 2026-04-29 23-12-33

... And Back Again.

Screenshot 2026-04-29 23-12-36

### Overlay Mode

The overlay mode allows for full visibility of the C64's RAM, including BASIC and Kernal. 

By default, the C64 is not frozen. However, if desired, its freeze status can be toggled via the `Z` shortcut. Great for examining memory without the C64 changing it.

Please excuse the image quality; the overlay is not visible in U64's video stream, thus I had to take photos. These are somewhat outdated as the code has moved on. For the latest code state, see the Freeze mode screenshots.

## Limitations

- C64 monitor only. No 1541 drive monitor is included here.
- No stepping, breakpoints, or execution control.
- In freeze mode, ROM/RAM visibility follows the monitor's selected CPU bank. The live 6510 `$0001` latch is not directly observable there, so the selected bank is treated as authoritative.
- No extra cartridge-specific banking UI beyond the existing platform behavior.


GideonZ:
Looks quite good. I was able to build the U64 target. As soon as I have set up my U64 again (still in a box from a Commodore party), I will test it out. I can see you are on the right track also with separating the editor by type. Have you been able to figure out what happens with files larger than 26K?

chrisgleissner:
Thanks Gideon! Pls also have a look at the other PR I raised which, unlike the one here, is ready to be merged as it had a much smaller focus, namely to introduce a hex viewer for files rather than the C64's RAM. 

The issue I had on that other PR with not being able to open files larger than 26KiB is already resolved. It was probably due to memory exhaustion caused by duplicating the required documents buffer space. The new approach simply uses the single source of truth buffer into which the file is read, then on the fly creates its hex representation as the hex viewer traverses it. Thus, only a buffer for the current view window is needed. I also did some quite low level conversion to keep it as fast as possible.

I am facing a few issues with this PR here and I described them in the PR overview at the bottom. Maybe you have an idea how to resolve them. Many thanks!

Zibri:
A full machine code monitor like the one in VICE would be great to have.

chrisgleissner:
Thanks @Zibri , I agree. 

I am currently hoping for my first PR (https://github.com/GideonZ/1541ultimate/pull/357) to be merged by @GideonZ and will continue working on this PR once that has happened and I find time again.

xahmol:
Would quite like this feature actually, missed a monitor sorely on developing my own recent project targettiing the U64.
As my porject is build in the form of a FinalCartridge 3 subtype 1 cartridge image, I could not use a cartridge to get a monitor.

So ended up by debugging using full memory dumps (and doing PC based comparisons and analysis using a PC based hex editor)

A VICE like monitor via the Ultimate UI would be awesome.

bvl1999:
> Would quite like this feature actually, missed a monitor sorely on developing my own recent project targettiing the U64. As my porject is build in the form of a FinalCartridge 3 subtype 1 cartridge image, I could not use a cartridge to get a monitor. And could not use VICE either, as my project is heavily using the Ultimate Command Interface which VICE cannot emulate (yet?)
> 
> So ended up by debugging using full memory dumps (and doing PC based comparisons and analysis using a PC based hex editor)
> 
> A VICE like monitor via the Ultimate UI would be awesome. And if you can only get it to work on the U64, well, IMHO that would be a great first step already ;-) (also being selfish a little bit here, use my UII+ carts on my C128s, and the C128 already has an inbuild monitor you can use in most cases)

Not to mention, supporting the C128's MMU from an app running on the UII+(L) is probably not possible, because the MMU is invisible for DMA.. So such a monitor would be of limited use for the C128... but for a C64 with UII+ it would still be quite usefull. The biggest challenge I see is supporting changing the contents of $01, as it is also hidden from DMA, tho I think this is far less of a limitation for the c64.

GideonZ:
Hi Chris,

In the last months I have been primarily busy with the Ultimate64 Elite-II, so I had very little time to address activities from other users. I understand that this is discouraging, sorry about that. The good news is that I have put a bit of effort in CI/CD through github, which means that changes in a PR will be automatically built and result in an update package. Until now that's only available in the u64ii branch, but I'll merge all my changes there towards master shortly.

software/u64/u64_machine.cc
```diff
@@ -3,28 +3,70 @@
...
+    after_memory_access(0, freezerMenu);
+    return byte;
+}
+
+void U64Machine :: poke(uint32_t address, uint8_t byte)
+{
+    bool freezerMenu = before_memory_access();
+    volatile uint8_t *ram = (volatile uint8_t *)C64_MEMORY_BASE;
+
+    ram[address] = byte;
```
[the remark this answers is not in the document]
chrisgleissner: Implemented in 7c7a57e3. U64Machine::peek() and poke() now sanitize the incoming address to the 16-bit C64 address space before indexing the aperture, so out-of-range callers can no longer walk past the 64 KiB window.

software/userinterface/editor.cc
```diff
@@ -123,7 +121,7 @@ void Editor :: init(Screen *scr, Keyboard *key)
     
 void Editor :: draw(void)
 {
-    struct Line line;
+    struct Line line, *line_ptr;
```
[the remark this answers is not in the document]
chrisgleissner: Already resolved on the current branch. Editor::draw() now declares only a single local Line instance, so the unused pointer warning is gone without needing a follow-up change in this push.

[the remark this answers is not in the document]
chrisgleissner: Already resolved on the current branch. Editor::draw() now declares only a single local Line instance, so the unused pointer warning is gone without needing a follow-up change in this push.

software/u64/u64_config.cc
```diff
@@ -1463,26 +1478,42 @@ SubsysResultCode_e U64Config :: executeCommand(SubsysCommand *cmd)
...
-                C64_POKE(addr, value);
-
-                machine->resume();
-            } else {
-                C64_POKE(addr, value);
-            }
-            portEXIT_CRITICAL();
+            char msg[20];
+            sprintf(msg, "Poke(%4x,%2x)=%2x", addr, value, verified_value);
+            cmd->user_interface->popup(msg, BUTTON_OK);
```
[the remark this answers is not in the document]
chrisgleissner: This one is no longer applicable on the current branch. The ad-hoc Peek/Poke popup handlers were removed in favor of the machine monitor flow, so the sscanf/sprintf path described here no longer exists.

software/userinterface/ass_editor.cc
```diff
@@ -0,0 +1,58 @@
...
+    for (int i = 0; i < line->length; i++) {
+        uint8_t c = (uint8_t) line->buffer[i];
+        add_hex_byte(hex_buf, HEX_COL_START + (3 * i), c);
+        // represent all non-printable characters as '.' based on the character set used by firmware version 3.10j
+        hex_buf[TXT_COL_START + i] = (char) ((c == 0 || c == 8 || c == 10 || c == 13 || (c >=20 && c <= 31) || (c >= 144 && c <= 159)) ? '.' : c);
+    }
+    
+    window->output_length(hex_buf, CHARS_PER_HEX_ROW);
+    window->repeat(' ', window->get_size_x() - CHARS_PER_HEX_ROW);
+}
```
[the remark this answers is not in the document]
chrisgleissner: Already resolved in the current PR state. The erroneous ass_editor.cc implementation referenced here is no longer present, so there is no duplicate HexEditor::draw() definition left to fix.

software/io/c64/screen.h
```diff
@@ -59,6 +59,7 @@ class Screen
 
     // for character mapped screens. Does not work for VT100.
     virtual void backup(void) { }
+        virtual bool prefers_full_refresh(void) { return false; }
```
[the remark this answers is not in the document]
chrisgleissner: Fixed in 7c7a57e3 by aligning prefers_full_refresh() with the surrounding virtual methods in screen.h.

software/monitor/machine_monitor.cc
```diff
@@ -0,0 +1,1463 @@
...
+
+void monitor_transfer_memory(MemoryBackend *backend, uint16_t start, uint16_t end, uint16_t dest)
+{
+    uint32_t length = (uint16_t)(end - start);
+    if (length == 0) {
+        return;
+    }
+    if (dest > start && dest < end) {
+        while (length) {
+            length--;
```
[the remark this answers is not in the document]
chrisgleissner: Handled in 7c7a57e3 by keeping transfer/compare semantics end-exclusive and tightening the parser to reject zero-length ranges like T C000-C000,DEST. That makes the command behavior explicit instead of silently accepting a no-op range.

software/monitor/machine_monitor.cc
```diff
@@ -0,0 +1,1463 @@
...
+{
+    uint32_t length = (uint16_t)(end - start);
+    uint32_t index;
+    int count = 0;
+    int pos = 0;
+
+    out[0] = 0;
+    for (index = 0; index < length; index++) {
+        if (backend->read((uint16_t)(start + index)) != backend->read((uint16_t)(dest + index))) {
+            pos = append_line_address(out, out_len, pos, (uint16_t)(start + index));
```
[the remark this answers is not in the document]
chrisgleissner: Handled in 7c7a57e3 alongside transfer parsing: compare keeps the existing end-exclusive implementation, and the parser now rejects zero-length ranges so C start-end,dest no longer accepts C000-C000 and then reports a misleading no-op result.

software/monitor/disassembler_6502.cc
```diff
@@ -0,0 +1,268 @@
...
+    "RTI          ", "EOR ($nn,X)  ", "HLT*         ", "LSE*($nn,X)  ",
+    "NOP*$nn      ", "EOR $nn      ", "LSR $nn      ", "LSE*$nn      ",
+    "PHA          ", "EOR #        ", "LSR A        ", "ALR*#        ",
+    "JMP $nnnn    ", "EOR $nnnn    ", "LSR $nnnn    ", "LSE*$nnnn    ",
+
+    "BVC rel      ", "EOR ($nn),Y  ", "HLT*         ", "LSE*($nn),Y  ",
+    "NOP $nn,x    ", "EOR $nn,X    ", "LSR $nn,X    ", "LSE*$nn,X    ",
+    "CLI          ", "EOR $nnnn,Y  ", "NOP*         ", "LSE*$nnnn,Y  ",
+    "JMP*$nnnn  $$", "EOR $nnnn,X  ", "LSR $nnnn,X  ", "LSE*$nnnn,X  ",
+
```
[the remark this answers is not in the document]
chrisgleissner: Fixed in 7c7a57e3 by normalizing the affected opcode templates to ,X and adding a regression test that exercises the illegal zeropage,X NOP rendering path.

software/userinterface/userinterface.cc
```diff
@@ -1,6 +1,11 @@
 #include "userinterface.h"
 #include <stdio.h>
 
+#include "machine_monitor.h"
+
+#include "disassembler_6502.cc"
+#include "machine_monitor.cc"
+
```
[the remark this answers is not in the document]
chrisgleissner: Verified, but I am leaving the current include model in place for this PR because changing the monitor source integration would widen the build surface well beyond the review defects addressed here. The current branch still relies on the embedded translation-unit pattern for these monitor sources, and switching that wiring safely would be a larger follow-up than the minimal convergence work for this PR.

bvl1999:
Looks like you are making quite a bit of progress here @chrisgleissner , very nice.  How close is this in functionality for others to take a look at it?

bvl1999:
yay, nice progress indeed!
image

chrisgleissner:
@bvl1999 Hi Bart,

Thanks for your feedback, it is much appreciated. I am in spring cleaning mode (based on the season), and wanted to finally get my old PRs over the line, as this one had been sitting there for close to 3 years.

It is now very close to promoting the PR from Draft mode. I just applied some final patches, and I cannot find any more bugs. If you find any, please let me know.

If you were wondering about keyboard shortcuts, they are combination of two of the most popular machine code monitors:
- Final Cartridge V
- Action Replay III

I am now working on final clean-up to ensure unnecessary refactoring is reduced to a minimum. Hopefully that will be done by tomorrow. But at this point, it is feature complete for the MVP.

Best wishes
Christian

bvl1999:
> But at this point, it is feature complete for the MVP.

Am I correct in assuming this is only for the U64 (II) and not for the Ultimate II+ cartridges?

chrisgleissner:
@bvl1999 Yes, I am only targeting Ultimate 64 devices for now. It can always be extended, but I only own an Ultimate 64 Elite I, which I use for testing, and a C64U FE. Would you be willing to try it out on an Ultimate II+ and see which features work and which ones don't?

Btw, I just updated the PR description to be in line with the Help page.

bvl1999:
> Would you be willing to try it out on an Ultimate II+ and see which features work and which ones don't?

From what I can tell currently none (monitor isn't in the developer menu, but didn't try if I can start it with c=+o).

chrisgleissner:
Thanks @bvl1999 . Yeah, that is the problem with not owning a certain device; it makes development for it difficult. I will see if there is an easy way to at least activate the monitor so you can select it in the menu, but as it is already feature complete for Ultimate 64 devices, I think full support for U2 could then be a next step after merge.

GideonZ:
> Thanks @bvl1999 . Yeah, that is the problem with not owning a certain device; it makes development for it difficult. I will see if there is an easy way to at least activate the monitor so you can select it in the menu, but as it is already feature complete for Ultimate 64 devices, I think full support for U2 could then be a next step after merge.

@chrisgleissner Or before merge with a bit of help from the mighty guru... ;-) 
And I think it's about time that I at least send you a cart as well.

bvl1999:
> Thanks @bvl1999 . Yeah, that is the problem with not owning a certain device; it makes development for it difficult. I will see if there is an easy way to at least activate the monitor so you can select it in the menu, but as it is already feature complete for Ultimate 64 devices, I think full support for U2 could then be a next step after merge.

I did a brief review of the code (tho before your last changes).

Currently there are a number of U64 specific things in there.
- using the character rom file from /flash/rom, this will not work on a UII/UII+ because that file won't be there (of course people can put the file there if they want to, but.. the UII/UII+ do not use it). I suspect this may also be an issue for a C64U implementation, but that needs a separate discussion anyway. This could be solved by including that rom in the firmware itself, but I suspect that has some licensing issues.
- you cannot access addresses 0 and 1 as the CPU sees them on a 'real' C64, so, the selection of rom/ram banks won't work. I don't see how that could be fixed without running code on the cpu. 

Beyond that I didn't see things which wouldn't work on the UII/UII+. 

So, with a few changes, and with some restriction on functionality, this could totally work on a UII+

As to integration, no idea if you already did any of this, but I have 2 suggestions:
- move the monitor from being a subdirectory of userinterface to a subdirectory of 'software'. I think it is substantial enough to be its own thing.
- move the code which sets up the 'tasks' for the developer menu into its own monitor_init.cc as part of that directory. While I get why you stuck it into u64_config.cc, this seems to not be the best place.

Last but not least, somehow u64_config.cc and u64_config.h ended up with dos line endings, the rest of your code has unix line endings like everything else in the repo.

Bart.

GideonZ:
> * using the character rom file from /flash/rom, this will not work on a UII/UII+ because that file won't be there (of course people can put the file there if they want to, but.. the UII/UII+ do not use it). I suspect this may also be an issue for a C64U implementation, but that needs a separate discussion anyway. This could be solved by including that rom in the firmware itself, but I suspect that has some licensing issues.

Correct. This ROM is not available in the /flash/roms directly on the cartridges. Including it in the ROM is OK for the C64U (is in fact already there), but not for the U64 / U64E2. What you could do instead is a DMA read of the cartridge ROM ;-) This will work on all machines / carts.

> * you cannot access addresses 0 and 1 as the CPU sees them on a 'real' C64, so, the selection of rom/ram banks won't work. I don't see how that could be fixed without running code on the cpu.

You can't fix this through DMA unfortunately. This will neither work on the U64/U64E2/C64U. DMA access to locations 0 and 1 will always go to RAM, never to the CPU's PIO.

> * move the monitor from being a subdirectory of userinterface to a subdirectory of 'software'. I think it is substantial enough to be its own thing.

Agreed.

> * move the code which sets up the 'tasks' for the developer menu into its own monitor_init.cc as part of that directory. While I get why you stuck it into u64_config.cc, this seems to not be the best place.

Agreed. We must even consider renaming u64_config to u64_hw_config. This module shall only contain code that is U64 specific and is related to the configuration of the hardware (-registers) that configure the operation of the C64/U64 specific hardware. 

> Last but not least, somehow u64_config.cc and u64_config.h ended up with dos line endings, the rest of your code has unix line endings like everything else in the repo.

@bvl1999 The repo has different line endings, unfortunately, because some code is legacy from when I was still developing under Windows (before 2016). Git can be configured to store text files always with unix line endings, regardless of what your editor does. Would this be a solution for the future?

bvl1999:
>  What you could do instead is a DMA read of the cartridge ROM ;-) This will work on all machines / carts.

I'm not sure I understand what you wrote here. Character rom on a C64/128 is not typically visible, and without changing either $01 on a C64, or both the mmu and $01 on a C128, you can't 'DMA' from there. And $01 as well as the MMU for the C128 can't be read/written using DMA. 

Unless you run some software on the machine itself which sets up a memory map in which chargen rom is visible for DMA.

But maybe I am missing something here?

> You can't fix this through DMA unfortunately

I think this is unfixable for the cartridges without running code on the cpu. And that code would be either C64 or C128 specific, or have to detect the machine type, as this also works differently on the 64 and 128 (where the memory map is mostly determined by the mmu, and only somewhat by $01, and neither mmu or $01 are visible for DMA).

So, for a UII(+) version, I rather consider this a limitation which cannot be addressed in a good way.

> Git can be configured to store text files always with unix line endings, regardless of what your editor does. Would this be a solution for the future?

Sure. I just noticed the named files have unix line endings in the main repo, but not in this branch. It is a bit annoying to read, but no big deal really.

Bart.

chrisgleissner:
Hi @bvl1999 and @GideonZ ,

Thanks for your very insightful comments.

Here some follow-up remarks:
- As @GideonZ already mentioned, the current repo has inconsistent line endings. I am keeping all my PRs minimal-invasive (though it may sometimes not seem that way - but things should always be as simple as possible, not simpler :-) ), thus I aligned the line endings to whatever is currently in place in a certain file. If that is not needed, I would very happily use Linux file endings (LF), but I discussed with Gideon before that line ending adjustments will be done in a separate, single PR, and from that point onwards, we can all just use Linux endings consistently.
- I will relocate the source code from `software/userinterface/monitor` to `software/monitor`. I doubt there will be a need for another kind of monitor soon, and if so, we can at that point rename `software/monitor` to `software/machine_monitor` or similar. I am a big proponent in keeping things simple and avoiding over-engineering, so I hope you like this name.
- My main ambition is to get this PR over the line for the Ultimate 64 (the device I own), unless fixing it for the U2+ is simple and does not massively expand scope. That way, it could potentially find its way into the C64U firmware, and there is now a large user base for these. In general, I strongly believe in small and frequent incremental releases to production.
- I did have a lot of trouble getting all the bank switching working. And as you pointed out @bvl1999 , writing to address `$0001` from the monitor indeed does not change the banks that the CPU sees. As far as I can tell in my tests, the bank switching works now. To clarify, the CPU bank switching as implemented allows the monitor to have the world view of the CPU "as though" address `$0001` had been modified to the value set via the `O` key. Thus, you can see RAM, Kernal, Basic, I/O, or Charsets. It does not change the banks as seen by the real CPU.

Today I did some final fixes to the Load / Save functionality to ensure it neatly integrates with the rest of the file picker (not yet pushed) and I am fixing the G (Go) command. Once that is done, it's ready for promoting the PR from Draft status.

Best wishes
Christian

bvl1999:
>     * As @GideonZ already mentioned, the current repo has inconsistent line endings. I am keeping all my PRs minimal-invasive (though it may sometimes not seem that way - but things should always be as simple as possible, not simpler :-) ), thus I aligned the line endings to whatever is currently in place in a certain file. If that is not needed, I would very happily use Linux file endings (LF), but I discussed with Gideon before that line ending adjustments will be done in a separate, single PR, and from that point onwards, we can all just use Linux endings consistently.

Thing is, when I look in Gideon's version of the repository, those files have 'unix' line endings. 
But anyway, this isn't a big issue to me, I just would like things to be consistent. 

Gideon's suggestion to 'auto enforce' 'unix line endings' might work, until for some reason someone includes a dos/windows specific file of course.

Regardless, I agree, it would be better to go through the entire repo at some point and fix all the cases where dos/windows line endings are used.

> 
>     * I will relocate the source code from `software/userinterface/monitor` to `software/monitor`. I doubt there will be a need for another kind of monitor soon, and if so, we can at that point rename `software/monitor` to `software/machine_monitor` or similar. I am a big proponent in keeping things simple and avoiding over-engineering, so I hope you like this name.

I'm totally fine with that name, simple and clear... but eventually, it is Gideon's call!

> 
>     * My main ambition is to get this PR over the line for the Ultimate 64 (the device I own), unless fixing it for the U2+ is simple and does not massively expand scope. That way, it could potentially find its way into the C64U firmware, and there is now a large user base for these. In general, I strongly believe in small and frequent incremental releases to production.

I understand your take on this. My primary concern here is to not introduce avoidable things which would make it impossible to also implement this for the cartridges without that being a conscious decision.

As to the C64 Ultimate firmware, keep this in mind:  the C64U firmware is currently not open source, and that means your code can't just move there, without your explicit permission or signing over the copyright.

From what you write, I assume you'd be willing to permit this, but I think it is good to not completely ignore the formalities of the GPL in this.

>     * I did have a lot of trouble getting all the bank switching working. And as you pointed out @bvl1999 , writing to address `$0001` from the monitor indeed does not change the banks that the CPU sees. As far as I can tell in my tests, the bank switching works now. To clarify, the CPU bank switching as implemented allows the monitor to have the world view of the CPU "as though" address `$0001` had been modified to the value set via the `O` key. Thus, you can see RAM, Kernal, Basic, I/O, or Charsets. It does not change the banks as seen by the real CPU.
> 

From what I saw, you are using machine->get_cpu_port() to get around this, which works on the U64, but won't work on the UII/UII+. 


> 
> Today I did some final fixes to the Load / Save functionality to ensure it neatly integrates with the rest of the file picker (not yet pushed) and I am fixing the G (Go) command. Once that is done, it's ready for promoting the PR from Draft status.

Thank you very much for all your work on this, I am sure this contribution will be valued by many a developer.

Bart.

chrisgleissner:
Thanks @bvl1999 ,

Yeah, I am totally fine for all the code that I submit to this repo to be used for the Commodore 64 U firmware. To be perfectly honest, the prospect of making it to that repo is the sole reason why I started contributing a lot recently. There is currently no direct way to contribute to the C64U firmware, so this is my next best route. 

Also, and this is totally unrelated, but have you seen Spiffy's take on the C64U firmware being also GPL v3 since it is unlikely all of its many [authors](https://github.com/spiffycrew/Spiffy_Ultimate/blob/spiffy-master-1.1.0/AUTHORS.txt) signed a dual license agreement. I am not sure this can be discussed as part of this PR, but it is an open question I've had ever since reading this. 

In terms of adding U2+ support, I would be very happy for this PR to be merged to `master` first for the U64 side of things. It can always be changed later - none of it is set in stone. Having a monitor is better than no monitor, and every road has a beginning. This is always my take on software - as long as it's written in a clear way, with tests, it is also easy to change and extend.

Many thanks for all your constructive feedback, it is much appreciated. I hope to promote the PR from Draft status either later today or tomorrow. Just final clean-up, and addressing the comment you left earlier about relocation.

Best wishes
Christian

bvl1999:
> Also, and this is totally unrelated, but have you seen Spiffy's take on the C64U firmware being also GPL v3 since it is unlikely all of its many [authors](https://github.com/spiffycrew/Spiffy_Ultimate/blob/spiffy-master-1.1.0/AUTHORS.txt) signed a dual license agreement.

I am aware of that, but I think that discussion is a bit too big for this PR. I just wanted to make sure you are aware of this aspect, and permit such use.

chrisgleissner:
Hi @GideonZ and @bvl1999 ,

I have pushed all my changes (including the relocation to a new folder, introduction of `monitor_init.cc`, and further reduction of formatting churn) and opened the PR up for review. 

This has undergone my usual testing and I also added a video walk-through. There are also extensive unit tests as well as an E2E Python-based test that interacts with the deployed machine monitor via Telnet.

Many thanks for considering a merge.

Best wishes
Christian

bvl1999:
Yay!

This is a C128 in 128 mode with UII+

image

Nice work @chrisgleissner !

chrisgleissner:
Hi @bvl1999 ,

That is excellent, thanks for the good news!

You may like two new features I added: bookmarks and polling. I think especially the latter could become very handy when watching the I/O area. 

I also added a short video link at the top of this PR to demonstrate them.

Best wishes
Christian

bvl1999:
> I also added a short video link at the top of this PR to demonstrate them.

Nice. but... do you ever sleep??? (hides)

bvl1999:
So, found one curious behavior with the UII+ on a C64 or C128 in 64 mode. It makes sense when you realize how the freezer works, and might be hard to avoid.

Anyway, because the cartridge puts the system in ultimax mode, the system roms at a000 and e000 are not accessible, and nor is the ram below it, what you get looks like disconnected address space. 

When using telnet instead of the local interface, you do get to see whatever is currently there based on the CPU's banking configuration. 

This doesn't happen on a C128 in 128 mode because it ignores the cartridge trying to use ultimax mode, and you always get the view of memory configured in the mmu. It has its own peculiarity tho, because dma is used for reading/writing memory, the monitor will see the ram bank configured in $d506 and not the one in $d500. Also, the mmu and vdc registers are invisible to dma, so those can't be viewed or polled. 

I imagine it might be possible to switch out of ultimax mode for the dma read/write, but don't know, @GideonZ would that work?

chrisgleissner:
Hi @bvl1999 and @GideonZ ,

Thanks Bart, that makes sense. And yeah, I pulled a rate late nighter as your comments and interest inspired me. :-)

Just a thought: based on my understanding, maybe the monitor could briefly leave ultimax only around the actual DMA access, then restore the previous `C64_MODE` straight away. So the cartridge mode would not be changed permanently.

Roughly:

1. Save the current `C64_MODE` in `C64::monitor_read_memory()` / `C64::monitor_write_memory()`.
2. If it looks like ultimax, and the address is outside `$D000-$DFFF`, temporarily set `C64_MODE` to `MODE_NORMAL`.
3. Do the DMA read/write.
4. Restore the saved mode.

That might let `$A000-$BFFF` and `$E000-$FFFF` see the normal memory layout instead of the ultimax/disconnected view. I would leave `$D000-$DFFF` alone, since that seems to be where the freezer/menu path benefits from ultimax for VIC/SID/CIA access.

There are some caveats of course: it still changes cartridge mode during DMA, block reads may get a bit more expensive if this happens per byte, and it would not address the C128-in-128-mode limitations such as the `$D506` DMA bank or MMU/VDC visibility.

So this may or may not be the right fix, but it looks small enough to consider. If there is a better pattern already used in the U2/UII+ code, I am very happy to follow that instead. Please let me know.

Other than that, I pushed some final (famous last words) improvements this morning to ensure the bookmark feature is readily accessible but does not result in accidental invocations, and I improved the binary view to make it more readable and useful, with a read-only hex view on the side; similar to how the memory (hex) view has a read-only ASCII view on the side.

So unless we want to extend this further, from my perspective, it is ready to be merged.

Best wishes
Christian

software/userinterface/tree_browser_state.cc
```diff
@@ -13,6 +13,72 @@
...
+        return true;
+    }
+};
+
+void prepend_picker_entry(IndexedList<Browsable *> *children)
+{
+    if (!children) {
+        return;
+    }
+    children->append(new SyntheticPickerEntry());
```
GideonZ: Interesting.. I saw that you created a 'prepend' function in the IndexedList, and here you seem to do it manually.

bvl1999:
> * Save the current `C64_MODE` in `C64::monitor_read_memory()` / `C64::monitor_write_memory()`.
> 
> * If it looks like ultimax, and the address is outside `$D000-$DFFF`, temporarily set `C64_MODE` to `MODE_NORMAL`.
> 
> * Do the DMA read/write.
> 
> * Restore the saved mode.

I think it should temporary set the mode to what it was before 'freezing' the system to do DMA reads for $8000-$ffff, to also support having a .crt loaded. 

As to the limitations in 128 mode, there is no way around that without making modifications to the hardware, or running software on the cpu. I wouldn't worry about that part.

software/userinterface/tree_browser_state.cc
```diff
@@ -13,6 +13,72 @@
 
 IndexedList<Browsable *>emptyList(0, NULL);
 
+namespace {
+
+static const char c_picker_create_file_label[] = "<Create File>";
+
+class SyntheticPickerEntry : public Browsable
```
GideonZ: There may not be a need for a SyntheticPickerEntry: there is a BrowsableStatic already defined.
naming convention: This should be "BrowsableSyntheticPickerEntry"

software/userinterface/ui_elements.cc
```diff
@@ -131,11 +173,11 @@ void UIPopup :: deinit()
 	delete window;
 }
 
-UIStringBox :: UIStringBox(UserInterface *ui, const char *msg, char *buf, int max) : UIObject(ui), message(msg), edit(buf, max)
+UIStringBox :: UIStringBox(UserInterface *ui, const char *msg, char *buf, int max, bool template_mode) : UIObject(ui), message(msg), edit(buf, max, template_mode)
```
GideonZ, line 176: Note that there are updates on the String Edit in the feature_ftp branch.

GideonZ (commented):
By adding a file picker in the tree browser, there are now two 'pickers'; this one and the one that was sneakily added in the feature_ftp branch, where the "Move to.." and "Copy to.." features were added. I think we should settle on one picker and not have two different ones.

Whether you need a picker for the directory within the monitor is questionable. You could just enter the monitor when you are in the right directory.

software/userinterface/tree_browser_state.cc
GideonZ: It might be super quirky, but I might have tried to have an virtual function that would be called that adds the header, which wouldn't do anything in the base class, but it would in a derived class. Maybe I am a bit of a "the type-system will fix this for you" kind of guy, avoiding existing classes to be peppered with additional 'features'.

software/u64/u64_machine.cc
```diff
@@ -3,28 +3,360 @@
...
-    } else { // if we were in freezer menu, the backup of the 1K-4K RAM should be used
-        // restore memory
+    } else if (pb) {
+        // if we were in freezer menu, the backup of the 1K-4K RAM should be used to restore memory
         memcpy(pb + 1024, screen_backup, 1024);
         memcpy(pb + 2048, ram_backup, 2048);
     }
 }
 
+bool U64Machine :: begin_monitor_session()
```
GideonZ: Contamination of montitor functions in abstract machine class.

software/u64/u64_machine.cc
```diff
@@ -3,28 +3,360 @@
 #include "FreeRTOS.h"
 #include <string.h>
 
-void U64Machine :: get_all_memory(uint8_t *pb)
+namespace {
+
+static uint8_t read_frozen_byte(volatile uint8_t *ram, bool freezerMenu, uint32_t address,
```
GideonZ, line 8: If you already have a derived class U64Machine (from C64Machine), why would you stlil need a memory backend class?

software/u64/u64_config.h
```diff
@@ -133,6 +134,7 @@ class U64Config : public ConfigurableObject, ObjectWithMenu, SubSystem
     void create_task_items(void);
     void update_task_items(bool writablePath);
     SubsysResultCode_e executeCommand(SubsysCommand *cmd);
+    static SubsysResultCode_e S_run_monitor(SubsysCommand *cmd);
```
GideonZ: I think there are ways to avoid this contamination. Monitor functions in a class that is purely for u64 hardware configuration is not desirable.

That said, I understand that the functions to control the ESP32 state don't belong here either. You know quick and dirty test code that makes it into master...  *hits himself on the head*

software/userinterface/context_menu.cc
```diff
@@ -45,6 +45,13 @@ ContextMenu :: ~ContextMenu(void)
 
 int ContextMenu :: get_items(void)
 {
+    FileInfo *info = contextable ? contextable->getFileInfo() : NULL;
+    if (state && state->browser && contextable &&
+        (state->browser->pick_mode != TreeBrowser::PICK_NONE) &&
+        info && !(info->attrib & (AM_DIR | AM_VOL)) &&
+        !contextable->isSyntheticPickerEntry()) {
+        actions.append(new Action("Select File", SUBSYSID_USER_STANDARD, 0));
```
GideonZ: This is contamination. The items shown in a context menu should depend on the type. I totally understand the reason behind this, because it's a pain to have it done in the clean way; but I think the only good way to do it is to populate the browser not with "BrowsableDirEntry" types, but with "BrowsablePickerEntry" types; which simply have a different implementation for "get_context_items" or whatever the function is called.

software/userinterface/task_menu.h
```diff
@@ -31,4 +39,57 @@ class TaskSubMenu : public ContextMenu
...
+            }
+        }
+    }
+
+    for(int i=0;i<objects->get_elements();i++) {
+        (*objects)[i]->update_task_items(writablePath);
+    }
+}
+
+inline Action *TaskMenu :: find_task_action(int subsysId, const char *actionName)
```
GideonZ: This function should not need to exist.

software/userinterface/tree_browser.cc
```diff
@@ -157,7 +180,21 @@ int TreeBrowser :: poll(int sub_returned)
             // create a return value of a GUI object, and call execute
             // with that immediately.
             state->draw();
-            ret = contextMenu->executeSelected(state->browser->getPath());
+            Action *selectedAction = contextMenu->getSelectedAction();
```
GideonZ: Brrrrrrrrr

software/userinterface/ui_elements.cc
```diff
@@ -46,20 +66,42 @@ void UIPopup :: init()
...
     screen->backup();
-    window = new Window(screen, x1, y1, window_width+2, 5);
+    window = new Window(screen, x1, y1, window_width+2, window_height);
     window->clear();
     window->draw_border();
     // window->no_scroll();
-    window->move_cursor(x_m, 0);
     window->set_color(get_ui()->color_fg);
-    window->output(message.c_str());
+    msg = message.c_str();
```
GideonZ, line 83: Multiline string box... noice!

target/u2/riscv/ultimate/Makefile
```diff
@@ -89,6 +89,7 @@ SRCS_CC	 =	small_printf.cc \
 			sdcard_manager.cc \
 			size_str.cc \
 			userinterface.cc \
+			run_machine_monitor.cc \
```
GideonZ, line 92: Mental test for clean integration:
If I were to just exclude the monitor-related files, would it still build?  (it should)

GideonZ (commented):
Some additional notes

bvl1999:
> . I think we should settle on one picker and not have two different ones.

Agree.

Bart.

chrisgleissner:
Hi @bvl1999 and @GideonZ,

Thanks for the review comments. I believe I have addressed them all and will reply to the inline comments separately.

---

After further investigation and refactoring, I ended up consolidating the picker/browser flow around `TreeBrowser`.

The monitor load/save flow now stays on `TreeBrowser`, and I rewrote `path_picker.cc` (cherry picked from the `feature_ftp` branch) as a `TreeBrowser`-backed wrapper instead of maintaining a second manual browsing implementation. I kept `path_picker.h` effectively unchanged, except for removing a few now-unused internal helpers, to make a later merge of `feature_ftp` simpler.

The goal was to preserve the useful superset of both paths while giving the user a consistent browsing experience, whether they enter the file browser directly or through a picker.

For the monitor, this keeps the existing browser behaviour that load/save needs, including availability of all F5 tasks (e.g. to create drive files to save memory dumps in), visible existing files during save, and drive-file traversal for saving memory regions into drive files. The save-specific behaviour remains localized in monitor-owned `Browsable` types rather than being pushed into generic browser code. It also ensure the file context menu remains available, which is useful for previewing file contents with the Hex Viewer before loading them into memory.

For `UIPathPicker`, the public API is preserved, including the synthetic “select this folder” row and printable-key fallback, but navigation, selection state, and redraw handling now come from `TreeBrowser`.

I also verified this to work on the `feature_ftp` branch under the assumption that this branch here is merged to `master` first and then files `path_picker.cc` and `path_picker.h` replace the two files of the same name on the `feature_ftp` branch. Here is a screenshot of the `feature_ftp` branch working with this consolidated `TreeBrowser`-based path picker approach:

Screenshot 2026-05-05 08-28-32

The same `TreeBrowser`-based picker is seen here used by the branch of this machine code monitor PR, allowing the user to load files into the machine monitor both from Ultimate files as well as from D64 (or other) disk files, in both cases consistently using the a new "Select" context-sensitive menu entry:

Screenshot 2026-05-05 09-05-52

Screenshot 2026-05-05 09-06-02

---

@bvl1999 I incorporated your suggested changes for ensuring cartridge, Basic and Kernal areas can be viewed on the U2. I could not test this as I only own a U64 and C64U. You can see the changes in `c64.cc` lines 621-626. Thanks for giving it another try.

---

Thanks again for reviewing. I hope this is now in better shape for merge.

Best wishes
Christian

software/userinterface/tree_browser_state.cc
chrisgleissner: Agreed. I removed the generic prepend-style approach. The synthetic save row is now introduced through monitor-owned Browsable types and makeChildEntry(...), so the generic browser code no longer needs to reshuffle entries just to support this flow.

Also, I reworked this in the type-driven direction you suggested. The generic side now only exposes narrow hooks (pickAsCurrentPath() and makeChildEntry(...)), and the monitor-specific save behavior lives in monitor-owned derived types instead of in generic browser control flow.

software/userinterface/tree_browser_state.cc
chrisgleissner: Agreed. The generic SyntheticPickerEntry approach is gone. The synthetic save row now lives in monitor-local entry types instead of becoming a generic browser concept.

software/userinterface/ui_elements.cc
chrisgleissner: Not sure if there is work to be done here. I had a look at the `feature_ftp` branch but could not see anything in `ui_elements.cc` that would cause merge conflicts.

software/u64/u64_machine.cc
chrisgleissner: I agree with that concern and kept only the generic machine primitives. C64 now exposes peek() / poke() instead of monitor-named memory APIs, and U64Machine exposes generic stopped-session control. The monitor-specific policy stays in software/monitor/, so the machine classes are cleaner than before even though the shared backend seam remains.

software/u64/u64_machine.cc
chrisgleissner: I tried removing the backend layer locally, and after much refactoring (now reverted), it turned out worse overall: production no longer used a shared seam, but the tests immediately needed a parallel fake-target seam again. I therefore restored one shared `MemoryBackend` contract that both production and tests use.

I do not think of it as a straight-through wrapper around `U64Machine`. On U64 it owns ROM shadowing, freeze/stopped-session handling, CPU/VIC bank policy, source naming, and monitor poll rate. On the cartridge targets it intentionally exposes a smaller capability set. 

In the host tests it gives the monitor the same seam without target hardware. That keeps the machine classes generic while keeping the monitor-specific policy out of them.

The following class diagram illustrates this further:

```
       +----------------+
       | MachineMonitor |
       +-------+--------+
               |
               | (uses)
               v
       +----------------+
       |  MemoryBackend |
       +-------+--------+
               ^
               | (inherits)
   ____________|__________________________________________
  |            |                |                         |
+----------+ +----------+ +------------+ +---------------------------+
|   U64    | |    U2    | |    Fake    | |           Fake            |
|  Memory  | |  Memory  | |   Memory   | |          Banked           |
|  Backend | |  Backend | |   Backend  | |      MemoryBackend        |
+----------+ +----------+ +------------+ +---------------------------+
```

software/userinterface/context_menu.cc
chrisgleissner: Agreed. `context_menu.cc` no longer carries the monitor-specific picker injection. The save behavior is now type-driven through monitor-owned `Browsable` entries instead of a hardcoded context-menu branch.

software/userinterface/task_menu.h
chrisgleissner: Agreed. I removed `TaskMenu::find_task_action()`. The monitor action is now registered once and retrieved directly from the monitor registry, so TaskMenu no longer needs a name-based search helper for this.

software/userinterface/tree_browser.cc
chrisgleissner: I narrowed that area substantially: the save behavior is now expressed through generic hooks plus monitor-owned entry types, and the direct monitor-launch path in `tree_browser.cc` now only activates when the monitor registry is actually linked. So the generic browser code no longer carries the earlier monitor-specific save-flow pollution.

target/u2/riscv/ultimate/Makefile
chrisgleissner: I tested your “exclude the monitor files” concern by building `target/u2plus/nios/recovery`, which still compiles even though it does not link the monitor objects.

I then reran the full firmware build as well, and that still succeeds for `u64`, `u64ii`, and `u2`.

software/u64/u64_config.h
chrisgleissner: Agreed. I removed the monitor callback declaration from `u64_config.h`. The launch helper is now file-local in `u64_config.cc`, and the task registration goes through the monitor registry instead of a header-level callback declaration.

chrisgleissner:
I applied some further minor improvements:
- Created [documentation](https://github.com/chrisgleissner/1541ultimate/blob/feature/151-machine-code-monitor/doc/machine_code_monitor.md) in the `doc` folder which expands on the Help (**F3/?**) screen that's bundled with the machine monitor.
- **W**(idth) now toggles the byte width (number of bytes per row) for both **B**(inary) (1->2->3->3S->4) and **M**(emory) views (8->16). 
- Rendering of ASCII and Screen views was improved, supporting better mapping of the two available C64 chargen ROM character sets into the glyphs available to the firmware menu. Details of the mapping were added to the `machine_monitor.md` documentation.
- The ASCII view now supports both lower- and upper-case characters, as does the `H`(unt) search
- Extended E2E test

E2E test results:
```
chris@mickey:~/dev/c64/1541ultimate/tools/developer/machine-code-monitor$ ./monitor_test.py 
[01] initial CPU7/KERNAL monitor status ... OK
[02] telnet blocks poll mode ... OK
[03] KERNAL $E000 hex view and REST match ... OK
[04] paging away and back keeps memory view stable ... OK
[05] KERNAL disassembly formatting ... OK
[06] KERNAL $E010 REST match ... OK
[07] CPU6 RAM under BASIC write/read ... OK
[08] CPU5 RAM under KERNAL status ... OK
[09] ASCII view width and scrolling ... OK
[10] ASCII and Screen mapping semantics ... OK
[11] HEX edit writes both nibbles ... OK
[12] CPU bank cycling reaches CHAR and RAM mappings ... OK
[13] COMPARE reports differing address ... OK
[14] G executes finite loop and returns to monitor ... OK
[15] G repeated execution updates RAM sentinel ... OK
[16] G handoff preserves stable VIC state ... OK
[17] bookmarks recall, set, list, and label edit ... OK
[18] memory bookmark jump restores width 16 ... OK
[19] binary width cycling and bookmark jump restores width 4 ... OK
monitor_test: OK (19 checks)
```

software/monitor/monitor_init.cc
```diff
@@ -0,0 +1,77 @@
...
+#include "userinterface.h"
+#if U64
+#include "u64_machine.h"
+#include "u64_memory_backend.h"
+#elif !defined(RECOVERYAPP)
+#include "u2_memory_backend.h"
+#endif
+
+namespace {
+static Action *machine_monitor_actions[256];
```
GideonZ, line 21: Hmm?

software/monitor/monitor_init.cc
```diff
@@ -0,0 +1,77 @@
...
+    dev->prepend(monitor);
+    if ((subsys_id >= 0) && (subsys_id < 256)) {
+        machine_monitor_actions[subsys_id] = monitor;
+    }
+    return monitor;
+}
+
+Action *get_machine_monitor_task(int subsys_id)
+{
+    if ((subsys_id < 0) || (subsys_id >= 256)) {
```
GideonZ, line 61: This doesn't seem to be congruent with how subsys_ids are intended? What is the purpose?

software/test/stubs/userinterface.h
```diff
@@ -69,6 +69,16 @@ class UserInterface
         return strlen(buffer);
     } // blocking
 
+    int  string_box(const char *msg, char *buffer, int maxlen, bool)
+    {
+        return string_box(msg, buffer, maxlen);
+    } // blocking
+
+    int  string_box(const char *msg, char *buffer, int maxlen, bool, bool)
```
GideonZ, line 77: hmm? What is the purpose of having a string box with parameters that are not used?

software/u64/u64_machine.cc
GideonZ, line 8: Hmm... Yes indeed. :-)  Clear to see that I never tested very much with the freeze menu, as I hardly ever use it on the U64. Good catch.

software/u64/u64_machine.cc
```diff
@@ -3,28 +3,360 @@
...
+            return;
+        } else if ((address >= 2048) && (address < 4096)) {
+            ((uint8_t *)ram_backup)[address - 2048] = value;
+            return;
+        }
+    }
+    ram[address] = value;
+}
+
+static void override_cpu_port(volatile uint8_t *ram, bool freezerMenu, uint8_t cpu_port,
```
GideonZ, line 37: This won't work, unfortunately. The CPU port is not accessible through DMA.

software/u64/u64_machine.cc
```diff
@@ -3,28 +3,360 @@
...
+    C64_DMA_MEMONLY = 0;
+    uint8_t value = read_frozen_byte(ram, freezerMenu, address, screen_backup, ram_backup);
+    C64_DMA_MEMONLY = saved_memonly;
+    return value;
+}
+
+static void write_visible_byte(volatile uint8_t *ram, bool freezerMenu, uint32_t address, uint8_t value,
+                               uint32_t *screen_backup, uint32_t *ram_backup)
+{
+    uint8_t saved_memonly = C64_DMA_MEMONLY;
```
GideonZ, line 72: Check if this register is readable.
NO.. This register is NOT readable, so saved_memonly will always return 0.

software/u64/u64_machine.cc
GideonZ, line 1: There seem to be a lot of clunky memory mapping mode, PLA emulation stuff in here, which is hard to test for correctness and completeness. Shouldn't we take a step back and think of ways to make the system behave AS IF the CPU was accessing these locations?

Now it's like this
```
* CPU -> PIO ----------------------------\
*       CPU -> bus -> |\                  |
*       VIC -> bus -> | \                 v
* DirectDMA -> bus -> | arbitration -> decode (PLA) -> CPU view
* ExternDMA -> bus -> |/                  ^
* External Pins (EXROM/GAME) ------------/
```
DirectDMA has an extra flag that influences the decode block to always ends up in RAM. IT's mostly the external pins that may put sud into the game.

It may be better to feed a virtual set of EXROM/GAME and PIO pins into the bus subsystem at the DirectDMA port,which are optionally different from what the levels of "system level" pins are. This way every bus agent can have a different view on the memory, even in a neatly interleaved way. What do you think? (in other words: pull the external and pio pins before arbitration and make them programmable for DirectDMA, which is driven from the application.)

software/userinterface/browsable.h
```diff
@@ -59,11 +61,14 @@ class Browsable
...
-    virtual bool getSelection() { return selected; }
-    virtual bool isSelectable() { return selectable; }
-    virtual void allowSelectable(bool b) { selectable = b; }
-    virtual int  getSortOrder(void) { return 0; }
+    virtual void setSelection(bool s) { selected = s; }
+    virtual bool getSelection() { return selected; }
+    virtual bool isSelectable() { return selectable; }
+    virtual void allowSelectable(bool b) { selectable = b; }
+    virtual FileInfo *getFileInfo() { return NULL; }
+    virtual bool pickAsCurrentPath() { return false; }
```
GideonZ, line 69: I see 'browsable' as any line in the line-based browser. This does not necessarily mean that it represents a file, and therefore getFileInfo is rather strange for an abstract browsable. I might have sinned against this idea elsewhere, but maybe this requires a bit of thought. I can also pick this up; you don't need to fix this now. (WE might not need to fix this now...)

software/userinterface/context_menu.cc
```diff
@@ -33,23 +34,35 @@ ContextMenu :: ContextMenu(UserInterface *ui, TreeBrowserState *state, int initi
...
-    if(contextable) {
-        contextable->fetch_context_items(actions);
-    }
-    return actions.get_elements();
-}
+int ContextMenu :: get_items(void)
+{
+    if(contextable) {
+        contextable->fetch_context_items(actions);
+        if (state && state->browser && state->browser->pick_mode != TreeBrowser::PICK_NONE &&
```
GideonZ, line 55: The context menu is now dependent on some variable in the tree browser, while the idea was to only make it dependent on the browsable. This would have been fixed by having a 'fetch_context_items' function of the type BrowsablePickEntry.

software/userinterface/path_picker.h
GideonZ, line 1: Why does this exist? Could it not be the tree browser itself?  (Forgive me for the naming of the objects. TreeBrowser is very abstract, but yes, it can be used for things other than files.)

GideonZ (commented):
Very nice piece of work. I'll merge it soon. Let's talk about whether and if so how the abstraction can be tightened around the picker.

I like the use of C64_MODE_NMI. It was defined, but never used!! ;-) I don't think it works when the machine is in External Slot mode, though. We might need to revisit this once we have breakpoints and what not.

chrisgleissner:
Thank you very much for merging it @GideonZ , this was much appreciated indeed.

If you'd like me to address the final review comments (thanks!) that you left 3 days ago, just let me know, and I could address them on a new branch created from `master`.

Best wishes
Christian

GideonZ:
@chrisgleissner Hi Christian! I think it is always good to take the review comments into account and maybe put what's relevant into a new issue, so that it can be planned and tracked. 

Note that I have not yet merged it to master. I have merged it to "test-merge", which is a temporary branch in which most of the new features come together.

* I noticed that there were two versions of the PathPicker and there was a UIPathBox. I ditched those, since the path picker was not referenced anywhere.
* The FTP feature, that also carried the "Move to" and "Copy to" functions still has to be fixed. I made it mergable by disabling the path picker for these functions.
* Once I have used your TreeBrowser-based path picker to re-enable these "Move/Copy to" functions, I will merge to master as one congruent whole.
* I squished the commits from your branch into one commit, in order to make it easier later on to cherry-pick this whole development into the Commodore branch. I did the same for the HTTP target feature. Once the FTP feature is congruent, I will do the same with the FTP feature as well.

chrisgleissner:
Thanks @GideonZ , I fully agree. I'll create an issue shortly and will link it here. In case I did not capture your intent correctly in the Github issue, just comment on it and I will incorporate your comments into the issue (or you can of course change it directly unless Github somehow blocks original posting changes by anyone but the author).

chrisgleissner:
@GideonZ I raised https://github.com/GideonZ/1541ultimate/issues/696

PR #370, opened 2023-08-11
Zibri fixes 2

Zibri:
This includes the right C64 pal clock


Zibri:
can you explain what's wrong with this?

GideonZ:
I already explained many times.

Zibri:
You just said the int conversion won't change. I wonder why then keep the wrong value in the code.

GideonZ:
1. It is not wrong, it is well within the accuracy range of the crystal.

2. It doesn't make any difference, so why change it?

Zibri:
1. it is wrong. In any manual I found the speed is stated as 985248.
2. Just to reflect the right value.

References:

-  https://www.lemon64.com/forum/viewtopic.php?t=36751
- http://forum.6502.org/viewtopic.php?f=4&t=4967
- https://codebase64.org/doku.php?id=base:cpu_clocking
- http://hitmen.c02.at/temp/palstuff/
- https://cyknh.schmitt-christian.com/c64-clock-rate/35505958

Zibri:
To be honest, on a couple of pages I found both values.. but only in those 2 pages...

HowlinAlan:
OMG, just let it go. It should be obvious to you, like it is to everyone else, that Gideon is NEVER going to make this change, not even if god himself rode down on a lightning bolt from heaven with the “correct” value etched into stone tablets. Never, ever. You’ve beat this horse into a watery puddle. Please move on. On Aug 12, 2023, at 1:35 PM, Zibri ***@***.***> wrote:﻿

it is wrong. In any manual I found the speed is stated as 985248.
Just to reflect the right value.

References:

https://www.lemon64.com/forum/viewtopic.php?t=36751
http://forum.6502.org/viewtopic.php?f=4&t=4967
https://codebase64.org/doku.php?id=base:cpu_clocking
http://hitmen.c02.at/temp/palstuff/
https://cyknh.schmitt-christian.com/c64-clock-rate/35505958

Zibri:
>  Please move on. 
LOL! I moved on so much that I am making my own fork.
At this point I don't really care if he is gonna make the changes I propose or not.
More will follow. And I will use them.
My bad in assuming this was not a toxic environment as it seems to be.
I don't care anymore, really.
I easily put up a few build environments for this and vice emulator.
That's the beauty of opensource.

GideonZ:
Maybe you could read up on basic accuracy theory. You see, both values are equally valid.

If a crystal has a 30 ppm accuracy, a variation of 30 Hz would be exptectes on a 1MHz clock. In this case, the accuracy is better, because it is trimmed by an adjustable capacitor near the VIC. But 2 Hz on 1 MHz is only 2 ppm! That is super accurate! (Not even to mention aging and stability here...)

The main reason for me not to accept this change is that I have explained to you various times why this will not result in any different behavior of the FPGA. EXACTLY ZERO difference. Why waste time and effort on this? Accepting this change would suggest in the change log that something has changed, while it hasn't. Why is this so hard for you to accept? Why is this so important to you, that you choose to push on this over and over again and make people irritated with your attitude? *What is your gain, Zibri? What do you win? Absolutely nothing!*

I know you are a smart guy. You don't need to prove it. Contributing to an open source project means that you make the project grow, not that you try to make yourself important. Let's focus on things that matter. If you find things that are wrong, I am willing to listen, also to you. But please, be humble, and avoid unnecessary noise that wastes other people's time. Some days ago, you made me chase another ghost because of some wrong assumptions. I will never claim that the ultimate is flawless, but please, let's agree that you check your assumptions before posting.

mrdudz:
> Some days ago, you made me chase another ghost because of some wrong assumptions

It's a common pattern, also with VICE. Don't let him drag you into it :)

Zibri:
<< try to make yourself important >>   I don't need that. Really.
I wouldn't even consider myself "more important" just because I contributed to some project.
You really got me all wrong.
Perhaps is the way I write, but those are absolutely not my intentions.
If you would just unblock me we could discusso some things instead of chasing ghosts.
If you don't want nor need my "help" then ok.
About the U2+ I bought it because os of now it's the best product around for the C64.
I never "belittle" it. I even talked about it with many friends and a few of them bought it from you because of my advice.
I understand there are "space limitations" and as in C64 software, tradeoff must be made.
But sometimes, a single small modification can make a big difference.

PR #380, opened 2023-11-09
Add "zicsr" to fix build issue

### LINKED ISSUE #379: "zicsr" build issue with new toolchain
I'm attempting to build the U2+L firmware for the first time and it fails with the following errors:

```
Compiling tasks.c
../../../../software/FreeRTOS/Source/tasks.c: Assembler messages:
../../../../software/FreeRTOS/Source/tasks.c:2094: Error: unrecognized opcode `csrc mstatus,8', extension `zicsr' required
../../../../software/FreeRTOS/Source/tasks.c:4221: Error: unrecognized opcode `csrc mstatus,8', extension `zicsr' required
../../../../software/FreeRTOS/Source/tasks.c:4259: Error: unrecognized opcode `csrs mstatus,8', extension `zicsr' required
../../../../software/FreeRTOS/Source/tasks.c:2039: Error: unrecognized opcode `csrc mstatus,8', extension `zicsr' required
../../../common/rules.mk:167: recipe for target 'tasks.o' failed
```

scjody:
Fixes https://github.com/GideonZ/1541ultimate/issues/379, tested on a U2+L with `rv32i-131023` (GCC 13.2.0) from https://github.com/stnolting/riscv-gcc-prebuilt

### DIFF
--- a/target/u2plus_L/riscv/bootloader/Makefile
+++ b/target/u2plus_L/riscv/bootloader/Makefile
@@ -3,7 +3,7 @@ CROSS	     = riscv32-unknown-elf-
 ELFTYPE      = elf32-littleriscv
 ARCHITECTURE = riscv
 # CPU architecture and ABI
-MARCH ?= rv32i
+MARCH ?= rv32i_zicsr
 MABI  ?= ilp32
 
 include ../../../common/environment.mk
--- a/target/u2plus_L/riscv/ultimate/Makefile
+++ b/target/u2plus_L/riscv/ultimate/Makefile
@@ -3,7 +3,7 @@ CROSS	     = riscv32-unknown-elf-
 ELFTYPE      = elf32-littleriscv
 ARCHITECTURE = riscv
 # CPU architecture and ABI
-MARCH ?= rv32i
+MARCH ?= rv32i_zicsr
 MABI  ?= ilp32
 
 include ../../../common/environment.mk
--- a/target/u2plus_L/riscv/updater/Makefile
+++ b/target/u2plus_L/riscv/updater/Makefile
@@ -3,7 +3,7 @@ CROSS	     = riscv32-unknown-elf-
 ELFTYPE      = elf32-littleriscv
 ARCHITECTURE = riscv
 # CPU architecture and ABI
-MARCH ?= rv32i
+MARCH ?= rv32i_zicsr
 MABI  ?= ilp32
 
 include ../../../common/environment.mk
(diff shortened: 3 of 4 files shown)


markusC64:
You shouldn't need that patch. Just use a proper toolchain.

git clone https://github.com/riscv/riscv-gnu-toolchainriscv-gnu-toolchain-rv32i
cd riscv-gnu-toolchain-rv32i
git submoduleupdate --init --recursive
mkdir build; cd build
../configure--with-arch=rv32i_zicsr --with-abi=ilp32 --prefix=/opt/riscv32i_ilp32

( note that the "prefix"on the configure command is just a directory, not to confuse with the prefixof the cross compiling toolchain )
make


Using that toolchain, the option in the makefile isn't necessary at all.

uffejakobsen:
IMHO I think that the change request is reasonable
The change would enable use of pre-built riscv gcc toolchain packages - many linux distributions already ship these...

As an alternative it should be possible to supply `make MARCH=rv32i_zicsr` and thereby overload the contents of the existing `Makefile`

markusC64:
Probably correct. On the other hand, in order to create merge requests with improvements, its better to have a toolchain that works with current sources.

GideonZ:
Well, in this case it is progress made on the definition of the RiscV
infrastructure. zicsr is no longer part of the RV32I definition, so in
fact, it is correct to add it to the MARCH.

GideonZ:
I tested this with my version of the compiler (riscv32-unknown-elf-gcc (GCC) 10.2.0), and this architecture specification does not work yet. So I'll leave it like this for now.

rgc2000:
I am using a tool chain which needs this "fix" but I don't need to change the Makefiles. I have declared in my .profile (or .bashrc if you prefer) the line :

`export MARCH=rv32i_zicsr_zifencei`

This way the Makefile will use the env variable instead of the default value (that's the meaning of ?= in the Makefile)

chrisgleissner:
Hi @rgc2000 and @GideonZ ,

is this PR still relevant. I am able to build just fine for the U2+L, plus there have not been any updates to this PR in 2.5 years. I suggest we close it.

Thanks
Christian

chrisgleissner:
Hi @scjody,

Thanks for the contribution and the discussion around it.

This PR has been open since 2023 without further movement. As the build setup needs to be considered across the current toolchain and all supported targets, it would be better to revisit this as a fresh, scoped change if there is still a need.

I’ll close this PR for now to keep the open PR list manageable. Please feel free to reopen it or submit an updated PR at any time.

Thanks again.

PR #382, opened 2023-11-13
Close telnet socket with lwip_close

### LINKED ISSUE #364: repeated cycles accessing telnet and FTP will fail
After 10 cycles of FTPing a file to the storage and then running it from the Telnet server, FTP will suddenly become unresponsive. You can trigger it by repeating a command line like

`curl -T u2rtc.prg ftp://bil/Temp/u2rtc.prg ; echo x | nc bil 23`

10 times. On the 11th time, the FTP portion will time out until the cartridge is powered off and back on. Ultimate-II Plus 3.10f (11D) on a U2+L.

scjody:
Fixes #364

### DIFF
--- a/software/network/socket_stream.cc
+++ b/software/network/socket_stream.cc
@@ -111,7 +111,6 @@ int SocketStream :: transmit(const char *buffer, int out_length)
 void SocketStream :: close()
 {
 	if(actual_socket > 0) {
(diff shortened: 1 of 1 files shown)


GideonZ:
This was already added... Thank you for the great find.

PR #384, opened 2023-11-24
Top and bottom margins are now user configurable

rgc2000:
You can change the top margin and the printable area size in the printer configuration menu.
Left and right margins are still not configurable.


rgc2000:
Wrong target for pull request

PR #440, opened 2024-12-04
Fix some keyboard shortcut bugs in popups

Gee-64:
The 'a' key now properly means 'All' and 'c' means 'Cancel'.

The bug was introduced 2010-07-25 in commit 2f5cbc58b.

The bug was likely not found or fixed since very few dialogs use these buttons and most people do not use the keyboard shortcuts. Additionally most of the Cancel/All buttons are in the updaters rather than the normal firmware.

Easy method to see the bug in action (will wipe your config!):

 1) Enter the menu by pressing the cartridge middle button
 2) Press F5
 3) Select "Configuration"
 4) Select "Clear Flash Config"
 5) Confirm with "Yes"
 6) In the "Cold Boot Required!" dialog you can choose between Ok and Cancel.
    Pressing "c" does not cancel, but pressing "a" does.

### DIFF
--- a/software/userinterface/ui_elements.cc
+++ b/software/userinterface/ui_elements.cc
@@ -88,21 +88,24 @@ int UIPopup :: poll(int dummy)
     if (c == -2) // error
     	return -1;
 
-    for(int i=0;i<btns_active;i++) {
+    int i;
+    int selected_button = -1;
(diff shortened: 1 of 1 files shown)


GideonZ:
HI Gee! :)

I think only the c and a needed to be reversed for the problem to disappear. In fact, I already received a merge request just days ago for this.

PR #444, opened 2024-12-16
Optimize the exFAT free disk space calculator

Gee-64:
The free disk space calculation done when pressing F4 in the menu on a UII+L is slow in certain circumstances.

With this optimization a 128GB USB stick with 512 byte sectors and 4k cluster size (4MByte bitmap) sees an 82% time reduction, from 18545ms to 3100ms.

Using USB disks/sticks with small cluster size avoids wasting disk space when storing large amounts of small files (.d64, .prg). For exFAT formatted disks however, the "free disk space" calculation is quite slow when using a small cluster size. The first time F4 is pressed after power on, the user interface freezes with no visual feedback until the calculation completes. Further presses of F4 are fast since the result is cached.

Optimizations:

 - Read 4k bytes instead of 512 byte sectors (uses an existing 4k buffer)
 - Pre-calculate a lookup table to avoid expensive bit operations
 - Unroll the inner loop slightly (optimal unroll is 4x)

Cost: About 300 bytes of additional RiscV code and roughly 256 bytes of stack

Futher performance gains could be made by increasing the read buffer. Some measurements on the same 128GB stick:

 - 3100ms - 4k buffer (current)
 - 2705ms - 8k buffer:  -13%
 - 2575ms - 12k buffer: -17%
 - 2515ms - 16k buffer: -19%
 - 2455ms - 24k buffer: -21%
 - 2430ms - 32k buffer  -22%
 - 2585ms - 40k buffer (worse!)
 - 2685ms - 48k buffer (worse!)

The additional buffer space can either be permanent or temporary (malloc+free). The latter would require adding a small API function to ff.[ch].

Not increasing buffer space for now, but this can certainly be revisited if we want to trade memory for additional speed in the future.

### DIFF
--- a/software/chan_fat/ff14/source/ff.c
+++ b/software/chan_fat/ff14/source/ff.c
@@ -4824,18 +4824,85 @@ FRESULT f_getfree (
 
 					clst = fs->n_fatent - 2;	/* Number of clusters */
 					sect = fs->bitbase;			/* Bitmap sector */
-					i = 0;						/* Offset in the sector */
-					do {	/* Counts numbuer of bits with zero in the bitmap */
-						if (i == 0) {
-							res = move_window(fs, sect++);
-							if (res != FR_OK) break;
+
+					/* Prepare a lookup table to speed up free cluster calculation */
+					BYTE lookup[256];
+					BYTE num_zero_bits;
+					for (i = 0; i < 256; ++i) {
+						num_zero_bits = 0;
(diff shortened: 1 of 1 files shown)


GideonZ:
Maybe it would be good to notify the maker of ChanFAT about this. This looks like excellent work!
I haven't tested it, though. How did you verify its behavior?

GideonZ:
By the way - who are you?

Issue #435, opened 2024-10-21
REST API's run_prg command


NusseTuffs64:
My current workflow with ALBERT involves using my Ultimate device's REST API's run_prg command with an attachment. I do this very frequently, to test artwork on a CRT screen. I observed that each time a run_prg command is run, a new /Temp/tempXXXX file is created on the Ultimate device. Does this raise any concerns, given the fact I use the run_prg so frequently? I am unlikely to get to /Temp/tempFFFF in a single sitting.

GideonZ:
I think that the first issues you will be encountering is that the ram disk will be full. I planned to make a solution that only the last 10 files or so, but it has not yet been implemented.

PR #451, opened 2025-01-04
Prevent /Temp filling up with temporary files

Gee-64:
Limits number of temporary files from the HTTP API as well as from the Assembly64 integration. House keeping retains at least 2 files, but may go up to 16 files or 1MB of total size, whichever is reached first. Older files are deleted before newer. The limits are counted separately for API files and Assemby64 files.

Temporary filenames were changed to have a slightly nicer (?) names and start numbering at 1 instead of 0 since this is (arguably) more intuitive to the user.

Also fixes a tiny typo in a comment.

This solves #435.

### DIFF
--- a/software/api/attachment_writer.h
+++ b/software/api/attachment_writer.h
@@ -5,18 +5,21 @@
 #include <stdlib.h>
 #include "indexed_list.h"
 #include "pattern.h"
+#include "filemanager.h"
 
 extern "C" {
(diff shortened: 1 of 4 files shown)


GideonZ:
@Gee-64 I checked the caches. Same hash. I don't understand! Grrrrr.... You could check my build.yml if you like.

PR #453, opened 2025-01-05
Make u64ii branch build for U2+L again

Gee-64:
NOTE: This PR is not toward "master" - it is toward "u64ii". I understand that branch is undergoing heavy development to get stuff ready for U64II, so please feel free to decline this PR if it is premature to do PRs there.

Either way, I managed to build and run the Ultimate app after applying the below fixes. I just did bare minimum and didn't add the event_pkt_connected struct which I noticed was added to the u64ctrl/main/rpc_calls.h file (I guess it is not needed for U2+ yet).

### DIFF
--- a/software/portable/riscv/riscv_main.c
+++ b/software/portable/riscv/riscv_main.c
@@ -18,6 +18,8 @@
 #include "profiler.h"
 #include "usb_nano.h"
 
+void print_tasks(void);
+
 typedef uint8_t (*irq_function_t)(void *context);
 
 typedef struct{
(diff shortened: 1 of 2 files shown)


GideonZ:
It's a bit premature to merge to this branch, but I'll merge it anyway. Thank you for fixing this.
It's temporary, because the WiFi code should become shared for the three different targets. Now they are copies of one another, which is not maintainable.

GideonZ:
The passwords are stored inside of the WiFi module indeed. It keeps a list of APs that were recently used and also marks the last one. This significantly speeds up the connection process, and avoids the issue in the firmware to have SSIDs and passwords shown in the config. In addition, it may allow the WiFi module to connect to the internet in the future, in case of a bricked device. (To be defined)

PR #459, opened 2025-01-12
Fix incorrect keyboard shortcut actions in popups

Gee-64:
**NOTE:** This PR covers the half of what still remains after #440 was closed.

When using keyboard shortcuts in popups an incorrect value is often returned from poll(), causing the wrong action to be taken.

The problem is that an "active button index" is returned rather than the expected "absolute button index".

Refactored code to return the correct value in all situations.

Bug reproduction:

1. Navigate to a random file you want to delete

2. Hit Enter

3. Select "Delete"

4. When you get the "Are you sure?" question popup, press the "Y" key on the keyboard - and you will see the file NOT being deleted

5. Try again, but press the "N" key, and the file WILL (incorrectly) be deleted

### DIFF
--- a/software/userinterface/ui_elements.cc
+++ b/software/userinterface/ui_elements.cc
@@ -88,21 +88,24 @@ int UIPopup :: poll(int dummy)
     if (c == -2) // error
     	return -1;
 
-    for(int i=0;i<btns_active;i++) {
+    int i;
+    int selected_button = -1;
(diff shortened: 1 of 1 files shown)


GideonZ:
@Gee-64 The caching doesn't seem to work, does it? Let's wait until it finishes, then I'll check what caches exist.

PR #460, opened 2025-01-12
Add startup state configuration (C64 / Ultimate Menu)

Gee-64:
Let users decide if they wish the machine to boot into the C64 or the Ultimate Menu. Convenient for users who always start their sessions by selecting disks or cartridges in the Ultimate Menu.

### DIFF
--- a/software/application/ultimate/ultimate.cc
+++ b/software/application/ultimate/ultimate.cc
@@ -120,7 +120,6 @@ extern "C" void ultimate_main(void *a)
 		c64 = C64 :: getMachine();
 		c64_subsys = new C64_Subsys(c64);
 		c64->init();
-		c64->start();
 	} else {
 		c64 = NULL;
 	}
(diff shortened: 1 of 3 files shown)


GideonZ:
Does this work correctly with the SID detection on the U64?

GideonZ:
> Quick question: Does the U64 FPGA return CAPAB_CARTRIDGE as well as CAPAB_ULTIMATE64?

Yes it does. When Cartridge would be false, it doesn't support reu, uci, sampler, cartridges, etc.

chrisgleissner:
Hi @GideonZ and @Gee-64 ,

this small PR has been open for 1.5 years. Is there still interest in merging it? 

Otherwise, I'd suggest closing it in order to clean up our PR list.

Thanks
Christian

Gee-64:
I didn't have a U64 at the time and was thus not able to test SID detection.

To keep things tidy I'll close this PR for now. I now have the required hardware and might pick this up a later point in time.

PR #466, opened 2025-02-11
Fix #464 printer configured device number ignored

### LINKED ISSUE #464: Virtual printer configured IEC device number is ignored
No mater if you select device ID 4 or 5 in virtual printer configuration, the virtual printer is always listening on IEC address 4

rgc2000:
This will call the configure method of IEC interface when printer device ID is changed

### DIFF
--- a/software/io/printer/iec_printer.cc
+++ b/software/io/printer/iec_printer.cc
@@ -139,7 +139,8 @@ void IecPrinter::effectuate_settings(void)
     uint8_t page_top, page_height;
 
     /* Printer ID on IEC from nvram */
-    int bus_id = cfg->get_value(CFG_PRINTER_ID);
+    printer_iec_addr = cfg->get_value(CFG_PRINTER_ID);
+    iec_if->configure();
 
     /* Load and apply virtual printer configuration from nvram */
     set_filename(cfg->get_string(CFG_PRINTER_FILENAME));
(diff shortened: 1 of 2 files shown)


radius75:
Hi,
I accidentally noticed that for some reason a very old commit is in this PR. Maybe it was overlooked a year ago and not added to Gideon/master branch then unfortunately. It may be needed for that old fix to work properly. Or it should be reverted from this PR.
Regards.
https://github.com/GideonZ/1541ultimate/pull/466/commits/241fb75d8f2954e25f31524fab078cabc815b842

radius75:
However, these changes are already in master. It must be some GitHub error that they are shown in this PR. Or there is a different character encoding and that is why the change
https://github.com/GideonZ/1541ultimate/blob/4b536242201e784485ce3f923928ad7894074590/software/io/printer/mps_png.cc#L33
https://github.com/GideonZ/1541ultimate/blob/4b536242201e784485ce3f923928ad7894074590/software/io/printer/mps_png.cc#L1872

rgc2000:
I have noticed this too. This is not the first time that GitHub shows me old commits that have already been committed in a pull request. I don't understand why and it may be a bug because my repository is also in GitHub as a fork of Gideon/master branch and they use the same commit IDs. Nevertheless, it looks harmless and those changes already committed are ignored.

rgc2000:
Will create a cleaner pull request with a rebased branch.

PR #471, opened 2025-02-20
Gather product related functions in one place

Gee-64:
The code to identify which hardware (i.e which product) is being used is duplicated in several places.

This merge request moves all of that half-messy code into a single "product"  unit, removing code duplication and keeping the calling code cleaner. A couple of future merge requests rely on this refactoring.

Note: This MR is toward the u64ii branch since the changes also cover the new U64E2 support.


GideonZ:
It should certainly also return true indeed. But I can fix this easily of course.

PR #474, opened 2025-02-24
Prevent /Temp filling up with temporary files

Gee-64:
**NOTE:** Replaces PR #451 which was accidentally closed.

Limits number of temporary files from the HTTP API as well as from the Assembly64 integration. House keeping retains at least 2 files, but may go up to 16 files or 1MB of total size, whichever is reached first. Older files are deleted before newer. The limits are counted separately for API files and Assemby64 files.

Temporary filenames were changed to have a slightly nicer (?) names and start numbering at 1 instead of 0 since this is (arguably) more intuitive to the user.

Also fixes a tiny typo in a comment.

This solves issue #435.

### DIFF
--- a/software/api/attachment_writer.h
+++ b/software/api/attachment_writer.h
@@ -5,18 +5,21 @@
 #include <stdlib.h>
 #include "indexed_list.h"
 #include "pattern.h"
+#include "filemanager.h"
 
 extern "C" {
     #include "multipart.h"
 }
 
+// The below will show up as: LOAD"/TEMP/WEB-UPLOAD",8,1
+// on the C64 screen. The truncation after 16 characters nicely
+// hides the slightly cryptic "-0001" part of the filename.
 #ifndef TEMP_FILE_PATH
 #define TEMP_FILE_PATH "/Temp/"
 #endif
+#define TEMP_FILE_PREFIX "web-upload-"
+#define TEMP_FILE_PATTERN "web-upload-????"
 
-#ifdef USE_FILEMANAGER
-#include "filemanager.h"
-#endif
 
 class TempfileWriter;
 
@@ -84,23 +87,24 @@ public:
         uint32_t dummy;
 
         HTTPHeaderField *f;
-
+        FileManager *fm = FileManager::getFileManager();
         switch(block->type) {
             case eStart:
                 fo = NULL;
                 break;
             case eDataStart:
-                sprintf(filename, TEMP_FILE_PATH "temp%04x", temp_count++);
+                sprintf(filename, TEMP_FILE_PATH TEMP_FILE_PREFIX "%04x", ++temp_count);
                 filenames.append(strdup(filename));
+                fm->house_keeping_delete(TEMP_FILE_PATH, TEMP_FILE_PATTERN);
 #ifdef USE_FILEMANAGER
-                FileManager::getFileManager()->fopen(filename, FA_WRITE | FA_CREATE_ALWAYS, &fo);
+                fm->fopen(filename, FA_WRITE | FA_CREATE_ALWAYS, &fo);
 #else
                 fo = fopen(filename, "wb");
 #endif
                 file_size = 0;
                 break;
             case eSubHeader:
-                sprintf(filename, TEMP_FILE_PATH "temp%04x", temp_count++);
+                sprintf(filename, TEMP_FILE_PATH TEMP_FILE_PREFIX "%04x", ++temp_count);
                 f = (HTTPHeaderField *)block->data;
                 for(int i=0; i < block->length; i++) {
                     if (strcasecmp(f[i].key, "Content-Disposition") == 0) {
@@ -123,8 +127,9 @@ public:
                     }
                 }
                 filenames.append(strdup(filename));
+                fm->house_keeping_delete(TEMP_FILE_PATH, TEMP_FILE_PATTERN);
 #ifdef USE_FILEMANAGER
-                FileManager::getFileManager()->fopen(filename, FA_WRITE | FA_CREATE_ALWAYS, &fo);
+                fm->fopen(filename, FA_WRITE | FA_CREATE_ALWAYS, &fo);
 #else
                 fo = fopen(filename, "wb");
 #endif
@@ -144,7 +149,7 @@ public:
                 if (fo) {
                     filesizes.append(file_size);
 #ifdef USE_FILEMANAGER
-                    FileManager::getFileManager()->fclose(fo);
+                    fm->fclose(fo);
 #else
                     fclose(fo);
 #endif
@@ -154,7 +159,7 @@ public:
             case eTerminate:
                 if (fo) {
 #ifdef USE_FILEMANAGER
-                    FileManager::getFileManager()->fclose(fo);
+                    fm->fclose(fo);
 #else
                     fclose(fo);
 #endif
--- a/software/filemanager/filemanager.cc
+++ b/software/filemanager/filemanager.cc
@@ -567,6 +567,33 @@ FRESULT FileManager::delete_recursive(Path *path, const char *name)
     return ret;
 }
 
+FRESULT FileManager::house_keeping_delete(const char *dirpath, const char*matchPattern, int min_files, int max_files, uint32_t max_size)
+{
+    Path dir;
+    dir.cd(dirpath);
+    IndexedList<FileInfo *> files(32, NULL);
+    FRESULT res = get_directory(&dir, files, matchPattern);
+    if (res == FR_OK) {
+        files.sort(compare_timestamp);
+        uint32_t total_size = 0;
+        for (int i = 0; i < files.get_elements(); ++i) {
+            total_size += files[i]->size;
+            if (i >= min_files && (total_size >= max_size || i >= max_files)) {
+                printf("House keeping: Deleting %s from %s\n", files[i]->lfname, dirpath);
+                delete_file(&dir, files[i]->lfname);
+            }
+        }
+    }
+    return FR_OK;
+}
+
+int FileManager::compare_timestamp(IndexedList<FileInfo *> *list, int a, int b)
+{
+    uint32_t a_timestamp = ((*list)[a]->date << 16) | (*list)[a]->time;
+    uint32_t b_timestamp = ((*list)[b]->date << 16) | (*list)[b]->time;
+    return (int)(b_timestamp - a_timestamp);  // Newest first (descending)
+}
+
 
 FRESULT FileManager::rename(Path *path, const char *old_name, const char *new_name)
 {
--- a/software/filemanager/filemanager.cc
+++ b/software/filemanager/filemanager.cc
@@ -569,6 +569,11 @@ FRESULT FileManager::delete_recursive(Path *path, const char *name)
 
 FRESULT FileManager::house_keeping_delete(const char *dirpath, const char*matchPattern, int min_files, int max_files, uint32_t max_size)
 {
+#ifdef U2
+    if (max_size > 480*1024) {
+        max_size = 480*1024;  // Only 1MB ramdisk for U2 (software/filesystem/ramdisk.cc), limit max_size to slightly below half
+    }
+#endif
     Path dir;
     dir.cd(dirpath);
     IndexedList<FileInfo *> files(32, NULL);
--- a/software/filemanager/filemanager.h
+++ b/software/filemanager/filemanager.h
@@ -119,6 +119,8 @@ class FileManager
 	FRESULT rename_impl(PathInfo &from, PathInfo &to);
 	FRESULT delete_file_impl(PathInfo &pathInfo);
 
+    static int compare_timestamp(IndexedList<FileInfo *> *list, int a, int b);
+
 //	friend class FileDirEntry;
 
     void lock() {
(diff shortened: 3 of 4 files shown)


GideonZ:
Yes. But since I am studying for an exam right now, I don't have much time to dive into it. I think it is great that you have implemented this! Maybe I would have gone for a slightly simpler approach, e.g. keep a list of 'N' downloaded files and simply delete one by one, once they fall out of the delay line / queue / fifo. When the size of the uploads is limited to 'x', then the maximum allocation on /temp also has an upper bound. However, your implementation seems to be less implicit, which is good. I haven't tested it, though.

GideonZ:
Thanks!

No need to remove the size limitation. Maybe the limitation is a bit low. The /temp disk is 16MB on most platforms. So when the api and assembly use 8 MB together, it is fine. No need to change anything really, if it works, it works.

GideonZ:
Ok, I did a bit deeper review, and I have some questions regarding the behavior, or let's say concerns.

Currently, the files are sorted based on their time stamp. This is only necessary since you are reading a directory. This would not have been necessary if the filenames were stored in an indexed list. Reading the directory every time a temporary file is created is also an expensive operation, which feels redundant to me. (Prove me wrong!) In addition, if the connection to a DNS fails, the U64E-II does not know the time, since it does not have a RTC.

Another concern that I have with this code is that some files might be open. I am not sure if the delete_file function simply returns an error (and if so, it is not checked!). So it needs to be verified whether it would be possible to pull the rug from under a mounted disk for example. In case of a .d81 for instance, all operations are done on the file, not on a copy thereof. So deleting a file that is still mounted should at least result in predictable behavior; either a disk eject, or an error while deleting, in which case the file remains open and the used space is not subtracted from the total in use.

chrisgleissner:
Hi @GideonZ and @Gee-64 , 

I think this would be a hugely valuable addition to the firmware. I know various other people ran into this - I recall someone tried to live-stream Petscii-converted video streams via REST and quickly ran out of space.

I noticed that there has not been progress on this in a year. Are you still working on it or do you mind if I take on this PR to complete it? 

My suggestion would be to delete all uploaded files automatically in the `TempFileWriter` destructor unless it is a mounted file, in which case it gets deleted on unmount.

That way, no housekeeping job is needed.

Thanks
Christian

Gee-64:
You've been on an absolute tear lately @chrisgleissner and you are more than welcome to pick this one up! You're doing awesome work!

This housekeeping PR was a quick non-ideal hack for an immediate problem. Doing it properly would mean tracking when the temporary files are no longer needed and delete them at that point - which is exactly what you are suggesting. I think a separate PR with your approach is cleaner (if you need anything from mine, just rip it). When your PR is merged I'll happily close this one as superseded.

GideonZ:
@Gee-64 @chrisgleissner Hi guys! I am sorry about leaving this issue untouched for quite some time. The PR from Gee-64 didn't make it because it was more complex than necessary, but in all fairness, I didn't come up with a cleaner solution. So in short: I have been a bit of an a$$.

I'd suggest to use a container component and push filenames into it and add the size of the files when new temporary files are created. Then start popping entries from it once the maximum number of files is reached, or when the totally allocated size exceeds the programmed limit. This is easier than loading the directory, sort them on time, etc.

Leaving a few files in temp is handy and needed for some assembly 64 stuff, titles that consist of more than one file. I expect newer titles to come with .cfg files for example and they will both be written before they are passed to the handler.

The most difficult issue is that some files can stay open for a long time, e.g. when they are mounted. They can only be deleted after they have been closed.

_Ideally, the file manager would have a concept of temporary files. In that way, the problem is solved globally and not for the ftp client, the assembly 64 client and the api server separately. In addition, the Filemanager could use special File* handles that wrap the actual file handle to implement a "delete on close if space exceeded" function._

Gee-64:
> but in all fairness, I didn't come up with a cleaner solution. So in short: I have been a bit of an a$$.

Nah, the BDFL gets to decide the bar for what goes in and what does not. Just because you at the time didn't have a better idea does not mean you have to merge something that is "below the bar". Keeps the projects coherent. This PR was thrown together in an evening or two, and although it is functional and works fairly well, it is just a workaround for the real problem of handling temporary files properly.

chrisgleissner:
Thanks for the praise @Gee-64 (btw, cool handle!) and for granting me permission to follow in your foot steps...

Two questions @GideonZ :

1. Are there plans to support Assembly64 / REST call interactions with `/Temp` that span multiple files? You mentioned Assembly64, but is something similar even in the works for REST, i.e. you can upload multiple disk files, or a program and its .cfg in one go?
2. Are we looking for a solution that smartly cleans `/Temp` from no longer needed / outdated files regardless of which part of the code created them? Because I was initially just looking at the REST POST endpoints - mostly because I care about these a lot due to another project :-) But very glad to expand scope and address this issue once and for all.

If the answer to both questions is "Yes", then most likely we need some logic that performs mount-aware deletion of oldest/largest files, i.e. some kind of housekeeping job for `/Temp` (as a gatekeeper to ensure `/Temp` never runs out of space) as was initially planned and implemented by @Gee-64 . I.e. keep `/Temp` size below configurable watermark level with sensible default (should we expose this in the UX?), prioritize oldest / larges files for removal, and ensure mounted images always remain.

That way the clean-up would be centrally located instead of being spread all over, and it would allow the use case you mentioned of having access to past files instead of them disappearing as soon as their job is done.

Thanks
Christian

chrisgleissner:
Hi @GideonZ and @Gee-64 ,

I started this work with the following assumptions:
- Answers to my questions in the previous posting are 'Yes'
- No user control of maximum `/Temp` fill level (hard-coded to 70%). If you want me to expose it (e.g. offer options 10% to 90% fill level in a drop-down somewhere, 10% steps), I can add that, but it adds test complexity and I'd rather keep this focused on the essentials initially.

Thanks
Christian

GideonZ:
> 1. Are there plans to support Assembly64 / REST call interactions with `/Temp` that span multiple files? You mentioned Assembly64, but is something similar even in the works for REST, i.e. you can upload multiple disk files, or a program and its .cfg in one go?

Yes, that is certainly in the making; and the attachment writer in the ultimate already supports this. I think for some SIDs Assembly 64 already returns the song length file together with the SID file.

> 2. Are we looking for a solution that smartly cleans `/Temp` from no longer needed / outdated files regardless of which part of the code created them? Because I was initially just looking at the REST POST endpoints - mostly because I care about these a lot due to another project :-) But very glad to expand scope and address this issue once and for all.

Yes. But, the parts that created the file do this with the intent for the file to be temporary. So, if a user saves a file on /Temp, it should *not* be deleted automatically by some sweeper function. This is the beauty of making the FileManager (VFS) aware of what a temporary file is. Then it doesn't matter anymore where it is located, it will just get deleted after some time. 

I am aware of the difficulty where the file system and the FileManager get out of sync. For instance, when the temporary directory is placed on a USB stick for instance; power cycling makes the FileManager reset its state, while the USB stick directory remains filled. However, this is currently not how it works; temporary files always go to the RAM disk.

Obviously, the temporary files from the API, FTP client and Assembly 64 could go into a deeper directory like /Temp/cache, while user files remain on /Temp. Similarly, if we ever decide to move the temp folder to USB (e.g. for U2, which doesn't have enough RAM), we could do a delete "<temp>/cache/*" at boot.

So in all fairness, I am not completely settled on the final idea. I like the idea of making the FileManager aware of what a temp file really is; so that it could also do the cleaning when the file gets closed a second time (The first time is the creation), or it gets deleted when it is closed, marked as temporary and the flags have FA_CREATE_ANY = 0

What I like about the fact that the user can see the cached files, is that a user can decide to copy a cached file elsewhere before it disappears. 

> If the answer to both questions is "Yes", then most likely we need some logic that performs mount-aware deletion of oldest/largest files, i.e. some kind of housekeeping job for `/Temp` (as a gatekeeper to ensure `/Temp` never runs out of space) as was initially planned and implemented by @Gee-64 . I.e. keep `/Temp` size below configurable watermark level with sensible default (should we expose this in the UX?), prioritize oldest / larges files for removal, and ensure mounted images always remain.

Yes, this is exactly what @Gee-64 implemented; although it does not make a difference between cached files and user files, and it needs to read the directory every time, sort by date/time etc. That's not wrong per se, but it is a hassle. If we just add a function to the FileManager "create_temp_file", it can keep the filename in an internal list; or two lists if you wish: an 'active' list and a 'ready for cleanup' list. When the file gets closed and the create flag is not set, it moves from 'active' to 'ready for cleanup'. Once the ready for cleanup list exceeds size 'x', the delete is called. Alternatively, instead of two lists, we could decide on one list and each entry has a close count, or other flag that gets updated when the files gets closed and the create flag is not set.

As I said; I am not fully settled on the idea, because it also feels like complexity that may not be strictly needed; functionality that may not be needed at all, if only .....  (and I don't know yet what that if only is...)

chrisgleissner:
Hi @GideonZ ,

thanks for your feedback and for the cache idea.

Based on our earlier discussion and some exhaustive code search looking into stakeholders that interact with `/Temp`, I suggest the following approach:

- Keep ordinary user-visible `/Temp` files as they are.
- Use `/Temp/cache` only for managed, reclaimable temp files such as REST upload scratch files, Assembly64 scratch/cache files, and socket DMA import files.
- Track those files explicitly in `FileManager`, so cleanup is based on ownership and lifecycle, not just on the pathname.
- Trigger cleanup when the managed files in `/Temp/cache` together reach 70% of the total `Temp` volume, and then remove the oldest reclaimable cache files until the managed-cache usage drops toward 50%.
- Remove the oldest reclaimable managed temp files first, i.e. the least recently created cache entries that are no longer open or otherwise in active use.
- Keep the newest 10 managed temp files protected as long as there are older reclaimable ones available for removal first. (Caters for multi disk scenario.)
- Never delete files that are still open, mounted, or otherwise in active use. (Caters for mount scenario.)

I think your cache idea is the right split. It keeps the user in control of normal `/Temp` contents, while still keeping the device operational for REST-driven workflows where uploaded files often get left behind. It also makes those REST operations more self-contained, instead of requiring a later FTP cleanup step.

If this matches your ideas, is this fine to start implementation with?

Thanks  
Christian

GideonZ:
@chrisgleissner Yes, that's pretty much it. Very well summarized. I would trigger the cleanup already at 10 files, not only when the 70% watermark is reached. This avoids that the list of managed files gets large with many small files.

Note that the /cache directory is strictly not even necessary if the FileManager keeps track of the temporary files that are created in the network services. (@chrisgleissner)

chrisgleissner:
Thanks @GideonZ , much appreciated.

Just to ensure that the implementation can pass the PR review without major glitches ;-) , please tell me if you want me to remove the dedicated `cache` subfolder.

TBH, I quite like it from a user perspective. One concern I had before you posited the `cache` subfolder idea was the fact that a user has very little insight into the decisions of the new house keeper. And just like a house keeper in your real house, you probably want to tell her/him: "Please only clean this room. Do not touch the other rooms, they are off limits even if they contain trash." 

The `cache` concept makes this very explicit and allows the user to move valuable items to a house-keeper safe area even when as per normal instructions, he/she is supposed to clean it. 

The only "con" argument I can think of is that this may break user expectations who have come to rely on REST-uploaded files to exist under `/Temp`, not under `/Temp/cache`. 

So if you agree, we could start with `/Temp/cache` and see how it goes. It will be configurable anyway, so if we find it makes no sense, we can always this constant to `/Temp`. We could even expose this to the user for backwards compatibility.

Please let me know if that sounds reasonable.

GideonZ:
Sounds very reasonable. We will keep the /Temp/cache as a root for now. Let's not forget to have one call somewhere that creates that dir. But I am sure you already have it.

chrisgleissner:
Hi @Gee-64 and @GideonZ ,

I hope you are both well.

I now raised https://github.com/GideonZ/1541ultimate/pull/686 for the change as discussed.

Best wishes
Christian

chrisgleissner:
Hi @Gee-64,

Good news - @GideonZ has merged https://github.com/GideonZ/1541ultimate/pull/686 into `master`.

Thanks again for raising the issue in the first place and putting together the initial implementation. That was really helpful in getting this across the line.

Would you be happy to close this PR so we can keep the list of open PRs up to date?

All the best,  
Christian

Gee-64:
Closing this PR as superseded by #686 . Nice work  @chrisgleissner !

Bit off-topic, but it seems like you’ve really adopted AI tools in your dev workflow. Curious how you’re using them and if they’ve made a difference in productivity or code quality.

chrisgleissner:
Thanks @Gee-64 . Yes, I can highly recommend GPT 5.5 as well as Claude Opus 4.7. They have most definitely improved my productivity. The rise of AI tools is the largest transformation I've seen in the field of software development in the past 4 decades I've been writing code.

I use these tools mostly via VS Code as well as via the Copilot plugin, but I also recommend the Codex and Claude Code plugins. These plugins are all quite similar. The right prompt, however, makes all the difference, as well as very careful review of the resulting code. The old garbage-in, garbage-out rule still holds. Also, test-driven as well as spec-driven development has become once more important in the age of AI. It is best to start with a spec that is clear and precise, and then ensure good test coverage.

I hope this helps. Thanks again for the idea you had with the Temp cleaning. In the age of AI, ideas are as important as always, if not more. And tenacity to see something through to the end is always still the most important quality of a software developer IMHO.

Gee-64:
Sorry for a fairly late answer @chrisgleissner, been a super busy week here. Thank you for sharing these nuggets of wisdom! I'm late to the AI party, but will definitely ramp up incorporating AI tooling after seeing what you are producing in this as well as other projects.

Cheers!

PR #476, opened 2025-02-25
Add unique_id to UDP discovery packet and /v1/info API endpoint

Gee-64:
This is an ID which uniquely identifies the current Ultimate device. This allows Assembly64 (and other software) to recognize individual Ultimate units over time even if they change firmware version, hostname, ip or switch from LAN to WiFi (MAC address changes).

The latest beta of Assembly64 makes good use of this id to manage it's list of devices.

### DIFF
--- a/software/api/routes.cc
+++ b/software/api/routes.cc
@@ -131,7 +131,8 @@ API_CALL(GET, info, none, NULL, ARRAY( { }))
 #ifdef U64
         ->add("core_version", core_version)
 #endif
-        ->add("hostname", hostname);
+        ->add("hostname", hostname)
+        ->add("unique_id", getProductUniqueId());
 
     resp->json_response(HTTP_OK);
 }
--- a/software/network/socket_dma.cc
+++ b/software/network/socket_dma.cc
@@ -557,6 +557,7 @@ void SocketDMA::identThread(void *_a)
 #endif
                     ->add("hostname", hostname)
                     ->add("menu_header", menu_header)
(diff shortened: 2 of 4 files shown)


FeralChild64:
This is really worrying. A super-cookie, impossible to delete :/

FeralChild64:
That’s indeed less worrying - I was afraid the software running on the device could get it’s serial number.

FeralChild64:
If the software can't do this - than indeed there is no problem.

Grrrolf:
@Gee-64, 

The potential for third parties to link the ID to an individual triggers GDPR obligations for Gideon.
A device with a unique ID accessible via the internet falls under the "connected product" definition of the EU Data Act.
It would be good to investigate the consequences first before implementing it in such a way that that prevents the end user to have control over sharing data. 

Please seek legal advice so that you know what the legal consequences are for Gideon. Then on how implement such a feature and update the privacy policy. Ideally you want the device owner to be in control of the data that is being shared by the device.

GideonZ:
@Grrrolf An easy change to this ticket would be, to let the user change the ident string, which could be preloaded with the value it gets now from the unique identifier.

GideonZ:
@Grrrolf In addition: _accessible via the internet_: Is the ultimate ever really accessible through the internet? It may be accessible through your LAN, but that's not the same thing. Once you decide to open your router towards the device, you also make the decision to expose some information. Accessing the internet is not the same as accessible via the internet.

PR #479, opened 2025-02-28
Make the U2+L FPGA core boot again

Gee-64:
Thank you to Gideon for showing what the problem was and how to fix it! Adding documentation below in case someone else needs it in the future.

The FPGA will not boot if the Lattice tools assigns for example riscv/* signals to any of the global primary nets. Unfortunately the tools keep changing net names and this can at any time suddenly get inappropriate signals assigned to the global nets causing an unbootable FPGA core.

Procedure that can help get a working FPGA again:

Search for a line like this in build.log to see which signals were assigned to global nets:

>   WARNING - par: The following 9 data signals have to use primary clock resources:

Also, since net names change, also look for obsolete PROHIBIT rules by searching for

>  Semantic error in "PROHIBIT PRIMARY NET

Finally update the rules in "u2p_ecp5.lpf" accordingly.


GideonZ:
We should verify whether the binaries of our github cache are the same. If they are not, but all the input files (and thereby its hash) are actually the same, we may have a problem.

PR #481, opened 2025-03-01
Fix memory leak in custom popup

Gee-64:


GideonZ:
I am amazed that you found this. Are you using klocwork or some other static analysis tool? :-)

PR #484, opened 2025-03-02
Add menu button API and web page button

Gee-64:
Simple way to "pause/unpause" the C64 by having it enter/leave the Ultimate menu.


GideonZ:
This new API call shall also be covered in the documentation. I should make a ticket for the documentation repo.

PR #488, opened 2025-03-08
Make Unique ID configurable

Gee-64:
Makes it possible to disable Unique ID or change the ID to a custom value.


GideonZ:
@Gee-64 As I mentioned in the other PR, I wonder if it is a good idea to have configuration strings that start with an escape value. Does this work well in the config browser, or does it cause alignment issues?

PR #489, opened 2025-03-09
Improve API for CFG_TYPE_STRFUNC config settings

Gee-64:
String settings with dynamically created lists of possible presets (CFG_TYPE_STRFUNC) were exposed in the API just as normal strings without the presets. This change adds a field "presets" containing the same values shown in the Ultimate Menu. This is similar to the "values" field for CFG_TYPE_VALUE.

The reason to use "presets" and not "values" is to make clear to the caller that an arbitrary string is also allowed (which it isn't for CFG_TYPE_VALUE with a "values" list).

### DIFF
--- a/software/api/route_configs.cc
+++ b/software/api/route_configs.cc
@@ -13,7 +13,7 @@ void emit_store(ConfigStore *st, JSON_Object *pobj, ArgsURI &args)
     for(int n = 0; n < st->items.get_elements(); n++) {
         i = st->items[n];
         if ((args.get_path_depth() < 2) || (pattern_match(args.get_path(1), i->definition->item_text))) {
-            JSON *obj;            
+            JSON *obj;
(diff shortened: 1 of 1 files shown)


GideonZ:
This is a nice addition indeed. For me, it wouldn't be necessary to check for strings that start with an escape. I think configuration values simply should never start with an escape. When was this introduced? :-)

GideonZ:
> I sort of agree, I just noticed it was introduced by one certain Gideon in 2021 to allow empty strings [..]

I don't even know that guy. Maybe I should better review this shitty code before I allow these low quality additions from Gideon.

> Just supply an empty string in the lists, and then the UI could render that particular case as "\er- None -". 

This is the way. An empty string is an empty string, and if you want to display that as '-None-' then it should be part of the display function, not stored like that. Yes, it should be cleaned up at some point. However, it is also convoluted to do this only for CFG_TYPE_STRFUNC and not for other strings. But then the display of many other optional string fields will start showing -None-, too.  On the other hand, if it is something specific of the drop down and to be shown only in the drop down menu, then it's logical to solve it there. Not sure what is the most user-friendly / intuitive solution here.

PR #494, opened 2025-04-10
issue #493 fix for 128D and 128DCR

### LINKED ISSUE #493: C128D issue with RETURN key in menus
## Issue Description
When navigating in menus, entering a menu using the RETURN key does not work as expected. Pressing RETURN key once does not do anything. Keep holding RETURN key for 5-20 seconds will eventually enter the menu.

Navigating using right arrow key works as expected. Unfortunately to enter popup menu is only possible by pressing the RETURN key.

In BASIC, all keys work as expected. CIA chips tested OK.

Only C64 mode has been used during testing.

The settings is phi1=200ns, phi2=200ns. All other settings are disabled (REU, kernal shadow RAM,...). Tested various combinations already.

## Hardware & Firmware Tested

**C128D** (plastic version) is impacted by the behavior described above.

**C128DCR** (metal version) is impacted also, but holding enter for up to 3 seconds usually will enter the menu.

Two cartrdigges were tested: Ultimate-II+L and Ultimate-II+L with wifi. Both having latest 3.11a firmware. Both bahaving exactly the same.

slesinger:
## Problem
Users of the 128D and 128DCR models have reported that the RETURN key feels unresponsive compared to other keys, causing typing frustration and affecting user experience when entering commands.

## Solution
This PR implements special handling for the RETURN key on C128D and 128DCR models by:

- Adding dedicated logic to detect and process RETURN key presses differently
- Reducing the debounce delay specifically for the RETURN key
- Cutting the initial key delay in half for RETURN key presses (using `first_delay / 2`)
- Accelerating the delay counter decrement to make repeat presses more responsive
- Using half the standard repeat delay (using `repeat_speed / 2`) for subsequent RETURN key presses

## Testing
The implementation has been tested on:
- C128D hardware
- C128DCR hardware
- Standard C64 to ensure no regressions

## Related Issues
Fixes #493 - "Slow RETURN key response on 128D and 128DCR models"

### DIFF
--- a/software/io/c64/keyboard_c64.cc
+++ b/software/io/c64/keyboard_c64.cc
@@ -178,23 +178,48 @@ void Keyboard_C64 :: scan(void)
         shift_prev = 0xFF;
         return;
     }
-            
-    if((shift_flag == shift_prev) && (mtrx_prev == mtrx)) { // this key was pressed before
-        if (delay_count == 0) {
-            delay_count = repeat_speed;
-            // do not return, store char
+    
+    // Special handling for the RETURN key - make it more responsive on C128D machines
+    // RETURN key is in keymap_normal[1] or on specific matrix coordinates
+    if(key == KEY_RETURN) {
+        // For RETURN key, reduce the debounce delay to make it more responsive
+        if((shift_flag == shift_prev) && (mtrx_prev == mtrx)) {
+            // RETURN key was pressed before, don't wait as long for repeat
+            if (delay_count <= 1) {  // Use a shorter threshold for RETURN key
+                delay_count = repeat_speed / 2;  // Use half the repeat delay for RETURN
+                // Continue to store the key below
+            } else {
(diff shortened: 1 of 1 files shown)


GideonZ:
Why does this help?

chrisgleissner:
Hi @slesinger ,

I see that @GideonZ asked a question about this PR over a year ago but there has been no reply.

Are you still interested in getting this merged?

Thanks
Christian

chrisgleissner:
Hi @slesinger,

Thank you for looking into this. 

The reported C128D/C128DCR behaviour still warrants investigation, but this change affects RETURN handling globally and appears to alter its repeat behaviour rather than address why the initial key press is not reliably detected.

As we have not received further details or a response to the earlier question, I’ll close this PR for now and leave the issue open. Please feel free to reopen with an updated, targeted fix if you would like to continue investigating it.

Thanks again for the contribution.

PR #505, opened 2025-05-16
Just for testing My current changes for 3.12

markusC64:
This merge request contains my current changes for 3.12. So that we have them together for testing.

You do not need to merge them if you prefer single merge requests.

### DIFF
--- a/software/drive/c1541.cc
+++ b/software/drive/c1541.cc
@@ -498,7 +498,7 @@ void C1541 :: insert_disk(bool protect, GcrImage *image)
 void C1541 :: map_gcr_image_to_mfm(void)
 {
     MfmDisk *mfm = mfm_controller->get_disk();
-    mfm->init(fmt_Clear);
+    mfm->init(fmt_Clear, 80);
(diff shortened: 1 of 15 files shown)


GideonZ:
Hi Markus! I pulled most of your changes. One thing that needs to be fixed is the rtc_i2c_sntp. This file shouldn't exist, or when it exists it should not address the I2C directly, because: code duplication. Instead, there should be a wrapper that uses either the I2C time OR the SNTP time, depending on the setting. For the U64E2 there is no RTC chip, so it can only use SNTP. By including the correct module in the build, this should automatically work.

markusC64:
Yeah, people are asking for RTC in U64 II. Not for the moment, but for a later time, you might support an USB RTC. Such like https://www.tindie.com/products/sbc/usb-rtc-type-c-breakout-ds3231-usb-rtc-breakout/

"rtc_i2c_sntp.cc" - right. But before deciding what is best for that, the following question should be answered: Is SNTP possible for the U2+(L)?
From that answer it depends what s best.,

PR #538, opened 2025-07-19
Updated Makefile and README.txt for devices with esp32

bvl1999:
Updated toplevel Makefile to add esp32 dependencies to u2pl, u64 and u64ii targets. Ensure make clean also cleans esp32 builds.

Also added Espressif to toolchains for devices with esp32 in README.txt


bvl1999:
> Did you forget to add README.txt changes to git?

Yes, added it now, thanks!

PR #584, opened 2025-12-29
Do not ask for authentication mode for empty passwords/public APs

TangoBravo64:
See GideonZ#581.


TangoBravo64:
Wait!

The dialog issue is resolved. But manual entry of an SSID still doesn't result in a successful connection to a public AP. Only "Select AP from list" works. Something else is up...

GideonZ:
Sorry, turned out that I had not saved all of my edited files... *shame*. I am denying your change for now. I think it was good, though. I swapped the questions: first the authentication mode, then the password if it is not "open".

TangoBravo64:
Ok, I see your changes. Looks like a better solution. Will test that, too.

PR #595, opened 2026-01-02
Update Live Monitor and BASIC Editor Pages

danielwcarey:
On Live Monitor page, updated the page to include the "help" output beneath the monitor. 
On the BASIC Editor page, added the buttons "Upload and Run" and "Run".
Added spacing to the buttons using top-right 5px spacing. 
Fixed html warnings for <a/> tag needing "rel='noopener'". 
Fixed html lang warning by adding lang="en".
Fixed head tag warnings adding title and meta charset.
Live Monitor Page
BASIC Editor Page

### DIFF
--- a/html/index.html
+++ b/html/index.html
@@ -1,6 +1,8 @@
-<html>
+<html lang="en">
 <head>
-<meta http-equiv="content-type" content="text/html; charset=UTF-8">
+<meta charset="utf-8">
+<title>Ultimate II/64</title>
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
 <script src="https://cdn.jsdelivr.net/npm/jquery"></script>
 <script src="https://cdn.jsdelivr.net/npm/jquery.terminal/js/jquery.terminal.min.js"></script>
 <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/jquery.terminal/css/jquery.terminal.min.css"/>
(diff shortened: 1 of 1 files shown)


html/index.html
```diff
@@ -181,6 +183,22 @@
         parseBASIC();
     });
 
+    $('#submitBASICAndRun').click(function (event) {
+        event.preventDefault(); // Prevents the default action of the anchor tag
+        parseBASIC();
+        runBASIC();
+    });
```
radius75: It seems to me that the RUN command is sent without confirmation whether the larger program has already been fully transmitted asynchronously, which causes errors.

radius75:
AI suggests, for example, such a solution.

```
// Helper function to create a delay
const delay = ms => new Promise(res => setTimeout(res, ms));

$('#submitBASICAndRun').click(async function (event) {
    event.preventDefault();

    try {
        // 1. Send the BASIC code to C64 memory via API
        // We use 'await' to ensure the transfer starts correctly
        await parseBASIC();

        // 2. Wait for a safety margin (500ms) 
        // This ensures the hardware has finished DMA memory writing
        await delay(500);

        // 3. Send the 'RUN' command to the keyboard buffer
        await runBASIC();
        
        console.log("Program transmitted and executed successfully.");
    } catch (error) {
        console.error("Error during transmission or execution:", error);
    }
});
```

danielwcarey:
I'll see if i can reproduce the run occurring before the transfer. async/await will be better than a delay:
$('#submitBASIC').click(async function(event){
    event.preventDefault(); // Prevents the default action of the anchor tag
    await parseBASIC();
});

$('#submitBASICAndRun').click(async function (event) {
    event.preventDefault(); // Prevents the default action of the anchor tag
    await parseBASIC();
    await runBASIC();
});

$('#runBASIC').click(async function (event) {
    event.preventDefault(); // Prevents the default action of the anchor tag
    await runBASIC();
});

PR #600, opened 2026-01-05
Use tokenized length and fix line number parsing

radius75:
Description of changes:
- Removed UI restriction: Deleted the hardcoded 160-character limit in the keypress event. This allows users to type long descriptive tokens (like {3 spaces} or {clear}) without the editor blocking input prematurely.
- Added Binary Size Validation: Implemented a real memory-limit check in parseBASIC(). Now, each line is validated after tokenization to ensure the total binary size (including address and line number headers) does not exceed the C64 limit of 256 bytes.
- Improved Line Number Parsing: Switched from split(' ') to a regex-based approach (match(/^\d+/)) to extract line numbers. This correctly handles BASIC shorthand where no space is used between the line number and the command (e.g., 10print).
- Code Cleanup: Optimized memory address calculation by caching the tokenized line length and updated comments for better clarity.
obraz

### DIFF
--- a/html/index.html
+++ b/html/index.html
@@ -1479,7 +1479,7 @@ function disassembler(startInt, byteArray) {
             <p><b>Special characters:</b></p>
             <div id="tableContainer"></div>
 
-            <textarea id="basiceditor" class="c64-textarea" cols="80">
+            <textarea id="basiceditor" wrap="off" class="c64-textarea" cols="80">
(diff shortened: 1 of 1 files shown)


radius75:
no line wrapping in the tokenizer window, which makes it easier to analyze long lines of code composed of many SPECIALs
obraz

Issue #613, opened 2026-01-13
U64 Elite 2 firmware 3.14a Remote Streams starting


grue74:
I noticed that U64 Streamer app cannot start remotely vic / audio streams anymore. 

Is this a bug or deliberate change in firmware? I can start the streams manually from the u64 menu and video/audio streams are working normally after that.

xahmol:
Problem is only that the new API method is extremely unreliable for streams.
With Assembly64, streaming only works via the 'traditional method', via API it just does not work. Starting the stream automatically does sometimes work, but not always.

So I indeed resort to manually starting the stream on the U64 side now.
And actually use the OBS C64 streamer plugin now that works quite well. But can not automatically start the stream remotely.

https://obsproject.com/forum/resources/c64-stream.2229/

Honestly, for stability of remote control, 3.12a is a step back from
3.11f. Did not try 3.14 yet.

xahmol:
Have to update my previous post: just tried 3.14a and for now, uStream in Assembly actually now works with auto starting and streaming over the REST interface (instead of legacy).
So that is good news. Apparently it does work, that is, if you disable WiFi. Know that streaming over WiFi does and can not work, but streaming over Ethernet also does not work with WiFi enabled as well.

Gee-64:
> Gideon mentioned there were only cosmetic changes. However...
> 
> In the previous firmware:
> 
> > * February 2025: Added optional password protection to all network services (FTP, Telnet, Web API, Socket/DMA). If a password is set, clients must include it.
> 
> This was already a breaking change, but due a bug it kept working.

It was not meant to be a breaking change, it was supposed to be fully backwards compatible for all protocols (unless a password is set of course). If something isn't working as before (without a password set), then there is a bug somewhere. @Grrrolf If know of any other situation where things broke when password support was added please let me know.

I have never used the streaming stuff (when I implemented the password support I only had a II+ which AFAIK does not support streaming).

I'll carve out some time to play with the streaming on my U64E2 and see if I can replicate the issue.

grue74:
I checked what u64 outputs to my syslog service when trying to start stream from U64 Streamer:

Jan 13 12:41:25	192.168.0.64	@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@dmaThread			newsockfd = 00000007. Client = 192.168.0.110 	47169
[...]

PR #614, opened 2026-01-13
Uninitialized string buffer + missing NUL terminator + invalid data streamer pointer fixes.

TangoBravo64:
Stack variables are not automatically initialized. There was a code path were "dest_host" was never written, but read (strlen/printf/...), resulting in undefined behavior (best case, just showed garbled text). This would likely fix the "Finnish host name" issue in #613.

Also, strncpy(...) was used, without guaranteeing string termination. If the copied string matches or exceeds the given maximum length, no NUL terminator is written by strncpy. The code seemed to avoid this by using a reduced length (36 instead of the buffer's actual 40 byte size). But this does not help, unless the buffer was zero initialized.

EDIT: Now also contains a second commit - for a more severe pointer issue with the dataStreamer object.

### DIFF
--- a/software/api/route_streams.cc
+++ b/software/api/route_streams.cc
@@ -32,12 +32,12 @@ API_CALL(PUT, streams, start, NULL, ARRAY ( { { "ip", P_REQUIRED } }))
     }
 
     if (streamIndex == 0) { // video streams require debug to be off
-        sys_command = new SubsysCommand(NULL, -1, (int)&dataStreamer, 2, "", "");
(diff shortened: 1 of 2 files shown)


TangoBravo64:
Now contains a second commit - for a more severe pointer issue with the "dataStreamer" object.

GideonZ:
These are VERY GOOD FINDS!!  I had indeed changed the dataStreamer from a global initializer to a pointer, and apparently didn't fix the references elsewhere. THANK YOU.

PR #616, opened 2026-01-14
Ultimate64/II BASIC tokenizer

### LINKED ISSUE #450: [Feature Req.] Web API. Basic Tokenizer/Editor - abbreviation
A proposal to allow entering code with command abbreviation.

obraz

Here is an example (index.html) that works for me. The person who deals with this part of the API could take a look at it.

I have also partially entered some tokens from CBMPrgStudio , but adapting this Tokenizer to fully cooperate with source codes from CBMPrgStudio and other Basic compilers would require a lot of work. And I am not sure if this is the assumption here.

```
                               //.map(line => line.trim().toLowerCase())  //Shifted is neded!
```

```
    const TOKENS = [
        ['restore', 140],	['reS', 140],		['resT', 140],		['restO', 140],		['restoR', 140],
        ['input#', 132],	['iN', 132],		['inP', 132],		['inpU', 132],		['inpuT', 132],
        ['return', 142],	['reT', 142],		['retU', 142],		['retuR', 142],
        ['verify', 149],	['vE', 149],		['veR', 149],		['verI', 149],		['veriF', 149],
        ['print#', 152],	['pR', 152],		['prI', 152],		['priN', 152],		['prinT', 152],
        ['right$', 201],	['rI', 201],		['riG', 201],		['rigH', 201],		['righT', 201],
        ['input', 133],		//No abbreviation
        ['gosub', 141],		['goS', 141],		['gosU', 141],
        ['print', 153],		['?', 153],
        ['close', 160],		['clO', 160],		['cloS', 160],
        ['left$', 200],		['leF', 200],		['lefT', 200],
        ['next', 130],		['nE', 130],		['neX', 130],
        ['data', 131],		['dA', 131],
        ['read', 135],		['rE', 135],
        ['goto', 137],		['gO', 137],		['goT', 137],
        ['stop', 144],		['sT', 144],		['stO', 144],
        ['wait', 146],		['wA', 146],		['waI', 146],
        ['load', 147],		['lO', 147],		['loA', 147],
        ['save', 148],		['sA', 148],		['saV', 148],
        ['poke', 151],		['pO', 151],		['poK', 151],
        ['cont', 154],		['cO', 154],		['coN', 154],
        ['list', 155],		['lI', 155],		['liS', 155],
        ['open', 159],		['oP', 159],		['opE', 159],
        ['tab(', 163],		['tA', 163],		['taB', 163],
        ['spc(', 166],		['sP', 166],		['spC', 166],
        ['then', 167],		['tH', 167],		['thE', 167],
        ['step', 169],		['stE', 169],
        ['peek', 194],		['pE', 194],		['peE', 194],
        ['str$', 196],		['stR', 196],
        ['chr$', 199],		['cH', 199],		['chR', 199],
        ['mid$', 202],		['mI', 202],		['miD', 202],
        ['end', 128],		['eN', 128],
        ['for', 129],		['fO', 129],
        ['dim', 134],		['dI', 134],
        ['let', 136],		['lE', 136],
        ['run', 138],		['rU', 138],
        ['rem', 143],		//No abbreviation
        ['def', 150],		['dE', 150],
        ['clr', 156],		['cL', 156],
        ['cmd', 157],		['cM', 157],
        ['sys', 158],		['sY', 158],
        ['get', 161],		['gE', 161],
        ['new', 162],		//No abbreviation
        ['not', 168],		['nO', 168],
        ['and', 175],		['aN', 175],
        ['sgn', 180],		['sG', 180],
        ['int', 181],		//No abbreviation
        ['abs', 182],		['aB', 182],
        ['usr', 183],		['uS', 183],
        ['fre', 184],		['fR', 184],
        ['pos', 185],		//No abbreviation
        ['sqr', 186],		['sQ', 186],
        ['rnd', 187],		['rN', 187],
        ['log', 188],		//No abbreviation
        ['exp', 189],		['eX', 189],
        ['cos', 190],		//No abbreviation
        ['sin', 191],		['sI', 191],
        ['tan', 192],		//No abbreviation
        ['atn', 193],		['aT', 193],
        ['len', 195],		//No abbreviation
        ['val', 197],		['vA', 197],
        ['asc', 198],		['aS', 198],
        ['if', 139],		//No abbreviation
        ['on', 145],		//No abbreviation
        ['to', 164],		//No abbreviation
        ['fn', 165],		//No abbreviation
        ['or', 176],		//No abbreviation
        ['go', 203],		//No abbreviation
		
        ['+', 170],
        ['-', 171],
        ['*', 172],
        ['/', 173],
        ['^', 174],
        ['>', 177],
        ['=', 178],
        ['<', 179],
        ['{pi}', 255]
    ];
```
```
	const SPECIAL = [
		['{rvs off}',			0x92],	['{reverse off}',	0x92],
		['{rvs on}',			0x12],	['{reverse on}',	0x12],
		['{up}',			0x91],
		['{down}',			0x11],
		['{left}',			0x9d],
		['{rght}',			0x1d],	['{right}',		0x1d],
		['{clr}',			0x93],	['{clear}',		0x93],
		['{home}',			0x13],
		['{del}',			0x14],	['{delete}',		0x14], 
		['{blk}',			0x90],	['{black}',		0x90],
		['{wht}',			0x05],	['{white}',		0x05],
		['{red}',			0x1c],	
		['{cyn}',			0x9f],	['{cyan}',		0x9f],
		['{pur}',			0x9c],	['{purple}',   		0x9c],
		['{grn}',			0x1e],	['{green}',    		0x1e],
		['{blu}',			0x1f],	['{blue}',     		0x1f],
		['{yel}',			0x9e],	['{yellow}',   		0x9e],		
		['{org}',			0x81],	['{orange}',		0x81],
		['{brn}',			0x95],	['{brown}',      	0x95],
		['{lred}',			0x96],	['{pink}',     		0x96],
		['{dgry}',			0x97],	['{dark gray}',     	0x97],
		['{mgry}',			0x98],	['{gray}',     		0x98],
		['{lgrn}',     			0x99],	['{light green}',	0x99],
		['{lblu}',     			0x9a],	['{light blue}',	0x9a],
		['{lgry}',     			0x9b],	['{light gray}',	0x9b],		
		['{arrow left}',		0x5f],
		['{^}',				0x5e],
		['{pi}',			0xff],
		['{cm z}',			0xad],	['{173}', 		0xad],
		['{cm s}',			0xae],	['{174}', 		0xae],
		['{cm a}',			0xb0],	['{176}', 		0xb0],
		['{cm x}',			0xbd],	['{189}', 		0xbd]
];
```
```
        // Check if character code is less than or equal to '@' or is '{pi}' or is '[' or ']'
        if (o <= '@'.charCodeAt(0) || o === '{pi}'.charCodeAt(0) || o === '['.charCodeAt(0) || o === ']'.charCodeAt(0)) {
            return o;
        }
```
```
            <p>Here, you may enter a BASIC program (write in lower case), and upload it directly to your machine.</p>
```

radius75:
**The changes introduced here are an evolution and continuation of my comments in https://github.com/GideonZ/1541ultimate/issues/450. Detailed technical descriptions of specific modifications are also included in the commit messages.**

## Description
This PR significantly improves the user experience (UX) and accessibility of the Ultimate64/II BASIC tokenizer. These changes make utilizing the full PETSCII range intuitive through a modern visual interface, a flexible tagging system, and a modular configuration architecture.
obraz

### Key Enhancements:

1. **Interactive PETSCII Table & Tooltips:** 
   - Integrated a clickable table (0x00-0xFF) directly into `index.html`.
   - Added tooltips providing character descriptions and physical C64 key combinations.
   - Clicking an item in the table allows for instant insertion of characters or sequences (e.g., `{delete}`) into the editor.
2. **Advanced Tagging System:** 
   - Users can now input special characters in two ways: by **typing manually** (e.g., `{white}`, `{sh a}`) or by **selecting from the provided table**.
   - This system mimics modern cross-development standards found in tools like CBM prg Studio.
3. **Modular Configuration System:**
   - **CBMprgStudio.js (New Default):** A new mapping set based on CBM prg Studio is now enabled by default, providing the familiar `{}` syntax.
   - ~~**Legacy Support:** The original internal mapping in `index.html` remains untouched, allowing maintainers to decide on its future development.~~
   - **SPECIAL.js (Control File):** A configuration file that lets users specify which `.js` mapping to load. ~~Leaving it empty reverts the tokenizer to the legacy internal mapping.~~
4. **Customizable Editor Content:** 
   - External `.js` files now allow users to define the default text (instructions or example BASIC code) that appears in the editor window upon startup.
5. **Abbreviations:**
   - Introduced BASIC command abbreviations, matching the original C64 shorthand syntax.   
6. **Improved Line Addressing & 0x00 Logic:**
  - **Refined the logic for calculating 'Next Line' pointers to ensure stability during direct memory injection for multi-line BASIC programs. This fix ensures that parsing for the current line stops upon encountering a 0x00 byte, while the pointer to the next line is correctly calculated based on the line's actual memory offset.
  - **Note for consideration: The tokenizer perhaps should continue parsing the line to its full extent even if a 0x00 byte is encountered; the correct end of the line should be determined by calculating the next line's address via link pointers. While the standard C64 editor might not create such lines, they can be manually constructed via POKE or memory manipulation to create non-standard BASIC structures. Supporting this would increase the tokenizer's flexibility for advanced users.
.

## Note on AI-assisted development & Future Refinements
- **AI Assistance:** Parts of this code were modified with the assistance of AI. A thorough review of the changes (especially the memory addressing and pointer logic) is recommended before merging.
- **Work in Progress (WIP):** Please note that **tooltips and custom clickable character sets may not be 100% final**. They may require future corrections if any tokenization inconsistencies or content errors are discovered during wider use.

## How to test
1. **Manual Entry:** Type tags like `{delete}` or `{white}` directly into the editor and verify correct tokenization.
2. **Interactive Table:** Select and click characters from the provided table to insert them into the code and verify the output.
3. **Tooltips:** Hover over table elements to check the accuracy of C64 key descriptions and shortcuts.
4. **Configuration & Refreshing:** Change the mapping set in `SPECIAL.js`, then refresh the page (**F5** and **Ctrl+F5**). Verify that the new table content, tooltips, and editor startup text load correctly.
5. **Stability:** Confirm that multi-line BASIC programs have correct link pointers in memory after injection.

**Sample test code in CBMprgStudio format:**
```!- --------------
!- CBM prg Studio
!- --------------
!-
!- userport i/o control
!-
!- clear screen, set uppercase and disable shift + c=
!- for default: set pb0-pb7 as inputs, disable outputs activity, set temp as 0
0 print"{clear}{uppercase}{switch off}":poke56579,0:poke56577,0:te=0
!- decimal to binary precalculation
5 dimb$(255),c$(15)
10 fora=0to15:readc$(a):next
15 a=0:forh=0to15:forl=0to15
20 b$(a)=c$(h)+c$(l):a=a+1:nextl,h
25 data "0000","0001","0010","0011","0100","0101","0110","0111"
30 data "1000","1001","1010","1011","1100","1101","1110","1111"
!- screen
35 print"{clear}";
!-
50 print "          {white}userport i/o control          ";
51 print "                                        ";
52 print "               pb7>{gray}B{white}76543210{gray}B{white}<pb0       ";
53 print "{gray}{sh asterisk*12}CCCCCCC{sh +}{sh asterisk*8}{sh +}CCCCCCCCCCC";
54 print "{white}port current status{gray}B{white}--------{gray}B {white}---  56577";
55 print "{gray}{sh asterisk*12}CCCCCCC{cm w}{light gray}^^^^^^^^{gray}{cm q}CCCCCCCCCCC";
56 print "      {white}entered value{gray}B{white}--------{gray}B {white}---  56577";
57 print "{gray}{sh asterisk*13}CCCCCC{cm w}{light gray}^^^^^^^^{gray}{cm q}CCCCCCCCCCC";
58 print " {white}set port as in/out{gray}B{white}--------{gray}B {white}---  56579";
59 print "{gray}{sh asterisk*12}CCCCCCC{cm e}{sh asterisk*8}{cm e}CCCCCCCCCCC";
60 print "                                        ";
61 print "{white}set line as in-{reverse on}0{reverse off} or out-{reverse on}1{reverse off}, press {reverse on}1{reverse off}-{reverse on}8{reverse off} key";
62 print "   change output state, press {reverse on}q{reverse off}-{reverse on}i{reverse off} key   ";
63 print "{light gray}if the pb line is set as an output then ";
64 print " it has a higher short-circuit current. ";
65 print " do not connect external out to pb out. ";
66 print "                                        ";
67 print "        {black}g {red}+                   {black}g         ";
68 print "        n {red}5               {pink}9 9 {black}n         ";
69 print "        d {red}v               {pink}v v {black}d         ";
70 print "       {white}{cm a}{reverse on}{cm o}{reverse off}{cm r}{reverse on}{cm o}{reverse off}C{reverse on}{cm o}{reverse off}C{reverse on}{cm o}{reverse off}C{reverse on}{cm o}{reverse off}C{reverse on}{cm o}{reverse off}C{reverse on}{cm o}{reverse off}C{reverse on}{cm o}{reverse off}C{reverse on}{cm o}{reverse off}C{reverse on}{cm o}{reverse off}{cm r}{reverse on}{cm o}{reverse off}C{reverse on}{cm o}{reverse off}{cm s}        ";
71 print "       {cm z}{reverse on}{cm u}{reverse off}{cm e}{reverse on}{cm u}{reverse off}C{reverse on}{cm u}{reverse off}C{reverse on}{cm u}{reverse off}C{reverse on}{cm u}{reverse off}C{reverse on}{cm u}{reverse off}C{reverse on}{cm u}{reverse off}C{reverse on}{cm u}{reverse off}C{reverse on}{cm u}{reverse off}C{reverse on}{cm u}{reverse off}{cm e}{reverse on}{cm u}{reverse off}C{reverse on}{cm u}{reverse off}{cm x}        ";
72 print "        {black}g   {green}p p p p p p p p   {black}g         ";
73 print "        n   {green}b b b b b b b b   {black}n         ";
74 print "        d   {green}0 1 2 3 4 5 6 7   {black}d";

!- main loop
165 geta$:ifa$<>""thena=val(a$)-1:a2=2^a:gosub200
170 se=peek(56579):cu=peek(56577)
175 print"{home}{white}{down}{down}{down}{down}";
180 print,,b$(cu),"{left}";cu;"{left}  "
185 print,,b$(te),"{left}";te;"{left}  "
190 print,,b$(se),"{left}";se;"{left}  "
195 goto165
200 ifa<0ora>7then215
205 if(seanda2)=a2thenpoke56579,se-a2:return
210 poke56579,seora2:return
215 ifa$="q"thena2=2^0:goto260
220 ifa$="w"thena2=2^1:goto260
225 ifa$="e"thena2=2^2:goto260
230 ifa$="r"thena2=2^3:goto260
235 ifa$="t"thena2=2^4:goto260
240 ifa$="y"thena2=2^5:goto260
245 ifa$="u"thena2=2^6:goto260
250 ifa$="i"thena2=2^7:goto260
255 return
260 if(teanda2)=a2thente=te-a2:poke56577,te:return
265 te=teora2:poke56577,te:return
!- end main loop
270 rem radius75, 2024
```

Closes #450


GideonZ:
The problem with this commit is, that the extra files are not copied to the Flash during an update. So this will give problems.

radius75:
I assume it's necessary to add entries in:
update_u2p/update_u64ii.cc
update_u2pl/update.cc
update_u2p/update_u64.cc
update_u2p/update.cc
rv_update_u2/update.cc

GideonZ:
Yes, but then the files also need to be converted to an object file in the makefile. 

Another way to do it, is to put all the functions in index.html

Gee-64:
You could  consider keeping the simple delivery mechanism of a single html file and instead introducing a build step that uses webpack, rollup or some other tool to create the html file. This gives a nice modular developer experience, while at the same time not adding complexity to the updater.

chrisgleissner:
Hi @radius75 ,

I see there has been some extensive discussion on this interesting PR but it then ended seven months ago. We recently had an intense sprint in the run-up to release 3.15. I think you may already have missed that release, but are you interested in still seeing it merged in a subsequent release?

If so, is there anything missing or is this tested and merge ready?

Also, is there any reason for the all upper-case name `SPECIAL.js` instead of `special.js`?

Thanks
Christian

radius75:
Hi Christian,

Sorry for the delayed reply, I missed your message.
I do not expect this editor version to be merged into 
the official firmware distribution file. 
However, if you copy or replace the index.html, CBMprgStudio.js, 
and SPECIAL.js files on your device, everything should 
work perfectly. I haven't noticed any major changes to the index file, 
so mine should still be up to date with the latest version.
As for what is missing, it could use some verification and better 
token descriptions in the tooltips. 
The uppercase letters in the filename are purely cosmetic 
and don't matter to me.

Best regards

radius75:
Hi,

I noticed some changes in the index.html file from the test-merge branch, 
which I pulled into my own branch, and that's when I encountered the issue.

After merging, the page stops working entirely on Mozilla Firefox (tested on Windows 7), 
though it still functions correctly in Opera. 
I haven't tested whether it also fails on Firefox under Windows 10 or Windows 11. 

The issue seems to be related to the a:-webkit-any-link selector in the JavaScript code, 
which appears to cause a critical crash in Firefox.

Best regards,

radius75:
However, if it works fine for you on Firefox in Windows 10 and 11, we can ignore this issue and put it down to outdated browsers on Windows 7.

radius75:
I checked that the website also stopped working on Win10 in Mozilla Firefox.

--
The issue is caused by jQuery failing to parse the :-webkit-any-link selector in the website's script
5744a9205b65bb95abec148c5727e1470952d764

radius75:
@chrisgleissner I fixed the Firefox compatibility issues on my side in my latest commit.

chrisgleissner:
Hi @radius75 , thanks for the test. Unfortunately the PR fails the build. Also, it needs to target test-merge instead of master. Please let me know if you want to change that, otherwise I can help to repoint it.

radius75:
Hi @chrisgleissner ,

As I mentioned a few days ago, this tokenizer code is not intended to be included in the OFW.

Gideon said that he doesn't see much point in distributing it this way, and as far as I know, nothing has changed on that matter. Therefore, whether this branch currently builds successfully or can be merged with master is a secondary issue.

To use this tokenizer, you simply need to copy all the files from the `/html/` folder to your Ultimate device.

BTW, the build failed because instead of generating a new SHA, I set "unsafe-inline" just to check if the theme persistence would work with api.html after reloading the page with Ctrl+F5.

If you want to get the Firefox-related fixes, you can cherry-pick the commits from this branch into `test-merge`.

chrisgleissner:
Hi @radius75,

Thank you for flagging the Firefox issue. It was very helpful, and the fix is now included in #830

Could you let us know what you would prefer to do with this PR? We would be happy to keep it open if there is a realistic path to merging it later. Equally, if it is not something we expect to take forward, closing it would keep things tidy while still leaving the work and discussion visible.

I just wanted to check what makes most sense from your perspective.

Thanks again,
Christian

chrisgleissner:
Hi @radius75,

Would you mind raising a new PR with just the fixed of the bugs you spotted, targeting test-merge? If I don't hear from you by tomorrow I will cherry pick your changes.

Thanks
Christian

radius75:
@chrisgleissner 
Do you mean the fixes I introduced in my branch for the dark/light theme?

chrisgleissner:
@radius75 Yes, basically all changes since you merged test-merge into your branch, from `ece2dac6dc02e7766af8b20c7b0d875ffbba02f6` onwards (inclusive).

It would be nice to have these in their own PR, targeting the test-merge branch, so we can just merge it in. Please keep changes minimal and avoid reformatting as we are close to the 3.15 release and I'd like to keep risk low.

If it's too much effort, please let me know and I can cherry pick your changes. So far I only applied one of the changes, but I saw you made several, and wanted to onboard them all.

Thanks.

radius75:
@chrisgleissner 
Done :)

chrisgleissner:
Thank you @radius75 ! Much appreciated.

radius75:
Hi @chrisgleissner 

The `basic-tokenizer-editor-abbreviation branch` is up to date with the latest `test-merge`. 
Compare: https://github.com/GideonZ/1541ultimate/compare/test-merge...radius75:1541ultimate:basic-tokenizer-editor-abbreviation

If you decide to merge it, please note that it currently lacks an installer procedure to extract the files to the `/html` directory on the device. I previously made a commit 74c5006c3e921519ad15e978921e1b5d1975151f with a temporary workaround for this (since reverted it).

I will leave the final decision up to you on how you would like to handle this piece of code! :)

Best regards

chrisgleissner:
Reviewed the branch and pushed four commits to it. The base is now `test-merge` rather than `master`, which is where this repository takes pull requests.

Thanks for the profile mechanism: moving the tables out of `index.html` and letting a user pick a mapping is a real improvement, and the tooltips make the table usable without a PETSCII chart beside the keyboard. Three defects, all in the new code paths, and two points that need your decision.

## Fixed

**1. A tag clicked in the table left the caret in the wrong place.** `$editor.setSelectionRange(newCursorPos, newCursorPos)` calls a DOM method on a jQuery object, which throws `TypeError`. The insert itself had already happened, so the tag appeared, but the caret was left wherever setting the value put it, at the end of the text. Clicking a second tag then inserted it in a different place than the first. Changed to `$editor[0].setSelectionRange(...)`, matching the `$editor[0].focus(...)` line above it.

**2. Uploading with no profile loaded sent a wrong program without saying so.** `TOKENS` and `SPECIAL` now only exist if a profile assigns them. With `SCRIPT_TO_LOAD` empty, `finalizeInitialization()` sets both to empty arrays, so `print` is no longer matched as a keyword and is uploaded as five plain characters. With a profile that fails to load, the 3 second timeout deliberately does not call `finalizeInitialization()`, so both names stay undefined and Upload throws `ReferenceError` in the console with nothing shown on the page. `parseBASIC()` now checks that both tables are non-empty and reports `No tokenizer profile is loaded. Check SCRIPT_TO_LOAD in /Flash/html/special.js.` instead of uploading or throwing.

**3. Uploading an editor with no numbered line threw.** The new line filter `/^\d/` drops everything that does not start with a digit, which is what makes the instruction block work, but it can also produce zero lines: the instruction block on its own, the failure text the textarea now starts with, or an editor emptied with the new Clear button. `tokenized_lines[i-1].next_addr = ...` then sets a property on `undefined`, so Upload did nothing and said nothing. That case is now refused with a message.

## Needs your decision

**The legacy fallback described in the pull request no longer exists.** The description says leaving `SPECIAL.js` empty reverts the tokenizer to the legacy internal mapping, but the `const TOKENS` and `const SPECIAL` blocks were removed from `index.html`, so there is nothing to revert to. Either restore those tables as the fallback, or correct the description and the comment in `special.js`. Until then the empty setting is a refusal with a message rather than a working legacy mode.

**Tokenizing is now case sensitive.** `parseBASIC()` no longer lower cases the input, so upper case is Shift, which is what the abbreviations and the shifted graphics characters need. The consequence is that an existing listing in upper case, which is how most published C64 sources are written, no longer tokenizes: every keyword becomes shifted graphics characters and the upload succeeds with a program that is not the one that was pasted. The profile documents the new rule, but the description does not present it as a change for existing users. Worth stating there.

## Also pushed

**File names.** `CBMprgStudio.js` and `SPECIAL.js` are now `cbm_prg_studio.js` and `special.js`, with the references in `index.html` and in both files updated. Everything else in the tree is lower case with underscores.

**A test suite for the editor.** `tests/e2e/web/tokenizer_test.py`, registered as `web-tokenizer`, drives the page in Chrome and Firefox against the existing stub device and reads the bytes the page posts to `machine:writemem`, which is the part no one can see in the browser. Nine checks per browser: the profile loads and fills the editor and the table; a program using tags, the `?` and `goS` abbreviations and `{space*2}` arrives as the exact bytes of a BASIC program including its link pointers; `{null}` ends the line and the rest of that line is dropped; a clicked tag lands at the caret and the caret ends after it; and the three refusals above. All 18 checks pass in both browsers.

Each fix was verified by reverting it and seeing the matching check fail: the caret check reports the caret at 20 instead of 17, the profile check reports that a program was uploaded with no keyword table, and the numbered-line check reports that nothing was sent and nothing was said.

chrisgleissner:
Hi @radius75 ,

we are now trying to add tests for all new functionality, so as part of a review and fix cycle, I have also added some tests.

Please have a look at the comments under "Needs your Decision" above to see whether these changes are intentional or not, especially the one around the casing change.

No rush - we are not going to merge this into 3.15 as the time was too tight, so there is time to polish this change further if and as needed.

I also changed the casing of the files so they are in line with the rest of the project, as requested by Gideon.

Thanks
Christian

radius75:
**1. `SPECIAL.js` / tokenizer tables**

The description about falling back to the old built-in token mapping is outdated.

`special.js` specifies the external file containing the tokenizer tables. The old built-in `TOKENS` / `SPECIAL` mapping was intentionally removed, as it was much less complete than the token definitions used by CBMprgStudio.

If the tokenizer tables cannot be loaded correctly, there is no fallback to the old mapping. Instead, the editor displays the following message:

> Data loading failed: please try reloading the page (Ctrl+F5).
> Check for syntax errors in /Flash/html/special.js,
> or your custom profile if modified.

I will update the description/comment so that it reflects this behavior.

---
**2. Uppercase / case sensitivity**

The removal of the automatic `map(line => line.trim().toLowerCase())` of the tokenizer input is intentional.

The previous tokenizer converted typed/pasted text to lowercase. This made it impossible to distinguish between a character entered normally and the corresponding character entered with Shift. This distinction is important for entering C64/PETSCII characters that depend on Shift.

The current implementation preserves the original case of the input and follows the approach used by CBMprgStudio.

The existing `uppercase checkbox` from master is still present. It only changes how the text is displayed in the editor; it does not modify the underlying input used by the tokenizer. Therefore Shift and non-Shift characters remain distinguishable even when the checkbox is enabled.

When the checkbox is disabled, the original case is displayed again — characters entered with Shift remain uppercase and those entered without Shift remain lowercase.

So this is an intentional change to the tokenizer input handling.

chrisgleissner:
Thanks for the explanations @radius75 .

As the change was peer reviewed, enriched with tests, and suggestions by Gideon about file naming were onboarded, I will now remove the 'Draft' flag, as it is ready for a wider review.

@GideonZ What are your thoughts about this PR?

radius75:
I decided to bring back `trim().toLowerCase()` as an optional, profile-dependent setting.

This may be useful for someone creating a profile for pasting BASIC code that is already written in uppercase and does not use shifted characters.

Personally, I currently handle such code by converting it to lowercase in Notepad++. With this option, the tokenizer can now handle that conversion itself.

The `cbm_prg_studio.js` profile keeps `NORMALIZE_TO_LOWERCASE = false`, since it relies on case-sensitive abbreviations and character mappings.

The commit description and the comments in the code document how this option works and when it should be used.

radius75:
The examples from the Commodore 64 Games Book 1 collection are particularly useful for testing this change because the programs were transcribed for use with CBMprgStudio.

The same applies to the other .bas programs in the repository, which provide additional real-world BASIC source prepared for use with CBMprgStudio.

https://github.com/TypeInBooks/books/tree/master/commodore/c64/gamesbook1

chrisgleissner:
@radius75 Sounds great. Thanks for also making sure all tests pass on CI.

radius75:
> Thanks for also making sure all tests pass on CI.

@chrisgleissner 
What kind of test supervision are you referring to?

I see that the test expects

`("all", *browser_lib.BROWSERS)`

but gets `("all",) + browser_lib.BROWSERS`

Can this be improved? Because the tokenizer works as is.

--
Only now, after reviewing tokenizer_test.py, do I fully understand how the tests work. I realize that my recent changes require updating the tests to cover the new behavior, including case handling, abbreviations, and normalization

chrisgleissner:
Hi @radius75 ,

I was referring to this broken build: https://github.com/GideonZ/1541ultimate/actions/runs/34208712219/job/102004311447#step:5:12

On a related note, in future, we will run both unit and e2e tests automatically on CI as part of each PR build. For the time being, it is the responsibility of anyone who raises a PR to run the unit tests (`make`) and E2E tests (`run-tests <host>`) against a locally built and flashed firmware. 

This ensures that merging in PRs does not regress functionality.

Btw, thanks *a lot* for having spotted the Firefox regression last week. Thanks to you I created a Selenium-based E2E test for the index page. The change went live with the 3.15 release yesterday. 

Best wishes
Christian

radius75:
Hi Christian,

I don't have a local firmware build environment on Windows 7. Instead, I install the latest generated master build and manually replace index.html, special.js, and cbm_prg_studio.js on the device.

This allows me to test the tokenizer functionality on real hardware, but I understand it doesn't replace the full local build and test workflow.

radius75:
Hi Christian,

Following our discussion about test coverage, here are two reference cases for the current tokenizer normalization behavior. They include the BASIC source, expected HEX output, and screenshots from the PC emulator and real C64.

These should help when extending the E2E tests to cover both normalization modes.

### Test 1: CBMprgStudio v4.10.0 profile compatibility (NORMALIZE_TO_LOWERCASE = false)

Verify that BASIC source using uppercase/lowercase keywords and supported abbreviations produces the same tokenized bytes as CBMprgStudio. The resulting HEX output from CBMprgStudio and Ultimate must be identical.
```
10 PRINT"AB":GOTO20
20 print"ab":goto30
30 ?"Ab":gO40
40 rem Ab"Ab"
50 data Ab,"Ab"
60 readA$,b$
70 printA$,b$

!- >C:0801  0f 08 0a 00  99 22 c1 c2  22 3a 89 32  30 00 1d 08   
!- >C:0811  14 00 99 22  41 42 22 3a  89 33 30 00  2b 08 1e 00   
!- >C:0821  99 22 c1 42  22 3a 89 34  30 00 38 08  28 00 8f 20   
!- >C:0831  41 42 22 c1  42 22 00 46  08 32 00 83  20 41 42 2c   
!- >C:0841  22 c1 42 22  00 51 08 3c  00 87 41 24  2c 42 24 00   
!- >C:0851  5c 08 46 00  99 41 24 2c  42 24 00 00  00
```
Screenshot from the PC emulator showing the resulting program and its BASIC listing. The HEX output is compared byte for byte with CBMprgStudio.
false




---
### Test 2: Legacy full-source normalization (`NORMALIZE_TO_LOWERCASE = true`)

Verify that the entire BASIC source is converted to lowercase before tokenization, including quoted strings and text following REM and DATA, preserving the behavior of the previous tokenizer version.

```
10 PRINT"AB":GOTO20
20 print"ab":goto30
30 PrInT"Ab":gOtO40
40 rEm Ab"Ab"
50 Data Ab,"Ab"
60 reaDA$,b$
70 printA$,b$

0F 08 0A 00 99 22 41 42 22 3A 89 32 30 00 1D 08 
14 00 99 22 41 42 22 3A 89 33 30 00 2B 08 1E 00 
99 22 41 42 22 3A 89 34 30 00 38 08 28 00 8F 20 
41 42 22 41 42 22 00 46 08 32 00 83 20 41 42 2C 
22 41 42 22 00 51 08 3C 00 87 41 24 2C 42 24 00 
5C 08 46 00 99 41 24 2C 42 24 00 00 00
```
Photo of the program running on a real C64, showing the resulting BASIC listing after full-source lowercase normalization.
true

radius75:
The latest fix corrects : handling in DATA statements, allowing tokenization to resume after the colon. It also prevents $83 (DATA) and $8F (REM) bytes inside quoted strings from changing the tokenizer state.

Tested against CBMprgStudio — the generated BASIC bytes are identical.
The test can be performed with both NORMALIZE_TO_LOWERCASE = false and NORMALIZE_TO_LOWERCASE = true.
```
10 data 1,2 :print "after data"
20 rem test :print "after rem"
30 print "{131}":print "after 131"
40 print "{143}":print "after 143"

!- >C:0801  1b 08 0a 00  83 20 31 2c  32 20 3a 99  20 22 41 46 
!- >C:0811  54 45 52 20  44 41 54 41  22 00 39 08  14 00 8f 20 
!- >C:0821  54 45 53 54  20 3a 50 52  49 4e 54 20  22 41 46 54 
!- >C:0831  45 52 20 52  45 4d 22 00  51 08 1e 00  99 20 22 83 
!- >C:0841  22 3a 99 20  22 41 46 54  45 52 20 31  33 31 22 00
!- >C:0851  69 08 28 00  99 20 22 8f  22 3a 99 20  22 41 46 54 
!- >C:0861  45 52 20 31  34 33 22 00  00 00
```
obraz

radius75:
Tests the BASIC line number limit introduced in the latest commit. Line numbers up to 65535 are accepted, while 65536 and 70000 are rejected to prevent 16-bit truncation and match CBMprgStudio behavior.
```
0 print "0"
1 print "1"
65535 print "65535"
65536 print "65536"
70000 print "70000"
```

radius75:
Regarding the latest commit limiting the number of special character repetitions, I tested the following program:
```
10 print "{space*999}"
20 print "{space*1000}"
30 print "{space*900}{space*99}{space*1}"
40 print "{999 space}"
50 print "{1000 space}"
60 print "{900 space}{99 space}{1 space}"
```
Each line tests a different case:

10 — a single CBMprgStudio-style repetition at the safety limit (999).
20 — a single CBMprgStudio-style repetition exceeding the limit (1000).
30 — multiple CBMprgStudio-style repetitions that individually stay within the limit, but exceed it when combined (900 + 99 + 1 = 1000).
40 — a single magazine-style repetition at the safety limit (999).
50 — a single magazine-style repetition exceeding the limit (1000).
60 — multiple magazine-style repetitions that individually stay within the limit, but exceed it when combined (900 + 99 + 1 = 1000).

This verifies both supported repetition syntaxes as well as the cumulative repetition-count check.

The safety limit prevents the browser from attempting to expand extremely large repetition counts, which could consume excessive memory and cause the page or browser tab to become unresponsive.

--
dd47dfc only adds a safety limit to special character repetition expansion. It is not the BASIC line length limit — the final tokenized line length is still checked separately and limited to 256 bytes.

radius75:
Hi @chrisgleissner 

I've done quite a lot of compatibility testing of the Ultimate tokenizer,
comparing its output and behavior with the CBM prg Studio tokenizer and
with a real C64. So far, these tests have been successful and I haven't
found any significant compatibility problems that would require changes
to the current tokenizer.

As an additional source of real-world test programs, I've been using the
listings from the two Commodore 64 Games Book volumes:

https://archive.org/details/commodore-64-games-book_202304/mode/1up

https://archive.org/details/commodore-64-games-book-2_202409/mode/2up

Transcribed versions of these and many other C64 book listings can also
be found here:

https://github.com/TypeInBooks/books/tree/master/commodore/c64

One thing worth noting is that the descriptions/instructions for some
games do not always match what the programs actually do. However, in the
cases I've checked, these discrepancies were already present in the
original books and are not tokenizer or transcription errors.
The code in the GitHub repository is about eight years old and was most
likely created using a CBM prg Studio 3.1x-era version. This should also
be kept in mind when using it as a compatibility reference, since the
current CBM prg Studio version is 4.10 and its tokenizer may have changed
in various ways since then.

One more question regarding the PR: should I now keep this branch updated
against `master` instead of `test-merge`?

chrisgleissner:
Hi @radius75 ,

excellent work!

About the target branch:

- The `test-merge` branch has been retired and should no longer be used.
- All PRs should target `master`.
- We will continue to cut releases from `master` for fixing 3.15 bugs, e.g. 3.15a (already cut), 3.15b, etc.
- Once all major issues on 3.15 have been fixed, in particular regression issues, we will repurpose `master` to accept larger work (new features etc.) in preparation for release 3.16.
- We may at that point create a `releases/3.15` branch for fixing 3.15 issues that cannot wait for 3.16. The need for that is likely to be small if we continue to focus on 3.15 fixes for now.

This is as per the Github Flow model aka trunk based development: Only a single `master` which is always release-ready.

The above has been discussed with @GideonZ (except for the `releases/3.15` part), but of course it is ultimately his call.

Thanks
Christian

UltimateC64Ultimate64fan:
@chrisgleissner wrote:

> About the target branch:
> 
> * The `test-merge` branch has been retired and should no longer be used.
> * All PRs should target `master`.
> * We will continue to cut releases from `master` for fixing 3.15 bugs, e.g. 3.15a (already cut), 3.15b, etc.
> * Once all major issues on 3.15 have been fixed, in particular regression issues, we will repurpose `master` to accept larger work (new features etc.) in preparation for release 3.16.
> * We may at that point create a `releases/3.15` branch for fixing 3.15 issues that cannot wait for 3.16. The need for that is likely to be small if we continue to focus on 3.15 fixes for now.
> 
> This is as per the Github Flow model aka trunk based development: Only a single `master` which is always release-ready.
> 
> The above has been discussed with @GideonZ  (except for the `releases/3.15` part), but of course it is ultimately his call.

In a managed open-source repository, retiring a primary development branch and migrating to Trunk-Based Development mandates:

1. **A Root CONTRIBUTING.md:** Clearly documenting the target branch (master), commit conventions, and testing requirements.
2. **Branch Protection Enforcement:** Enabling GitHub branch protection on master to enforce the Four-Eyes Principle so that master truly remains "always release-ready".
3. **Formal Communication:** Posting an official pinned announcement or updating the repository README.md so the entire community is aligned.

PR #642, opened 2026-02-14
html: Live Monitor fixes, Power Button, Disk Image mounting

TangoBravo64:
A number of smaller fixes and additions, split into smaller commits:

- Live Monitor: addresses containing an "e" did not work (e.g. "m 1e0"), as the terminal interpreted these as decimal numbers written in "exponential form" (1E0 = 1, 1E3=1000, ...).
- Adds "power off" option to the nav bar for supported boards (Ultimate 64 II)
- Adds "firmware version" to the nav bar (read from /v1/info).
- Remote player now accepts and mounts D64/D71/D81 files (always mounted in drive A).
- Other smaller fixes.

poweroff

livemonitor

### DIFF
--- a/html/index.html
+++ b/html/index.html
@@ -230,9 +230,10 @@ $(document).ready(async function () {
     }
 
     // PRG/CRT(/SID) loader
-    async function submitRunner() {
-        let body = $('#runnerfile')[0].files[0];
-        let extension = getFileExtension(body.name);
+    async function submitRunner(file, drive) {
+        if (drive === undefined) drive = "a";
+        if (file === undefined) file = $('#runnerfile')[0].files[0];
+        let extension = getFileExtension(file.name);
         let params = {};
         $("#runmsg").hide();
 
@@ -243,14 +244,28 @@ $(document).ready(async function () {
             route = "/v1/runners:run_crt";
         else if (extension === "SID")
             route = "/v1/runners:sidplay";
+        else if (extension === "D64") {
+            route = `/v1/drives/${drive}:mount`;
+            params = {"type":"d64"};
+        }
+        else if (extension === "D71") {
+            route = `/v1/drives/${drive}:mount`;
+            params = {"type":"d71"};
+        }
+        else if (extension === "D81") {
+            route = `/v1/drives/${drive}:mount`;
+            params = {"type":"d81"};
+        }
         else
             alert("Unsupported file extension: '."+ extension.toLowerCase()+"'");
 
         if (route !== null) {
-            let [status_code, content] = await make_post_request("http://" + serverIP + route, params, body);
+            let [status_code, content] = await make_post_request("http://" + serverIP + route, params, file);
             if(status_code !== 200)
                 $("#runmsg").show();
-	}
+            return status_code == 200;
+        }
+        return false;
     }
 
     // BASIC/tokenizer page event handlers
@@ -1644,7 +1659,7 @@ function disassembler(startInt, byteArray) {
                 <br />
                 or press button to browse
                 <br /><br />
-                <input type="file" id="runnerfile" name="file" accept=".prg,.crt,.PRG,.CRT,.sid,.SID" title="Select PRG/CRT/SID file to be run">
+                <input type="file" id="runnerfile" name="file" accept=".prg,.crt,.sid,.d64,.d71,.d81,.PRG,.CRT,.SID,.D64,.D71,.D81" title="Select PRG/CRT/SID file to be run">
                 <br /><br />
                 <input type="button" id="runnersubmit" value="Run Again!" title="Upload and RUN the selected file again">
             </div>
--- a/html/index.html
+++ b/html/index.html
@@ -270,6 +270,8 @@ $(document).ready(async function () {
     $('#uppercaseBASIC').on('change', async function (event) {
         $("#basiceditor").toggleClass("uppercase", this.checked);  // change upper/lowercase display
     });
+    // browser may have cached the checkbox state on a page reload: synchronize the initially displayed state
+    $("#basiceditor").toggleClass("uppercase", $('#uppercaseBASIC').prop('checked'));
 
     // convenience: write memory - show error message when access fails
     async function writeMemory(param, errorMsg) {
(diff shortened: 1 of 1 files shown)


chrisgleissner:
Hi @TangoBravo64 ,

thanks for having raised this PR. 

Reading through your change list, it seems to me that these bug fixes and improvements would be really useful to have in the new firmware version. In particular the inability to view hex addresses containing an `e` seems quite a large bug to me, and it's great this is fixed in your PR based on your PR description.

I know that @GideonZ  is currently working on 3.15 and merged a bunch of PRs into the `test-merge` branch. 

If you are interested in getting this merged, it may be worth to redirect its merge target to `test-merge` (instead of `master`) and to also perform a final test, as a lot of changed have occurred in `test-merge` since you created this. I do not see anything that could result merge conflicts or clashes with your changes, but just to be on the safe side, it may be useful.

Also, once you are done with the tests, could you please update here which devices you tested it on, and how you tested it?

Thanks
Christian

chrisgleissner:
Hi @TangoBravo64,

Thanks again for your work on this! I have now checked the JS against the firmware’s API, tested the local merge, and really appreciated how you split it into four clean, clear commits.

Overall, the code is in great shape! It’s not quite merge-ready yet, but the remaining work is very small. Here is what we need to get it across the finish line:

## Blocking Issues
- Merge Conflict: The branch needs a rebase onto test-merge. #646 (dark mode) recently updated html/index.html, causing a conflict with your CSS structure changes. Resolving this manually should be straightforward.
- Power-off Allow-list: Right now, ProductsWithPwrCtrl hides the power button on hardware that actually supports it. Firmware implements MENU_C64_POWEROFF for both U64 generations. Please add "Ultimate 64" and "Ultimate 64 Elite" to the list. 

## Optional Polish

If you happen to touch the code again, here are a few tiny cleanups (none of these block merging):

- HTML Markup: In the terminal header, wrap `<span class="title">` in a `<li>` so it's valid `<ul>` syntax.
- CSS Colors: Consider using `var(--color)` instead of the hardcoded `rgb(160, 182, 247)` to keep themes consistent.
- API Tweaks: You can drop the `legacy start = start.toString()` lines in `m` and `d`. You could also add `g64/g71` extensions to the mount tool if you want to expand support.
- Unused Param: `submitRunner(file, drive)` parameter drive is currently unused (always defaults to "a").

Thanks,
Christian

chrisgleissner:
Hi @TangoBravo64,

This is merged. I retargeted the PR to `test-merge` and merged it there as c7790507, then pushed the follow-ups below on top. Thanks again — the four-commit split made it easy to follow what each change was for.

## The merge

Two conflicts, both in `html/index.html`, against the dark-mode work that landed on `test-merge` in the meantime:

- #646 split the banner into an outer `#banner` and an inner `#bannertitle`, so that the theme buttons can sit beside the title. Your branch still writes the product string to `#banner`, which would have replaced those buttons with the title. It now goes to `#bannertitle`, with your power-off and firmware-version code kept beside it.
- Two sets of stylesheet rules added at the same place: the dark-mode hover colour for the navigation, and your `#firmware-version` and `#menu-power` rules. Both are kept.

## Follow-ups

**Power-off allow-list.** `GET /v1/info` reports `getProductString()`, which is one of the seven names in `product_name[]` in `software/system/product.cc`. `"C64 Ultimate"` is not one of them, so that entry never matched anything. `MENU_C64_POWEROFF` writes the power register when `U64 == 1`, asks the controller that owns the power rail when `U64 == 2`, and returns `SSRET_NOT_IMPLEMENTED` otherwise. The list is now `["Ultimate 64", "Ultimate 64 Elite", "Ultimate 64-II"]`.

**I was wrong about `var(--color)`.** I suggested it in my earlier review; your literal was right. `--color` is `rgba(160, 182, 247, 0.877)`, and `.trinity-dialog header::before` paints a `#251180` bar behind the title, which would show through the text at that alpha. The colour is now `--title-background`, declared beside the other variables of that dialog, so the value still lives in one place. The title is also a `<li class="title-item">`, so the header's `<ul>` holds only `<li>` elements, and it is positioned within the header rather than sharing the rule that positions the whole dialog. Rendering is unchanged: measured across four viewport widths, the title's box is identical to your branch's to the pixel, and the suite re-measures it in both browsers on every run.

**Mounting.** The device accepts `d64, g64, d71, g71, d81`, so G64 and G71 are in, along with the `accept` list on the file input. A mount that succeeds now shows a confirmation: running a program is visible on the C64, mounting a disk is not, and the only feedback the page had was the red box it shows on failure. The `file` and `drive` parameters of `submitRunner` are gone, since no caller passed either; the drive is a constant beside the list of image types.

**Smaller ones.** The firmware version goes in with `.text()` rather than `.html()`, since it is a string the device supplies. The two `start.toString()` lines in `m` and `d`, and the `end.toString()` in `d`, do nothing now that `processArguments` is off, so they are removed, and the option carries a comment explaining what parsing did to a hex address. The page heading, the file input and the navigation link named PRG and CRT only, and now name the disk images too.

## Tests

`tests/e2e/web/index_test.py` is new: 33 checks in Chrome and Firefox, no device needed. It drives the real page against `tests/lib/device_double.py`, the fake Ultimate the observability tests already use, which now also serves `html/`. The browser reaches it through an HTTP proxy, so one address serves both the page and the API, and the checks assert what the page sent rather than only what it drew.

Covered: the Power Off item against all seven product names, both ways round; the firmware version as text; the confirmation answered yes and no; each of the five image types through `POST /v1/drives/a:mount` with the type stated and the uploaded bytes intact; PRG, CRT and SID still reaching their runners; an unknown extension reaching no route at all; `m`, `d` and `f` asking for the address that was typed, which is the `1e00` case your first commit fixed; the Live Monitor's title bar inside its header; and the BASIC editor following a tick the browser restored on a reload.

Every check was confirmed to fail when the code it covers is reverted.

One thing I did not touch, because it predates this PR: `serverIP` is `window.location.hostname`, with no port, so the page's REST calls only reach the device when the page itself was served on port 80. That is always true on a device and never true anywhere else, which is why the test proxies the browser rather than pointing it at a high port. Worth a separate change at some point.

Christian

PR #646, opened 2026-02-22
Added dark mode support to index.html

danielwcarey:
Updated the index.html to default to system's selected light/dark mode selection and updated the banner to include toggle icons between light and dark. If the browser doesn't support theming, there should be no visible change, and the toggle icons won't appear.

I tried to make minimal edits but also provide something that may enable more themes if others are so inclined to pursue.

For reference, this is the snapshot of primary theme and then some screenshots of the dark theme 
when applied to several pages:

### DIFF
--- a/html/index.html
+++ b/html/index.html
@@ -21,7 +21,7 @@ $(document).ready(async function () {
         // This also works with older firmware that do not support password protection (will get a 404 for /v1/info)
         let [status_code, content] = await make_get_request("http://" + serverIP + "/v1/info");
         if (status_code == 401 || status_code == 403) {
-            $("#banner").html("&nbsp;");
+            $("#bannertitle").html("&nbsp;");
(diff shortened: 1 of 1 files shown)


chrisgleissner:
Hi @danielwcarey ,

this is a great addition for anyone who is either sensitive to bright backgrounds (such as myself) or whoever wants a consistent look between U64 menu system and HTTP menu.

@GideonZ is currently working on 3.15 and merged a bunch of PRs into the `test-merge` branch in preparation for the release. 

I see you have done extensive testing and also included screenshots, which is very useful. It may be worth to change the target branch to `test-merge` and also mention which device(s) you tested this on.

Thanks
Christian

danielwcarey:
Squashed the two commits and rebased on the branch, "test-merge".

Confirmed on Edge browser, Version 150.0.4078.65. The update uses base CSS and JavaScript and should work on other browsers.

PR #680, opened 2026-04-12
Fix USB HID status ownership and joystick-menu name display

chrisgleissner:
## Summary

Improve USB mouse and keyboard status handling in the joystick menu.

This fixes cases where the wrong device name could be shown after reconnects, and also makes sure the menu starts in a sensible state even if no USB HID devices have been connected yet.

## Changes

- Track the visible USB mouse and keyboard by device + interface instead of USB address
- Update the joystick menu status:
  - when the menu is created
  - when USB HID state changes
- Add host-side tests for:
  - device/interface matching
  - which device should be visible

## Display Logic

The joystick menu keeps exactly:
- one visible mouse
- one visible keyboard

### Selection rules

A newly detected device does not automatically replace the one that is currently shown.

A device is shown if one of the following applies:

1. Nothing is currently shown for that role
2. It matches the previously shown device/interface
3. It becomes the active device due to user input

### Activity-based priority

If multiple devices of the same type are connected:
- the displayed device is the one with the most recent activity
  - key press for keyboards
  - movement, button, or wheel input for mice

### Reconnect behaviour

Reconnects alone do not necessarily change what is shown.

This is intentional:
- A single physical device can expose both keyboard and mouse interfaces
- That device can continue to occupy one or both roles
- A newly connected device only takes over after actual user activity

## Testing

Manual testing was done on an Ultimate 64 Elite I with:

- Logitech G500s (wired mouse)
- Logitech MX Master 3 (via Unifying receiver)
- Keychron keyboard (wireless receiver)

Test scenarios included:
- connecting multiple devices at the same time
- repeated connect and disconnect sequences
- mixed setups with composite devices


chrisgleissner:
Hi @GideonZ ,

This PR fixes a few issues with rendering the connected USB devices.

Many thanks for merging,
Christian

PR #714, opened 2026-07-07
Keep freeze UI open during REST DMA

chrisgleissner:
## Summary

This PR fixes REST memory reads and writes while the Ultimate 64 is in **Freeze** mode.

`GET /v1/machine:readmem` and `PUT`/`POST /v1/machine:writemem` now keep the active freeze UI open instead of closing it before the DMA transfer, and the memory they expose while frozen matches the real C64 state as closely as the freeze mechanism allows.

## Benefits

This change benefits REST clients by making memory access less disruptive for the user:

* the freeze menu stays open
* the freeze monitor stays open
* REST clients can inspect or modify memory without interrupting the user
* Kernal, Basic, and cartridge ROM reads return the correct C64-visible contents
* bytes written while frozen persist correctly across unfreeze, for every range except screen RAM
* memory written in one UI mode (Freeze or Overlay on HDMI) reads back identically in the other

## What changed

The raw REST DMA path no longer unconditionally releases `c64->client`.

When the machine is frozen, memory access now goes through `C64::dma_transfer_frozen()`. This helper splits the transfer at three boundaries and handles each region according to what the freezer menu actually does with it:

* **Kernal / Basic / cartridge ROM** (`$8000`-`$CFFF`, `$E000`-`$FFFF`): the freezer menu may have its own ultimax cart banked in, so the helper briefly restores the C64 mode that was active when freeze was entered (with `C64_DMA_MEMONLY=0`), so hardware ROM/cart decode is used and the returned bytes match what the C64 would see, then restores the menu's mode.
* **`$0800`-`$0FFF`** (2KB the freezer menu repurposes as its own scratch RAM) and **`$D800`-`$DBFF`** (color RAM, which the menu also reuses for its own display): these are served from the `ram_backup`/`color_backup` arrays that `C64::unfreeze()` already restores onto real RAM when the menu closes. Reading/writing through the *live* (bypassed) memory here — the first version of this fix's approach — meant a REST write would appear to succeed but then get silently discarded the moment the user closed the menu, since `restore_io()` unconditionally overwrites both ranges from those backups. Routing through the same backup arrays means both reads and writes during freeze now reflect and affect the real, persisted C64 state.
* **Everything else, including screen RAM** (`$0400`-`$07FF`): unchanged, fast `C64_DMA_MEMONLY=1` bypass straight to the underlying RAM array. Screen RAM is the one range left showing the live Ultimate UI while frozen — see "REST API behaviour change" below.

## Why the new DMA helper is needed

Freeze mode is special because the menu or monitor owns the active C64 client. Releasing that client closes the UI.

However, simply keeping the UI open and forcing memory-only DMA for the whole transfer is also wrong, for two separate reasons:

1. `C64_DMA_MEMONLY=1` bypasses the C64 memory map and reads the underlying RAM array directly. That's valid for plain RAM, but not for Kernal, Basic, or cartridge ROM ranges, because those bytes aren't stored in the underlying RAM array at all.
2. The freezer menu doesn't just *display* over the C64 — it physically **reuses** `$0800`-`$0FFF` and `$D800`-`$DBFF` as its own scratch RAM and color RAM while frozen, backing up the real contents into `ram_backup`/`color_backup` first. A naive bypass read/write against the live (menu-owned) memory in those ranges works for reads by coincidence today, but writes are lost the instant the menu closes and `restore_io()` overwrites them from the (stale) backups.

`C64::dma_transfer_frozen()` gives us the behaviour we actually need: keep the freeze UI active, read mapped ROM/cart regions correctly, and make writes to the menu's scratch/color RAM survive unfreeze — all via a handful of large `memcpy` calls per contiguous region rather than a byte-at-a-time loop, so there's no meaningful performance cost versus the plain bypass path.

## REST API behaviour change

In Freeze mode, screen RAM (`$0400`-`$07FF`) access now reflects the active Ultimate UI if the menu or monitor is open, both for reads and writes. This is the one intentional exception, because that's the range the menu physically draws its own display into.

Previously (on `test-merge`), the REST call closed the UI first, so screen RAM was accessed only after the menu had already been torn down. If a client wants that old effective behaviour, the user can close the menu before making the REST memory request.

Everywhere else in the 64KB address space, including `$0800`-`$0FFF` and color RAM (`$D800`-`$DBFF`), REST reads and writes while frozen now behave exactly as if the menu were closed:

* Kernal and Basic ROM reads while frozen return the correct ROM contents instead of RAM underneath the ROM overlay.
* Writes to `$0800`-`$0FFF` and color RAM while frozen now persist across unfreeze, instead of being silently discarded.
* A byte written via REST in Freeze mode reads back identically via REST in Overlay-on-HDMI mode, and vice versa.

## Tested Hardware

I tested this PR on the following platforms:
- Ultimate 64 Elite I
- Ultimate II+L

## End-to-end test: `tools/api/readmem_writemem_test.py`

Added an end-to-end test that exercises the whole 64KB address space, across both `Interface Type` settings (`Freeze` and `Overlay on HDMI`), against real firmware. It follows the existing `tools/api/*_test.py` conventions (self-contained script, `RestSession` + `check()`/`Failure` helpers, `-H/-r/-p/-t/--test` CLI args, env var fallbacks).

**What it does**, per interface mode combination:

1. Writes a full-range, address-derived test pattern (`byte = (address XOR key) & 0xFF`) via `POST /v1/machine:writemem`, in one or two large chunks per test.
2. Reads the full 64KB back via `GET /v1/machine:readmem`.
3. Compares byte-for-byte, with three documented, deliberate exceptions:
   * **Screen RAM** (`$0400`-`$07FF`): expected to mismatch whenever Freeze mode is involved (the menu owns it); expected to match in an Overlay-only round trip.
   * **Color RAM** (`$D800`-`$DBFF`): compared on the low nibble only — it's physically 4-bit hardware, and the unconnected upper nibble reads back differently depending on access path/width, confirmed on real hardware. The low nibble is the only part that's real, and it matches exactly.
   * **VIC-II/SID (`$D000`-`$D7FF`) and CIA/cart I/O (`$DC00`-`$DFFF`)**: left untouched entirely (writing hardware registers blindly risks audible/visible side effects) and excluded from comparison (they're not RAM and are inherently time-varying).
   * **Basic (`$A000`-`$BFFF`) / Kernal (`$E000`-`$FFFF`) ROM**: in same-mode self-checks, a written pattern never "sticks" there because real ROM ignores writes — that's correct behaviour, so these are excluded from the self-check comparison. In cross-mode checks this isn't needed: the comparison is against a "ground truth" read captured right after the write (in the write-mode), which already reflects the real, unaffected ROM content on both sides.
4. Self-calibrates against ordinary system noise: before any writes, it takes several full-memory reads a fraction of a second apart while the C64 is running, and treats whatever addresses changed on their own (jiffy clock, IRQ-driven cursor blink, etc.) as expected background noise rather than failures. A mismatch found later that has *already changed again* on a quick re-read is treated the same way. A small, explicitly-logged tolerance for isolated single-byte drift is applied only to the Overlay self-check, where the C64 keeps running for the whole multi-second write+read (Freeze halts the CPU for the same window, so no such tolerance is needed or applied there).

Four checks, run against real hardware:

* `selfcheck-freeze`: write + read both while Freeze — verifies the fix doesn't just work in isolation, but round-trips as a whole (minus the screen exception).
* `selfcheck-overlay`: same, while Overlay on HDMI (the pre-existing, unmodified live/non-frozen path).
* `overlay-to-freeze`: write while Overlay on HDMI, read while Freeze — this and the next check are the core requirement this update addresses.
* `freeze-to-overlay`: write while Freeze, read while Overlay on HDMI.

#### Ultimate 64 Elite I

Command: `python3 tools/api/readmem_writemem_test.py --host u64`

Output:
```
[01] read current Interface Type config ... OK
  current Interface Type: Freeze
[02] reset C64 to a clean, stable state ... OK
[03] switch to Overlay on HDMI to probe live background activity ... OK
[04] probe addresses that change on their own while the C64 runs ... OK
  7 address(es) treated as expected live background noise: $00A1, $00A2, $00CD, $00CF, $01EC, $01F2, $0287
[05] set Interface Type = Freeze ... OK
[06] open menu (Freeze) ... OK
[07] write full-range pattern (Freeze) ... OK
[08] read back full range (Freeze) ... OK
[09] menu still open after write+read (Freeze) ... OK
--- selfcheck-freeze: write and read both happen in Freeze (screen RAM excluded, menu redraws it continuously) ---
  [selfcheck-freeze] ignored ROM mismatches (writes to real ROM are no-ops): 16330
  [selfcheck-freeze] screen ($0400-$07FF) mismatches: 908 (expected, menu owns this range while frozen)
  [selfcheck-freeze] ignored known-live-noise mismatches: 0
  [selfcheck-freeze] color RAM (low nibble) mismatches: 0
  [selfcheck-freeze] RAM mismatches: 0
[10] close menu (Freeze) ... OK
[11] set Interface Type = Overlay on HDMI ... OK
[12] open menu (Overlay on HDMI) ... OK
[13] write full-range pattern (Overlay on HDMI) ... OK
[14] read back full range (Overlay on HDMI) ... OK
[15] menu still open after write+read (Overlay on HDMI) ... OK
--- selfcheck-overlay-on-hdmi: write and read both happen in Overlay on HDMI (even screen RAM must round-trip) ---
  [selfcheck-overlay-on-hdmi] tolerating 6 isolated single-byte mismatch(es) as ordinary background CPU activity during a multi-second live write+read (addresses: $0007, $00AA, $2C39, $3942, $3DB2, $5760)
  [selfcheck-overlay-on-hdmi] ignored ROM mismatches (writes to real ROM are no-ops): 0
  [selfcheck-overlay-on-hdmi] screen ($0400-$07FF) mismatches: 0 (expected 0 in same-mode round trip)
  [selfcheck-overlay-on-hdmi] ignored known-live-noise mismatches: 0
  [selfcheck-overlay-on-hdmi] color RAM (low nibble) mismatches: 0
  [selfcheck-overlay-on-hdmi] RAM mismatches: 0
[16] close menu (Overlay on HDMI) ... OK
[17] set Interface Type = Overlay on HDMI ... OK
[18] open menu (Overlay on HDMI) ... OK
[19] write full-range pattern (Overlay on HDMI) ... OK
[20] capture ground truth in Overlay on HDMI mode ... OK
[21] close menu (Overlay on HDMI) ... OK
[22] set Interface Type = Freeze ... OK
[23] open menu (Freeze) ... OK
[24] read full range (Freeze) ... OK
[25] menu still open after cross-mode read (Freeze) ... OK
--- overlay-on-hdmi-write_freeze-read: bytes written while Overlay on HDMI must read back the same while Freeze, except screen RAM (menu-owned while frozen) ---
  [overlay-on-hdmi-write_freeze-read] screen ($0400-$07FF) mismatches: 996 (expected, menu owns this range while frozen)
  [overlay-on-hdmi-write_freeze-read] ignored known-live-noise mismatches: 0
  [overlay-on-hdmi-write_freeze-read] color RAM (low nibble) mismatches: 0
  [overlay-on-hdmi-write_freeze-read] RAM mismatches: 0
[26] close menu (Freeze) ... OK
[27] set Interface Type = Freeze ... OK
[28] open menu (Freeze) ... OK
[29] write full-range pattern (Freeze) ... OK
[30] capture ground truth in Freeze mode ... OK
[31] close menu (Freeze) ... OK
[32] set Interface Type = Overlay on HDMI ... OK
[33] open menu (Overlay on HDMI) ... OK
[34] read full range (Overlay on HDMI) ... OK
[35] menu still open after cross-mode read (Overlay on HDMI) ... OK
--- freeze-write_overlay-on-hdmi-read: bytes written while Freeze must read back the same while Overlay on HDMI, except screen RAM (menu-owned while frozen) ---
  [freeze-write_overlay-on-hdmi-read] screen ($0400-$07FF) mismatches: 1021 (expected, menu owns this range while frozen)
  [freeze-write_overlay-on-hdmi-read] ignored known-live-noise mismatches: 0
  [freeze-write_overlay-on-hdmi-read] color RAM (low nibble) mismatches: 0
  [freeze-write_overlay-on-hdmi-read] RAM mismatches: 0
[36] close menu (Overlay on HDMI) ... OK
restored Interface Type to 'Freeze'

=== SUMMARY ===
  selfcheck-freeze: PASS
  selfcheck-overlay: PASS
  overlay-to-freeze: PASS
  freeze-to-overlay: PASS
readmem_writemem_test: OK (36 checks)
```

Re-run three times consecutively against the same hardware with identical results (PASS across all four checks each time) to confirm repeatability.

#### Ultimate II+L

Command: `./readmem_writemem_test.py --host u2 --test all`

Output:

```
[01] read User Interface Settings config ... OK
  no 'Interface Type' setting on this device -- freeze-only hardware, restricting to freeze-mode tests
  skipping (requires Overlay-on-HDMI, unsupported here): selfcheck-overlay, overlay-to-freeze, freeze-to-overlay
[02] reset C64 to a clean, stable state ... OK
  skipping live-noise probe: Freeze halts the CPU entirely, no drift is possible
[03] open menu (Freeze) ... OK
[04] write full-range pattern (Freeze) ... OK
[05] read back full range (Freeze) ... OK
[06] menu still open after write+read (Freeze) ... OK
--- selfcheck-freeze: write and read both happen in Freeze (screen RAM excluded, menu redraws it continuously) ---
  [selfcheck-freeze] ignored ROM mismatches (writes to real ROM are no-ops): 16330
  [selfcheck-freeze] screen ($0400-$07FF) mismatches: 110 (expected, menu owns this range while frozen)
  [selfcheck-freeze] ignored known-live-noise mismatches: 0
  [selfcheck-freeze] color RAM (low nibble) mismatches: 0
  [selfcheck-freeze] RAM mismatches: 0
[07] close menu (Freeze) ... OK

=== SUMMARY ===
  selfcheck-freeze: PASS
readmem_writemem_test: OK (7 checks)
```


chrisgleissner:
Hi @GideonZ ,

thanks for your review and hopefully merge. Any changes needed, just let me know.

Best wishes
Christian

Issue #715, opened 2026-07-07
Long Filenames broken


GameTimeWarp:
Let me start out by saying this issue is is not in the last official 3.14d release.
I am seeing this in dev build 3.15 https://github.com/GideonZ/1541ultimate/actions/runs/28717417081
on an Ultimate64 Elite I

**To Reproduce**
Browse to a filename on a USB drive on the Ultimate64 that is at least 67 characters long not including the extension
Try to Mount Disk or Run Disk and get an error "Opening the disk file failed."
Try to rename the file and get an error "ERROR: FILE DOESN"T EXIST"

While long filenames can be seen in the Ultimate 64 file browser, you can't do ANYTHING with said files

An example file name that is giving this error is
Rambo First Blood Part II NTSC - Thunder Mountain 1985 (EasyLoad64).d64

If I remove (Easyload64) from the filename then it is short enough to then be mounted
Rambo First Blood Part II NTSC - Thunder Mountain 1985.d64

**Expected behavior**
To be able to mount filenames this long or longer like I was able to on official firmware 3.14d

**Applies to (please complete the following information):**
 - Ultimate64 Elite I
 - Git tag: v3.15
 - Git Hash: ee005041
 
**Additional context**
Reporting so this issue doesn't make it's way in to the official 3.15 release

chrisgleissner:
Hi @GameTimeWarp ,

Thanks for your very precise bug report. I was able to reproduce this very bug and provided a fix in https://github.com/GideonZ/1541ultimate/pull/718. 

You can find the latest build for that branch in https://github.com/GideonZ/1541ultimate/actions/runs/29213574163

Thanks for giving it a try and let me know if it works. If there are remaining issues to be fixed, please let me know.

Thanks
Christian

GameTimeWarp:
I can confirm that the issue is fixed on my Ultimate64 Elite I
Should I close the issue now or wait for Gideon to accept the merge request?
Thank you for fixing this issue!

chrisgleissner:
That is excellent news, thanks for the positive feedback!

Yes, please keep it open for now. We can close it once @GideonZ merges PR #718

chrisgleissner:
Good news @GameTimeWarp ,

Gideon kindly merged this PR to `test-merge` which is the branch on which the 3.15 release tags get created:
https://github.com/GideonZ/1541ultimate/actions/runs/29277714131

Thus, I suggest to close this issue now.

[...]

PR #718, opened 2026-07-12
Fix long browser filenames

chrisgleissner:
Fixes https://github.com/GideonZ/1541ultimate/issues/715.

## Overview
This PR fixes long filename handling in the Ultimate 64 browser when acting on FAT-backed entries.

User-visible failures before this change:
- `Mount Disk` and `Run Disk` could fail with `Opening disk file failed.`
- `Rename` could fail with `ERROR: FILE DOESN'T EXIST`.
- The browser could display long filenames correctly, but still fail to act on them.

## Bugs
1. `BrowsableDirEntry::getName()` converted browser action names through a fixed 64-byte buffer.
2. `FileInfo::generate_fat_name()` did not guarantee NUL termination for direct FAT names.
3. The shared filesystem path walker still matched entries through a fixed 64-byte scratch buffer, so long direct names could not be resolved reliably even after the browser handed them off.

## Fixes
- Return direct FAT names from `BrowsableDirEntry::getName()` without truncating them.
- Keep CBM-name conversion behavior unchanged, but size that conversion buffer from the actual entry length.
- Make `FileInfo::generate_fat_name()` always NUL-terminate direct-name output.
- Widen the common `FileSystem::walk_path()` FAT-name scratch buffer and pass its full size into `generate_fat_name()` so shared path resolution handles long names end to end.

- `make -C software/filemanager/tests temp-auto-cleanup`
- Added `GenerateFatNameKeepsLongDirectNames`
- Added `FileManagerHandlesLongDirectNames`

The following e2e test was performed against an Ultimate 64 Elite I. 

It successfully demonstrates rename and mounting of filenames exceeding 64 chars, specifically the name `zzzz_long_filename_browser_regression_0123456789_0123456789_0123456789_0123456789.d64`.

Invocation: `./tools/io/filemanager/browser_long_filename_test.py`

Output:
```
[01] seed long-name fixture for rename ... OK
[02] browser rename on long filename ... 
[03] open menu ... OK
OK
[04] reseed long-name fixture for mount ... OK
[05] browser mount on long filename ... 
[06] open menu ... OK
OK
browser_long_filename_test: OK
```


chrisgleissner:
Hi @GideonZ ,

as mentioned on #715 , this PR fixes the issue as shared by its reporter.

Thanks for review and merge.

Best,
Christian

Issue #717, opened 2026-07-09
System crashes completely when printing with Build 703 (v3.15) from Actions


xinteksik:
I'm testing version from Build 703 (https://github.com/GideonZ/1541ultimate/actions/runs/28717417081) and I've run into an issue with printing to virtual Printer. When the file is supposed to be saved, the screen turns off and the system crashes completely. A reset doesn't help—the only way to recover is to power it off and then back on. 

On Ultimate64 Elite 2

@GideonZ  I know this is a test build, but I wanted to report it just in case.

Thank you.
Tom

[Printer Settings]
IEC printer=Enabled
Bus ID=4
Output file=/Usb0/printer
Output type=PNG B&W
Ink density=Medium
Page top margin (default is 5)=1
Page height (default is 60)=66
Emulation=Epson FX-80/JX-80
Commodore charset=USA/UK
Epson charset=Basic
IBM table 2=International 1

xinteksik:
**To Reproduce**
Steps to reproduce the behavior:
1. GEOS with printer driver EX-800 v3.5 - print one page.
2. or Print Master - printing page longer than full page. Simple print more times. Test page printing from settings menu is ok.
3. or Pagefox 

**Applies to:**
 - Hardware type: U64E2
 - Firmware Version 3.15
 - Git Gash ee005041

chrisgleissner:
@xinteksik Thanks for your very detailed reproduction steps, and for giving 3.15 a go. If you have time, would you mind bisecting this issue? In other words, it would help to determine the precise Git commit ID that broke things, and the quickest way to achieve that would be via a Git bisect.

I appreciate that is a lot more effort, but it would help.

xinteksik:
Simple test to reproduce it:

10 open 1,4
20 for i = 1 to 100
30 print#1,"line no "; i
35 print#1, chr$(13)
40 next i
50 end

Then Flush fom menu.

chrisgleissner:
Thanks @xinteksik for the reproduction test. That is excellent. 

Two questions:
1. Did you also need to invoke "Flush" via the menu in the "print one page via GEOS" test approach you described earlier, or does the issue manifest itself itself without any flush.
2. Does your simple Basic test have the same effect as the one you noticed when printing a page via GEOS, i.e. complete system hang, black screen, and the need to power cycle your device in order to restore it?

I am investigating and will report back once I have findings.

chrisgleissner:
Hi @xinteksik ,

[...]

PR #719, opened 2026-07-13
Harden virtual printer page scanning and PNG allocation

chrisgleissner:
Fixes and hardens the virtual-printer paths implicated by #717.

## What this PR fixes

| Area | Failure mode | Fix |
| --- | --- | --- |
| Page-number scan | No-slash `Output file` names dereference NULL / build an invalid fallback path | Split only when a slash exists; otherwise scan `.` |
| Page-number scan | Rejected directory entries leak `FileInfo` objects | Release every scanned entry |
| Page-number scan | Directories and same-prefix junk names can be treated as pages | Accept only exact `<base>-NNN.png` files |
| PNG save path | `realloc()` reads FreeRTOS private heap metadata at the wrong offset and can copy out of bounds | Keep allocation sizes in an aligned wrapper-owned header |
| Allocation failure | Wrapped `malloc`/`calloc` panic forever; `realloc` can `memcpy` to NULL | Return NULL to C callers, check overflow/failure, and preserve the original `realloc` block |

## Root cause of the Flush/Eject crash

The confirmed first-bad commit is `76d887bdead10a57212e4d6731c04f94f53d3864`.
It moved U64 builds to FreeRTOS `heap_4` and added global allocator wrappers.
Its `__wrap_realloc()` assumes an 8-byte heap header and reads a size at
`ptr - 4`. On the RISC-V port, `heap_4` aligns its header to 16 bytes, making
that address application memory rather than allocator metadata.

The PNG encoder grows `ucvector` and `uivector` with `realloc()` while saving
a page. The bad size therefore causes an out-of-bounds copy during
Flush/Eject, which can corrupt firmware state and degrade the entire device.

The replacement wrapper does not inspect heap internals. It uses a private,
alignment-sized header only for C malloc-family allocations and implements
standard `realloc(NULL, n)`, `realloc(p, 0)`, overflow, and failure behavior.
Existing fatal C++ `new` behavior is intentionally unchanged.

## E2E coverage

`tools/io/printer/printer-e2e.py` now:

- tokenizes and runs the issue reporter's literal BASIC V2 reproducer;
- triggers the same menu **Flush/Eject** action;
- chooses output storage in order: `USB0`, `USB1`, `USB2`, then `Temp`;
- validates generated PNG structure over FTP.

## Hardware validation: Ultimate 64 Elite I

Testing was performed against the connected **Ultimate 64 Elite I** (`u64`)
using JTAG-deployed U64 firmware and the real IEC virtual-printer path.

| Test | Result |
| --- | --- |
| Literal issue #717 BASIC program → menu Flush/Eject | PASS: device remained REST/menu responsive; valid PNG written to the selected USB2 volume |
| Epson bitmap PNG | PASS |
| Epson text PNG | PASS |
| Commodore bitmap PNG | PASS |
| Commodore text PNG | PASS |
| Python harness compilation, 6510 assembly, `git diff --check` | PASS |

The U64/U64-II firmware build phases completed before this commit. The target
had USB2 mounted; automatic selection chose it after USB0 and USB1 were absent.

## Historical comparison status

The issue reporter already established the immediate-parent / first-bad
boundary (`2367b2ff` good, `76d887bd` bad). A local automated replay of that
comparison is not yet valid: the historical parent predates the REST menu and
synthetic-input APIs used by the harness, so attempting that automation tests
the missing legacy API rather than the printer Flush/Eject path. This PR does
not overstate that comparison as locally reproduced.

## Risk

Low. The firmware behavior change is confined to allocator wrappers introduced
by the first-bad commit. IEC protocol behavior, FileManager behavior, and the
existing #719 page-number semantics are otherwise unchanged.

### DIFF
--- a/software/io/printer/mps_printer.cc
+++ b/software/io/printer/mps_printer.cc
@@ -765,27 +765,28 @@ MpsPrinter::calcPageNum(void)
         for(int i=0;i<infos->get_elements();i++)
         {
             FileInfo *inf = (*infos)[i];
+            const char *suffix = inf->lfname + baselength;
 
-            if (!strncmp(basename, inf->lfname, baselength))
+            if (!(inf->attrib & AM_DIR) && !strncmp(basename, inf->lfname, baselength))
             {
                 /* Basename matches, then look if rest of filename is -XXX.png */
-                if (inf->lfname[baselength] != '-') goto next_file;
-                if (inf->lfname[baselength+1] < '0') goto next_file;
-                if (inf->lfname[baselength+1] > '9') goto next_file;
-                if (inf->lfname[baselength+2] < '0') goto next_file;
-                if (inf->lfname[baselength+2] > '9') goto next_file;
-                if (inf->lfname[baselength+3] < '0') goto next_file;
-                if (inf->lfname[baselength+3] > '9') goto next_file;
-                if (inf->lfname[baselength+4] != '.') goto next_file;
-                if (inf->lfname[baselength+5] != 'p') goto next_file;
-                if (inf->lfname[baselength+6] != 'n') goto next_file;
-                if (inf->lfname[baselength+7] != 'g') goto next_file;
-                if (inf->lfname[baselength+8] != '\0') goto next_file;
+                if (suffix[0] != '-') goto next_file;
+                if (suffix[1] < '0') goto next_file;
+                if (suffix[1] > '9') goto next_file;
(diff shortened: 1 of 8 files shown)


chrisgleissner:
Hi @GideonZ ,

even though this PR fixed other issues that can result in full device degradation, the problem reported in  #717 is not fully fixed as the author of that ticket shared. 

I want to investigate this a bit further.

Thanks
Christian

PR #721, opened 2026-07-14
FTP filesystem E2E test + fix control-connection netconn leak

chrisgleissner:
## Summary

Adds a real-device end-to-end test for the FTP remote filesystem and fixes a firmware bug uncovered by that test.

The bug was that removing a configured FTP host leaked its open control connection. Under repeated use, this exhausted the shared lwip `netconn` pool and wedged the entire network stack: REST, FTP, and telnet all stopped responding until reboot.

This targets `test-merge` because that is where the FTP filesystem feature lives.

## Bug and fix

`BrowsableFTPServer::S_remove()` unlinked the server node with `IndexedList::remove()`, but never deleted the node or closed its FTP control socket.

`IndexedList::remove()` only unlinks the item. It does not delete it.

As a result, every add -> enter -> remove cycle leaked one lwip **netconn**. With `MEMP_NUM_NETCONN = 16`, shared by the HTTP, FTP, and telnet servers, as well as the FTP client, roughly a dozen cycles were enough to exhaust the pool.

Once no free netconns remained, no server could `accept()`, so all inbound services went dark at once. This produced a permanent network wedge that required a reboot.

Pure REST churn did not trigger the issue. A run of 4000 sequential REST requests stayed healthy. The leak was specifically in the FTP client lifecycle, draining the same pool the servers also need. MicroHttpServer already rejects requests correctly when full, so no server-side change was needed.

The fix in `software/filesystem/filesystem_ftp.cc` is intentionally minimal: drop the control connection, then free the node and its filesystem when the host is removed.

```cpp
serv->drop_connection();          // close control socket -> free the netconn
ftp_root->children.remove(serv);
FileSystemFTP *fs = (FileSystemFTP *)serv->get_file_info()->fs;
serv->get_file_info()->fs = NULL;
delete fs;
delete serv;
```

`drop_connection()` is null-safe and `disconnect()` is idempotent, so the later destructor remains safe.

`S_edit()` already dropped the connection, so it did not have the same leak.

## End-to-end test

Adds `tools/io/filemanager/ftp_client_test.py`.

The test drives the **real firmware UI** using only the supported REST APIs:

- `machine:menu_button`
- `machine:menu_screen`
- `machine:input`

It stands up a controlled `pyftpdlib` server with deterministic fixtures and a full command log. The test then configures the server as a remote host through the New Host form, browses it under `/ftp`, and verifies every FTP operation against the server's real filesystem and command trace.

Supported stages are:

`smoke`, `core`, `edge`, `matrix`, `negative`, `soak`, `all`

Coverage:

- Host lifecycle: create, list, edit, remove.
- Browsing: nested directories, paths with spaces, scrolling.
- File retrieval: `RETR` across multiple file sizes, with byte-count verification.
- Device-side behaviour: cache-hit behaviour.
- Real menu actions: `MKD`, `STOR`, `RNFR` + `RNTO`, `DELE`, `RMD`.
- Multiple simultaneous hosts: password-protected and anonymous.
- Host variations: alias, path, and port variations.
- Context menus: per-file-type context menu coverage.
- Negative cases: bad password, unreachable port, missing path, permission denied, server stop/restart.
- Soak testing: repeated soak loop.

The test never uses firmware-internal hooks, always cleans up after itself, and restricts cleanup to its own `E2EFTP...` hosts.

## Test housekeeping

Renamed the other `tools/` Python tests to follow the `*_test.py` underscore convention:

- `printer-e2e.py` -> `printer_test.py`
- `temp-auto-cleanup-perf-test.py` -> `temp_auto_cleanup_perf_test.py`

Updated all references accordingly.

## How it was tested

Built and JTAG-deployed to an Ultimate 64 Elite running firmware 3.15. The controlled FTP server ran locally on the test host.

Regression proof of the fix: 25 create/enter/remove cycles, well past the 16-netconn pool size.

```text
PASS: survived 25 cycles, device still responsive, resets=0
```

Before the fix, the device wedged around cycle ~12 and required JTAG redeploy.

All stages passed on the U64 running firmware 3.15:

| Stage | Result |
| --- | --- |
| `smoke` | Host create/list/enter/RETR/remove |
| `core` | Browse nested directories, paths with spaces, and 30-entry scroll; `RETR` of 0/16/4096/4097/16384/1024 bytes with byte verification; cache-hit; `MKD`; `STOR` with byte-match; `RNFR`/`RNTO`; `DELE`; `RMD` - all OK |
| `edge` | 24 checks OK: two hosts at once, passworded + anonymous; password vs anonymous login; alias edge cases; subfolder root isolation; alternate port; per-file-type context menus for TXT/BIN/PRG; View + Hex View; hostname connect |
| `matrix` | Root `/` and `/DIRA` create + LIST + RETR - OK |
| `negative` | Bad password, unreachable port, missing path, permission-denied delete `550`, server stop/restart - each recovers, device remains responsive, no stuck keys |
| `soak 10m` | 84 host-enter/browse/read cycles, 2549 FTP commands, device responsive throughout - no wedge |

Interaction speed was maximised while keeping REST access single-threaded and sequential only:

- Typing runs with no backoff at approximately 50 keys/s.
- Character input remained character-perfect.
- Batching is avoided because it overflows the device key buffer.
- Navigation runs at approximately 17 taps/s.


chrisgleissner:
Hi @GideonZ ,

here is the E2E we discussed.  

Other changes:
- Fixed a bug related to repeated removal of FTP servers which could degrade the device.
- Ensure Python test filename consistency to always use underscores and a `test.py` suffix. Some earlier E2E tests had used dashes instead; fixed now.

Thanks for your review and merge.

Best,
Christian

Issue #729, opened 2026-07-23
Two options broken when browsing inside a d64 or loading a PRG file outside a d64


GameTimeWarp:
**Describe the bug**
"DMA" option is broken  when browsing inside a d64 and pressing RETURN on any PRG file or when pressing ENTER on a PRG file outside of a d64.

"REAL RUN" option is broken then browsing inside a d64 and pressing RETURN on any PRG file and gives a FILE NOT FOUND ERROR

Has the DMA menu option been deprecated? Should it be removed from the sub menu options?

**To Reproduce DMA issue**
1. Browse inside a d64 image file using right cursor
2. Press return on any PRG file inside OR outside the d64
3. In the next menu, select the DMA option and press RETUIRN
4. You will be dropped back to basic and nothing will happen

**Expected behavior**
I expected the option to do a DMA load of the chosen PRG file

**To Reproduce REAL RUN issue**
1. Browse inside a d64 image file using right cursor
2. Press return on any PRG file inside the d64
3. In the next menu, select the REAL RUN option and press RETUIRN

Machine resets then does a LOAD and gets a file not found error like shown below
LOAD"GAMENAME",8,1
?FILE NOT FOUND ERROR
READY,
RUN

READY.

**Expected behavior**
In firmware 3.14d this option was working properly and did a real load and run on the selected file

**DMA Issue replicated on the following:**
 - Ultimate Elite I
 - Firmware 3.14d and 3.15
 - 40a41caa
 
**REAL RUN Issue replicated on the following:**
 - Ultimate Elite I
 - Firmware 3.15
 - 40a41caa

chrisgleissner:
Thanks for your report. Excellent as usual.  This looks like a regression to me and I will investigate.

chrisgleissner:
@GameTimeWarp I could not reproduce any issues with the DMA and REAL RUN behaviour on the latest 3.15. 

However, I kept digging into the code and found a bug that may affect long filenames. This is fixed in https://github.com/GideonZ/1541ultimate/actions/runs/30140091261 

For further context, this is what I saw on the latest 3.15 (built from the `test-merge` branch) without my fixes:
- DMA (PRG within or outside of D64) loaded correctly. I could then issue a `RUN` command to launch whatever I had DMA-loaded into RAM.
- REAL RUN (PRG within or outside of D64) loaded and started program.

[...]

PR #731, opened 2026-07-25
Fix unbounded boot-cart name copy

chrisgleissner:
## Overview

Removes an unbounded string copy in the boot-cart loader that can crash the app. 

In order to increase E2E test coverage and as part of investigating (#729), this PR also adds an end-to-end test that drives **every** action the context menu offers for a PRG and verifies each one from outside the browser, plus a top-level script that runs all hardware end-to-end suites against one device.

## Issue

`format_bootcrt_display_name()` shortens long paths to `...` plus the last 13 characters before the boot cart displays them. The pass-through branch copied the name into a 17 byte stack buffer with no bound:

```cpp
char display_name[17];
...
strcpy(display_name, name);   // name can be a full FAT long file name
```

Any PRG whose file name is longer than 16 characters and has no path separator - an ordinary long file name - overruns that buffer inside `dma_load()`.

On an Ultimate 64 Elite, selecting **Run** on a PRG with a 101 character name **kills the firmware outright**: no REST, no FTP, no ping, recoverable only by a power cycle or a JTAG download. It is a regression introduced after `v3.14d` and present in `v3.15`.

## Fix

The helper now returns the caller's name unchanged for the pass-through case, and only writes into the caller's buffer for the trimmed case. The trimming itself is untouched.

Passing the name on untouched is also what `CbmFileName` wants: it cuts to 16 PETSCII characters itself, and it needs the file extension that pre-trimming would remove.

## Context

Same class of bug as #718 (*Fix long browser filenames*), which tightened `generate_fat_name()` and grew the FAT name buffer in `FileSystem::walk_path()` from 64 to 132 bytes: a fixed-size name buffer meeting a long file name. This is the same pattern one layer further on, where the browser hands that name to the boot-cart loader. The new end-to-end test also follows the one added there, `tools/io/filemanager/browser_long_filename_test.py`.

## Evidence

On an Ultimate 64 Elite I, same test both sides, only the firmware swapped. The deployed ELF was checked for the fix before each run so a stale image cannot be mistaken for a result.

**The bug, 10 attempts per side.** Every failing attempt takes the device down, so each attempt gets a fresh JTAG deploy first:

```
prg_context_menu_test.py -H u64 --repeat 1 --scenario long-name

before:  [01/1] FAIL Run a PRG with a 101 character name -> device stopped responding
         === PRE-FIX: 0 ok, 10 failed out of 10

after:   [01/1] ok   Run a PRG with a 101 character name
         === FIXED: 10 ok, 0 failed out of 10
```

Fully deterministic in both directions.

**Nothing else moved.** The full context-menu matrix on the fixed build:

```
     offers: Run, Load, DMA, View, Hex View, Copy to..., Move to..., Rename, Delete
     offers: Run, Load, DMA, Mount & Run, Real Run, View, Hex View, Copy to..., Move to..., Rename, Delete
prg_context_menu_test: OK (23 actions)
```

## On #729

Neither of the two symptoms in that report survived investigation as a bug of its own:

* **DMA** behaves identically before and after this change: the program lands in RAM, BASIC's pointers are set, and `RUN` starts it - 10/10 on both firmwares. Leaving the Ultimate menu afterwards is the intended behaviour.
* **Real Run** never failed for me. 10 consecutive runs on each firmware, plus 17 CBM name shapes on the RAM disk against both firmwares (spaces, `/`, `.PRG`, `+`, `*`, brackets, 16 character names, 8 and 16 shifted letters, 16 PETSCII graphics characters, mixed) and 6 shapes from a folder on a USB stick. All of them loaded and ran.

The overrun above is the only change on that code path between `v3.14d` and `v3.15`, and how much damage it does depends on the stack layout of the build, so it stays a plausible cause on another unit - but I am not claiming it proven. @GameTimeWarp, if Real Run still fails for you on a build with this fix, the disk image that does it is what I need.

## The new test: `tools/io/filemanager/prg_context_menu_test.py`

Follows the existing `tools/io/**/..._test.py` conventions: self-contained, `RestSession` + `check()`/`Failure` from `menu_screen_test`, `-H/-p/-t` with environment fallbacks.

It uploads a small BASIC + machine-code program to `/Temp` three times - as a plain file, under a 101 character name, and as the only file in a D64 - then reads what the context menu offers for a PRG in each location and drives all of it. A new menu entry that no check covers fails the run.

Nothing is taken on the browser's word: `machine:readmem` for the signature the program writes at `$C000`, the load image at `$0801` and the text screen; FTP for the plain file; and the D64 pulled back over FTP and its directory decoded for the disk entry.

| Action | Checked |
| --- | --- |
| `Run` (plain, in-D64, long name) | signature written and the program's output on screen |
| `Load`, `DMA` (plain, in-D64) | program in RAM, and for DMA that `RUN` then starts it |
| `Mount & Run`, `Real Run` | D64 on drive A, no drive error, program ran |
| `View`, `Hex View` | viewer opens over the browser, hex dump shows the fixture's bytes |
| `Copy to...`, `Move to...` | file lands in the picked directory, origin left alone or emptied |
| `Rename`, `Delete` | the directory entry really changed |

`--repeat N [--scenario NAME]` re-runs the load/run scenarios N times, which is how the numbers above were produced. The test pins the **Freeze** interface for the run and restores the previous setting, so a result does not depend on how the device happened to be configured.

`./run-e2e-tests [-H host] [-p password]` runs all hardware suites sequentially and prints a summary; `-l` lists them, `-s` picks one, `-x` stops at the first failure.

## Tested hardware

- Ultimate 64 Elite I, firmware built from this branch and deployed over JTAG

`u64` and `u64ii` build clean. The local `u2` software-only build fails on a missing FPGA bitstream checksum (`rv700dd.chk`); that failure is identical on an unmodified tree here, since this machine has no Xilinx ISE.

### DIFF
--- a/run-e2e-tests
+++ b/run-e2e-tests
@@ -1,0 +1,180 @@
+#!/usr/bin/env bash
+# run-e2e-tests — run every hardware end-to-end test against one Ultimate device.
+#
+# Each suite drives the real firmware over REST/FTP/Telnet, so a device has to
+# be reachable and idle. Suites run sequentially; the exit status is non-zero if
+# any of them fails.
+
+set -uo pipefail
+
+SCRIPT_NAME=$(basename "$0")
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+
+HOST="${U64_HOST:-u64}"
+PASSWORD="${U64_PASS:-}"
+TIMEOUT="30.0"
+SELECTED=()
+LIST_ONLY=false
+STOP_ON_FAIL=false
+
+C_BLUE='\033[1;34m'
+C_GREEN='\033[1;32m'
+C_RED='\033[1;31m'
+C_YEL='\033[1;33m'
+C_NC='\033[0m'
+
+# name|relative path|space separated arguments (@HOST@/@PASS@/@TIMEOUT@ are substituted)
+SUITES=(
+  "readmem-writemem|tools/api/readmem_writemem_test.py|-H @HOST@ -p @PASS@ -t @TIMEOUT@ --test all"
+  "menu-screen|tools/api/menu_screen_test.py|-H @HOST@ -p @PASS@ -t @TIMEOUT@ --test all"
+  "input|tools/api/input_test.py|-H @HOST@ -p @PASS@ -t @TIMEOUT@"
+  "prg-context-menu|tools/io/filemanager/prg_context_menu_test.py|-H @HOST@ -p @PASS@ -t @TIMEOUT@"
+  "browser-long-filename|tools/io/filemanager/browser_long_filename_test.py|-H @HOST@ -p @PASS@ -t @TIMEOUT@"
+  "ftp-client|tools/io/filemanager/ftp_client_test.py|-H @HOST@ -p @PASS@ -t @TIMEOUT@"
+  "prg-load-path-trim|tools/io/prg-load-path-trim-test.sh|-H @HOST@ -p @PASS@"
+  "temp-auto-cleanup|tools/io/temp-auto-cleanup/temp-auto-cleanup-test.sh|-H @HOST@ -p @PASS@"
+  "printer|tools/io/printer/printer_test.py|-H @HOST@ -p @PASS@"
+)
+
+show_help() {
+    cat << EOF
+Usage: $SCRIPT_NAME [OPTIONS] [HOST]
+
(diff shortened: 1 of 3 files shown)


chrisgleissner:
Hi @GideonZ,

I found another instance of the "long filename" bug we encountered and fixed before. This could crash the app, so would be good to merge.

Also, this PR adds a new E2E test and a wrapper across all already existing E2E tests so we can run them more conveniently.

Thanks for your review,
Christian

PR #732, opened 2026-07-25
Spike: Audio streaming over WiFi

chrisgleissner:
## Overview

I wanted to explore whether the existing audio stream might be sent over WiFi without changing the FPGA. 

Based on the firmware interfaces I could find, I have not yet found a practical way to do that.

I am opening this as a draft mainly to share the findings and check whether there is an existing FPGA-to-CPU path that I may have missed.

## Motivation

Streaming currently only supports Ethernet, not WiFi. One of the few workarounds is to use this setup: `U64 -(Ethernet)-> Router -(WiFi) -> Client`. This requires an Ethernet cable to a WiFI-enabled router. Not all users have such a device.

The U64 WiFi bandwidth is considerably too low for video stream, but would be sufficient for an audio-only stream.

Offering audio streaming over WiFi would allow more users to enjoy listening to SID songs wirelessly which is important for portable clients (smartphones, tablets, etc).

## What the spike does

`PUT /v1/streams/{stream}:start` accepts an optional `wifi` parameter:

```text
PUT /v1/streams/audio:start?ip=192.168.1.185:11001
PUT /v1/streams/audio:start?ip=192.168.1.185:11001&wifi=true
```

The default Ethernet behaviour is unchanged.

With `wifi=true`, the FPGA stream is redirected back to the U64, received by a relay task, and then forwarded over WiFi.

## Why this is not viable

From what I can see, the FPGA generates the stream packets directly on the Ethernet transmit path. The CPU neither produces nor receives their payloads.

The only software-only route I found was:

```text
FPGA -> Ethernet -> router/AP -> WiFi -> ESP32 -> CPU
     -> ESP32 -> WiFi -> destination
```

This route is clearly impractical:

* It still requires an Ethernet cable.
* Packets leave the device only to return to it.
* The payload crosses the ESP32 UART twice.
* It depends on Ethernet and WiFi being bridged by the external network.

The relay demonstrates that the REST and WiFi forwarding parts work, but it is not a reasonable implementation to build on.

## Main question

Is there already a mode in the FPGA stream generator that exposes generated packets to the CPU instead of sending them directly to the Ethernet MAC?

I could not find such a path in the code available here, but the relevant top-level FPGA implementation is not included, so I may have missed an existing capability.

Without a CPU-visible packet or audio-sample path, true WiFi-only streaming would appear to require an FPGA change.

## Separate issue found while testing

I also found that unicast streaming can fail when Ethernet and WiFi are active on the same subnet.

The UDP probe may be routed through WiFi, placing the ARP entry in the WiFi netif, while `startStream()` checks only the wired netif and reports `Network Host Resolve Error`.

The patch currently falls back to checking the other ARP tables. I am not certain that this belongs in this spike or that it is the best general solution, so it can be split out or removed.

Multicast and broadcast destinations are unaffected.

## Test

`tools/api/audio_stream_test.py` captures streams over both transports and checks for packet loss while tolerating duplicates and reordering.

```text
[05] ethernet: 2500 packets from 192.168.1.15 (249.9/s, sizes=[770],
     expected=2500, missing=0, duplicates=0, reordered=0) ... OK
[08] wifi:     2504 packets from 192.168.1.71 (250.2/s, sizes=[770],
     expected=2501, missing=0, duplicates=3, reordered=0) ... OK
[09] wifi stream uses a different interface than the ethernet stream ... OK
audio_stream_test: OK (9 checks)
```

No packet loss was observed during the test.

## Build status

* `u64`: built successfully
* `u64ii`: built successfully
* `u2`: compiles and links, but I could not run the updater step because it requires a prebuilt bitstream

### DIFF
--- a/software/api/route_streams.cc
+++ b/software/api/route_streams.cc
@@ -14,7 +14,20 @@ public:
 };
 static StreamNameToInt streamDict;
 
-API_CALL(PUT, streams, start, NULL, ARRAY ( { { "ip", P_REQUIRED } }))
+// Query parameters are strings; a flag is set by naming it, or by giving it a truthy value.
+static bool arg_is_true(const char *value)
+{
+    if (!value) {
+        return false;
+    }
+    if (!*value) {
+        return true;
+    }
+    return (strcasecmp(value, "true") == 0) || (strcasecmp(value, "1") == 0) ||
+           (strcasecmp(value, "yes") == 0) || (strcasecmp(value, "on") == 0);
+}
+
+API_CALL(PUT, streams, start, NULL, ARRAY ( { { "ip", P_REQUIRED }, { "wifi", P_OPTIONAL } }))
 {
     const char *streamName = args.get_path(0);
     SubsysCommand *sys_command;
@@ -38,7 +51,8 @@ API_CALL(PUT, streams, start, NULL, ARRAY ( { { "ip", P_REQUIRED } }))
     }
 
     sys_command = new SubsysCommand(NULL, -1, (int)dataStreamer, streamIndex, args["ip"], "");
-    sys_command->direct_call = DataStreamer :: S_startStream;
+    sys_command->direct_call = arg_is_true(args.get_or("wifi", NULL)) ? DataStreamer :: S_startStreamWireless
+                                                                     : DataStreamer :: S_startStream;
     SubsysResultCode_t retval = sys_command->execute();
     resp->error(SubsysCommand::error_string(retval.status));
     resp->json_response(SubsysCommand::http_response_map(retval.status));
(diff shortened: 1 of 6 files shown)


chrisgleissner:
Hi @GideonZ,

I looked into the WiFi audio streaming idea we discussed yesterday.

From what I could tell, audio bandwidth may be manageable over WiFi, unlike video. However, the firmware-only approach I tried does not yet seem practical, as it still requires Ethernet and routes the stream out through Ethernet before relaying it back over WiFi.

I opened the draft mainly to share what I found and to ask whether there may already be an FPGA-to-CPU path for the generated stream that I have missed. Some of the supporting work, particularly the tests, may still be useful later.

Thanks,
Christian

chrisgleissner:
Closing this spike to keep list of open PRs concise. Can revisit later.

PR #734, opened 2026-07-26
Fix menu wedge when SID auto address mirroring is off

### LINKED ISSUE #733: Disabling "Auto address mirroring" semi-bricks device when interface type is set to FREEZE
**Describe the bug**
Two settings seem to have to be configured a certain way to trigger the semi-brick
Interface type must be set to "Freeze"
Auto address mirroring must be set to "Disabled"

With the settings configured this way, when you press the quick action button or F10 on a wireless keyboard you will get a black screen and be stuck on that screen with your only option being to force power off the device

Seems to affect 3.15 builds
Does not effect 3.14d

**To Reproduce**
1. Press F2 to go to the settings menu
2. Select "SID Addressing"  -> "Auto address mirroring" and set it to "Disabled"
3. Select "User interface settings"  -> "Interface type" and set it to "Freeze"
4. Press F5 -> Configuration -> Save to Flash
5. Press F5 -> C64 Machine -> Power OFF
6. Power machine back on
7. Press the quick action button to bring up the U64 menu
8. You should then be stuck on a solid black screen where you can't do anything
9. Hold down the quick action button for 5+ seconds or until the U64 powers off
10. Hold down the restore key (and keep holding it) then press the quick action button to power on to the initial setup screen
11. Revert changes and save to flash -> Power OFF -> Power ON -> You can now enter the menus again!

**Expected behavior**
I normally disable "Auto Address mirroring" and set my ARMSID in Slot1 to $D400 without issue

**Screenshots**

 - Ultimate Elite I
 - 3.15
 - Git Gash: 26f408ce

**Additional context**
3.14d does NOT have this issue

Enabling "Auto address mirroring" alone is enough to get the device working again
Setting interface type to "Overlay over HDMI" alone is also enough to get the device working again

If "User interface settings"  -> "Interface type" is set to "Overlay over HDMI" then system works as it should even when Disabling "Auto address mirroring"

chrisgleissner:
Fixes #733.

## Symptom

`Interface Type` = `Freeze` plus `Auto Address Mirroring` = `Disabled` wedges the U64 on menu entry: black screen, dead menu button, forced power off to recover. Reported on 3.15, absent on 3.14d.

## Root cause

The freezer mutes the SIDs at three fixed addresses:

```c
#define SID_VOLUME   *(... + 0xD418)   // backup_io()  -> 0
#define SID2_VOLUME  *(... + 0xD438)   // restore_io() -> 15
#define SID3_VOLUME  *(... + 0xD518)
```

That is the layout of a real C64, where the SID decode is mirrored across `$D400-$D7FF` by the machine. On the U64 the decode is programmable: `get_sid_addresses()` produces `mask = 0xFE`, one 32-byte window, and only `auto_mirror()` widens it to `mask = 0xC0`.

| Auto Address Mirroring | Decoded by a SID | `$D418` | `$D438` | `$D518` |
|---|---|---|---|---|
| Enabled  | `$D400-$D7FF` | yes | yes | yes |
| Disabled | `$D400-$D41F` | yes | **no** | **no** |

So with mirroring off the freezer writes to I/O addresses that nothing claims, and the Nios stalls on an access that never completes. Measured on hardware: the device stops answering REST **and ICMP**, and `nios2-download` cannot even pause the CPU — the whole processor is stalled, not just the UI task.

There are exactly two stall sites, and both were reproduced:

| Site | When | Result |
|---|---|---|
| `backup_io()` | after `VIC_CTRL = 0`, before `init_io()` re-enables the screen | black screen, dead menu button |
| `restore_io()` | after the VIC/CIA restore, before `resume()` | C64 left stopped **in Ultimax mode**, where the VIC fetches characters from the cartridge window: the "garbled characters" from the issue thread |

This also explains the rest of the report. Freeze only, because `backup_io()`/`restore_io()` run only for the C64 host and never for the HDMI overlay. And 3.14d unaffected, because this firmware path is byte-for-byte identical between v3.14d and v3.15 — only the FPGA core changed in between. The firmware bug is older; the new core made it fatal.

Separately, the old code is functionally wrong with mirroring off even where it does not wedge: it mutes `$D418` regardless of where the SIDs are, so a SID at, say, `$D500` keeps playing while the menu is open.

## Fix

Resolve the SID bases from the live decoder configuration instead of hard-coding them.

- **`u64_config.cc/.h`** — new `u64_get_mapped_sid_bases()` returns the base address of every SID the FPGA currently decodes, from the existing shadow registers. It skips disabled sockets, unmapped decoders (`base` bits the `mask` filters out, which is how `UNMAPPED_BASE` works), masks that would not decode register `$18`, and anything outside the `$D400-$D7FF` / `$DE00-$DFFF` windows. Split configurations are expanded so every half gets muted.
- **`c64.cc`** — new `write_to_all_sids()` helper uses it. The symbol is referenced **weakly**, like `u64_dispatch_usb_hid_status_refresh()` in `userinterface.cc`, because the U64 recovery and updater targets compile `c64.cc` but not `u64_config.cc`. When it is absent, and on every non-U64 target, the helper falls back to the historical `$D400 / $D420 / $D500` trio — U2/U2+ behaviour is unchanged.
- **`c64.h`** — the six `SIDn_VOLUME` / `SIDn_DUMMY` macros are replaced by the two register offsets they were really expressing, so a fixed SID address cannot creep back in. They had no other user in the tree.

`U64Config` also pokes `$D400` / `$D500` during SID detection, but every one of those paths is bracketed by `S_SetupDetectionAddresses()`, which maps the sockets there first. Safe, and untouched.

## Testing

New end-to-end test `tests/e2e/io/c64/freeze_menu_test.py`, in the same shape and CLI as the other `tests/e2e/io` device tests. Three scenarios: mirroring off with a menu open/close, mirroring switched off while the menu is open, and mirroring on as a regression guard. Three oracles: `machine:menu_screen` proves the menu rendered and closed, `version` catches the wedge, and the KERNAL jiffy clock at `$A2` via `machine:readmem` proves the C64 was frozen while the menu was open and resumed cleanly afterwards.

Two independent red/green rounds on an Ultimate 64 Elite, firmware 3.15, FPGA 122, core 1.4B, deployed over JTAG.

**Red** — round 2 ran against the stock flashed 3.15, so this is not just a property of my rebuild:

```
[03] open the menu ... FAIL          # scenario 1, backup_io() stall
freeze_menu_test: FAIL: open the menu: the device stopped answering REST requests

[06] close the menu ... FAIL         # scenario 2, restore_io() stall
freeze_menu_test: FAIL: close the menu: the device stopped answering REST requests
```

Recovery each time needed a volatile `quartus_pgm` reconfig of `external/u64.sof`.

**Green** — same tree with the fix:

| Round | Result |
|---|---|
| 1 | 19/19, and 57/57 with `--repeat 3` |
| 2 | 95/95 with `--repeat 5` |

The device was left with its original configuration restored.

## Note on the FPGA side

The firmware should not be writing to unclaimed I/O addresses, which is what this PR fixes. Whether the core should also survive such an access rather than stalling the bus forever is your call — as it stands, the same pattern would take the device down again from anywhere else that pokes a fixed I/O address.


chrisgleissner:
Hi @GideonZ ,

thanks for review and hopefully merge.

Best wishes
Christian

chrisgleissner:
As discussed with @GideonZ , the best way to resolve this issue is at its root, in VHDL. 

This ensures that any attempt to interact with unmapped SID addresses (e.g. via a BASIC program) will work. The fix in this PR only covers the concrete menu interaction bug reported by the user. 

This PR will be closed as soon as the VHDL-side fix is done. The E2E of this test has already been merged to `test-merge`.

chrisgleissner:
Closing since Gideon has now fixed this on the Vhdl side.

PR #736, opened 2026-07-26
Debounce held USB keys when leaving the menu

chrisgleissner:
## The bug

Play a SID tune, open the menu, then press ESC on a USB keyboard to leave the menu: the SID player reappears for a fraction of a second and the menu immediately re-opens. The only way out is to tap ESC so briefly that it is already released by the time the menu closes.

It does not happen with the C64's own RUN/STOP key, and not via the `machine:menu_button` REST endpoint. Normal menu navigation with the same USB keyboard, including leaving the menu to a plain BASIC screen, is also fine - so it looks specific to the menu -> SID player transition.

## Root cause

USB ESC decodes into two things at once:

* `KEY_ESCAPE` in the character queue, which the menu consumes and turns into `MENU_HIDE`
* C64 matrix row 7 / bit 7 = **RUN/STOP** (`keymap_usb2matrix[0x29] = 0x3F`, `keyboard_usb.cc`)

While the menu is open `enableMatrix(false)` keeps those bits off the C64, but `matrix_state` still tracks the live report. `ultimate.cc` calls `enableMatrix(true)` as soon as `run_once()` returns, so a key that is *still physically held* is asserted on the C64 immediately.

The SID player scans the matrix itself and treats RUN/STOP press-then-release as "go back to the Ultimate menu" (`software/6502/sidcrt/player/advanced/keyboard.asm`: `cmp #$7f` sets `runStopPressed`, and `gotoUltimateMenu` fires a UCI Freeze command once all keys are released). So it saw a fresh RUN/STOP and re-opened the menu. Other programs simply ignore RUN/STOP, which is why only the SID player shows the symptom.

The mechanism that is supposed to prevent this already exists: `wait_free()`, called from `TreeBrowser::poll()` whenever `handle_key()` returns a negative code. But there was an asymmetry:

* `Keyboard_C64::getch()` falls back to `system_usb_keyboard.getch()` - the menu can be driven from the USB keyboard
* `Keyboard_C64::wait_free()` only polled the **hardware** row register

So a held USB key was never waited out. The C64's own RUN/STOP debounced correctly, REST had no physical key to debounce, and USB ESC fell straight through.

## The fix

`Keyboard_C64::wait_free()` now also waits for the USB report to go idle, mirroring the fallback that `getch()` already has. The change is purely additive - a self-contained block prepended, with the original body untouched apart from replacing the `2000` literal with the shared `KEYBOARD_WAIT_FREE_TIMEOUT_MS` constant. Each keyboard keeps its own full budget, so the added wait cannot eat the matrix scan's.

Supporting changes in `Keyboard_USB`:

* new `anyKeyPressed()` - true while the last report still holds any key or modifier
* `wait_free()` refactored onto it and given the same bounded timeout (it was an unbounded spin loop; it has no live caller today, but the loop was a latent hang)
* the poll interval now converts ms to ticks explicitly via `pdMS_TO_TICKS`. `configTICK_RATE_HZ` is 200, so the old `vTaskDelay(10)` was a 50 ms poll, not 10 ms.

## Why this is safe

The hardware wait polls a register that updates on its own, but the new USB wait depends on the USB task clearing `last_data`. If the UI task could starve it, this would deadlock. Verified it cannot:

* `configUSE_PREEMPTION = 1`, `configUSE_TIME_SLICING = 1`
* the UI loop is `"U-II Main"` at `PRIO_MAIN = 0`; the USB tasks are `PRIO_POLL` / `PRIO_DRIVER = 1`

The higher-priority USB tasks preempt the busy-wait, so the release is always observed.

Other paths checked:

* **REST is unaffected.** `anyKeyPressed()` reads only `last_data`, written solely by `process_data()` from the USB HID task. REST keys go to `injected_buffer`, so REST-driven automation cannot stall.
* **No USB keyboard attached -> zero change.** `last_data` stays zero and the loop never iterates.
* **Disconnect with a key held** clears the report via `usb_hid_remove_keyboard_source()`, so there is no permanent stall.
* All seven `wait_free()` call sites (menu close, context menu, popups, edit boxes, machine monitor, SID editor) are one-shot screen transitions, never per-keystroke - arrow keys return 0 from `handle_key()` and never reach it. Every one of them already behaved this way for the C64 keyboard; the USB keyboard now matches.

Held longer than the 2 s timeout the key still gets through, which is the same bound the C64 keyboard has always had.

## Testing

**Host unit tests** - `software/api/tests` (that suite already links `keyboard_c64.cc` + `keyboard_usb.cc` and stubs `wait_ms`, so the busy-wait can be driven deterministically):

* `HeldUsbEscapeIsRunStopOnTheC64Matrix` - documents why the wait is needed
* `MenuExitWaitsForHeldUsbKeyToBeReleased` - the regression test: asserts the wait happened and no phantom RUN/STOP reaches the matrix
* `MenuExitWaitsForBothKeyboards` - fake `GenericHost` + fake row register; proves the added USB wait does not shorten the existing matrix scan
* `MenuExitWaitIsBoundedWhenTheKeysAreNeverReleased` - worst case is exactly `2 x KEYBOARD_WAIT_FREE_TIMEOUT_MS`

Plus `anyKeyPressed` / bounded-`wait_free` tests in `software/io/usb/tests/usb_keyboard_queue_test.cpp`.

Red/green proven: with only `keyboard_c64.cc` reverted, 2 of 27 fail; with the fix, 27/27 pass.

Full host suite green: 55 USB, 27 input-api-state, 18 filemanager, machine monitor + bookmarks.

**Hardware** - built and JTAG-deployed to an **Ultimate 64 Elite I**, tested with a wireless **Keychron** USB keyboard. Confirmed the SID player no longer bounces back into the menu: holding ESC now closes the menu on release and it stays closed.

`u64` and `u64ii` build clean. The `u2` target fails on a missing `rv700dd.chk` rule in my environment, which reproduces unmodified on the base commit and is unrelated to this change.

### DIFF
--- a/software/api/tests/input_api_state_test.cpp
+++ b/software/api/tests/input_api_state_test.cpp
@@ -7,8 +7,29 @@
 
 #include <initializer_list>
 
+// Drives Keyboard_C64::wait_free(), which busy-waits through wait_ms(). Letting
+// the stub count calls and let go of the keys makes the wait deterministic.
+struct WaitFreeProbe {
+    int calls;
+    int release_usb_at;          // 0 = never
+    volatile uint8_t *row_register;
+    int release_row_at;          // 0 = never
+};
+
+static WaitFreeProbe wait_free_probe = { 0, 0, 0, 0 };
+
 extern "C" void wait_ms(int)
 {
+    wait_free_probe.calls++;
+    if (wait_free_probe.release_usb_at && (wait_free_probe.calls >= wait_free_probe.release_usb_at)) {
+        uint8_t release[USB_DATA_SIZE] = { 0 };
+        wait_free_probe.release_usb_at = 0;
+        system_usb_keyboard.process_data(release); // the user lets go of the USB key
+    }
+    if (wait_free_probe.release_row_at && (wait_free_probe.calls >= wait_free_probe.release_row_at)) {
+        wait_free_probe.release_row_at = 0;
+        *wait_free_probe.row_register = 0xFF;      // ... and of the C64 key
+    }
 }
 
 namespace {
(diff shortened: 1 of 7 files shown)


chrisgleissner:
Hi @GideonZ ,

I spotted and fixed bug in the SID player in combination with a USB keyboard. Details in the PR description.

It was not possible to write an E2E test here since doing so would require a way to send USB events to the device which in turn requires hardware such as a wireless USB receiver or similar, so I resorted to extending the existing unit tests.

Thanks for reviewing and merging.

Best wishes
Christian

PR #738, opened 2026-07-29
Fix stale browser listings and consolidate hardware tests

chrisgleissner:
### Overview

This PR fixes stale file-browser listings and consolidates the associated hardware test infrastructure.

It also preserves the SID freeze-wedge regression coverage from #739, imports the network soak/stress harness, and standardizes test layout, naming, execution, cleanup, and failure propagation.

The changes are combined intentionally into this PR because several of them modify `run-e2e-tests`. Splitting or stacking them into a set of PRs would create avoidable merge and rebase conflicts. Keeping them in one PR simplifies the merge to `test-merge`, while the separate commits keep each responsibility independently reviewable.

### Commits

| Group | Commit | Responsibility |
|---|---|---|
| Browser fix | `0f1fbf07` | Fix stale browser listings after filesystem changes. |
| SID regression coverage | `9274d729` | Preserve #739's SID-mirroring freeze-wedge E2E test as the manual `freeze-menu` suite. Test-only, with no firmware changes. |
| E2E organization | `f8770c9a` | Relocate all hardware E2E files into the centralized `tests/e2e/` tree. |
| E2E reliability | `40860080` | Propagate detected FTP E2E failures to the runner through a non-zero exit status. |
| Network soak coverage | `2ec68a53` | Import the ViviPi U64 network soak/stress harness under `tests/soak/network/`. |
| E2E organization | `b5491f85` | Standardize test naming, documentation, clean-start contracts, and final runner teardown. |
| E2E organization | `84d1919a` | Align the C64 and printer E2E paths with their `software/io/` production packages. |
| E2E reliability | `92d5b98a` | Stabilize UI cleanup and remove redundant per-scenario resets while preserving all behavioral assertions. |
| Network soak coverage | `b91a00c7` | Give soak probes capability-based names and correct their stream, Telnet, and stress behavior. |
| E2E documentation | `b912c05e` | Prefix all 14 suite entry points with concise `# E2E:` purpose comments. |

## Browser staleness: three independent causes

A file browser could display a listing that no longer matched the filesystem, with no way to repair it except by leaving and re-entering the directory.

| What the user sees | Root cause | Fix |
|---|---|---|
| A context-menu rename or delete intermittently does not appear | `TreeBrowser::poll()` returns from its context-menu branch before calling `checkFileManagerEvent()`. During a human-scale editor interaction, the eight-entry observer queue can overflow and drop the operation's own refresh event. | After a context action, mark the browser for a full directory reload instead of trusting the cached list. |
| An uploaded or copied file retains its creation-time size, commonly `0` | `fopen_impl()` announces creation, but a writable `fclose()` announces nothing after committing the final directory-entry size. | Remember the write intent and publish one `eRefreshDirectory` event after close. Read-only closes remain silent. |
| A directory created through FTP `MKD` never appears | `create_dir(Path *, name)` publishes the caller's path and raw argument instead of the resolved entry it created. Absolute and nested FTP paths therefore notify the wrong directory. | Publish the parent path and filename from the resolved `PathInfo`, matching the sibling overload. |

The first symptom is particularly intermittent because unrelated background filesystem traffic determines whether the short observer queue fills while the rename editor is open.

Delete often appears less affected because `eNodeRemoved` can also remove the cached entry directly and force a later refresh. Rename depends on its own notification surviving.

The production change is intentionally small: 20 added lines across:

- `tree_browser.cc`
- `file.h`
- `filemanager.cc`

The context-menu reload is deliberately performed after the action. Draining events while the menu is open would be unsafe because `eNodeRemoved` mutates the cached child list to which the menu still points.

The writable-close event:

- is gated on `FA_WRITE`;
- is emitted once per committed operation rather than once per block;
- captures the parent path before `File::close()` deletes the object.

## Browser refresh E2E

`tests/e2e/filemanager/browser_filesystem_refresh_test.py` maintains three independent observers on the same directory without navigating away:

- the on-device Menu browser through `machine:menu_screen`;
- the Telnet `TreeBrowser`;
- an FTP session parked in the directory.

REST `:info` acts as the independent filesystem oracle.

The suite exercises every relevant mutation path through Menu, Telnet, FTP, and REST origins:

- rename;
- queue-overflow rename;
- delete;
- bulk delete;
- directory creation;
- disk-image creation;
- FTP write and short-write;
- `Copy to...`;
- paste;
- move;
- required failure cases.

For the intermittent context-menu defect, the test deterministically fills the observer queue with ten sibling-directory events while the rename editor remains open, then completes the rename.

Using identical test code on a U64 Elite:

| Firmware | Failing cells | Passing cells | Wall time |
|---|---:|---:|---:|
| Production files reverted | **26** | 88 | 215 s |
| This fix | **0** | **140** | 182 s |

## Centralized hardware tests

The test hierarchy now follows the production subsystem that owns each behavior, rather than the protocol used to exercise it:

```text
tests/
├── README.md
├── e2e/
│   ├── api/
│   ├── filemanager/
│   ├── filesystem/
│   ├── io/
│   │   ├── c64/
│   │   └── printer/
│   ├── monitor/
│   └── network/
└── soak/
    └── network/
```

All 14 E2E suites are registered under stable, kebab-case selectors:

Every suite entry point begins with a concise, uniform `# E2E:` purpose comment so its coverage is clear at a glance.

- 10 remain part of the automatic release gate;
- 4 remain manual because they require an operator decision, elevated privileges, a long-running test, or a known unsafe condition.

Each E2E suite establishes its own initial reset and clean state, so it remains safe when selected independently. Individual scenarios within a suite no longer perform redundant consecutive resets.

After the final selected suite, the runner performs one defensive release, menu close, and reset. This protects the next user even when a suite leaves a nested UI open, such as the Assembly 64 query form.

The input suite now shares:

- one controlled BASIC echo workload across its rate checks;
- one menu-editor session across related assertions.

It continues to verify:

- the exact input sequence at 5, 10, and 20 Hz;
- matrix state;
- atomic batches;
- held-key repeat;
- post-release invariants.

The simplification removes setup churn, not coverage.

The top-level and suite READMEs document:

- placement;
- naming;
- execution;
- extension;
- cleanup;
- failure propagation.

They do so without duplicating the contents of each script's `--help` output.

## SID freeze-wedge coverage

The test-only commit from #739 is registered as the manual `freeze-menu` suite.

It preserves coverage for the known failure that occurs when SID auto-address mirroring is disabled.

The complete product fix belongs in the FPGA address decoder. An unclaimed access within `$D400-$D7FF` must acknowledge the access and return open bus rather than stalling the C64 core.

Until that decoder fix is available, the scenario can require JTAG recovery and therefore remains manual.

## Network soak and stress coverage

The ViviPi network harness is available at:

```text
tests/soak/network/connection_test.py
```

It is kept separate from deterministic E2E tests.

The harness covers:

- ICMP;
- UDP/64 identity discovery;
- the distinct, authenticated TCP/64 DMA command channel;
- Telnet;
- FTP;
- REST;
- the optional modem listener;
- audio and video UDP streams.

These surfaces are exercised through bounded soak and stress profiles.

The canonical probe names describe capabilities rather than a particular product:

- `ident` is the stateless UDP/64 JSON identity exchange;
- `dma` is the stateful TCP/64 binary command service.

The stress profile saturates the four firmware FTP and Telnet slots without exceeding them.

Stream validation ignores only the firmware's documented two-byte ARP-resolution primer. Validation remains strict for every real audio or video packet.

Telnet writes use the UI's actual `+` and `-` adjustment contract and independently verify the resulting value.

Any probe or stream `FAIL` now produces a non-zero process status.

The README documents:

- active mutations;
- cleanup;
- concurrency;
- profiles;
- logging;
- extension rules.

## Validation

- Browser refresh red/green result: **26 failing cells before, 0 after, and all 140 applicable cells green**.
- `software/filemanager/tests`: **18/18** host unit tests pass.
- Final automatic hardware sweep: **10/10 suites passed in 732 s** on a U64 Elite, including:
  - **48/48** input checks;
  - **23/23** PRG context-menu actions;
  - **140/140** browser-refresh cells;
  - a successful final clean-slate teardown.
- Every network soak surface was exercised against the U64 in focused runs using a 30-second duration:
  - the complete read/write soak profile, including identity, DMA, Telnet, FTP, REST, ICMP, audio, and video;
  - the four-slot stress profile;
  - open-session, invalid-input, and incomplete-session modes;
  - modem, standalone audio, standalone video, and focused Telnet read/write.
- The Telnet vanish lane used the same requested 30-second duration, plus its protocol-specific 45-second reap allowance. All four half-open slots recovered after approximately 36 seconds, matching the firmware's 20 s idle plus three 5 s keepalive schedule.
- Audio and video both completed with zero lost, reordered, malformed, or timed-out packets in both the complete profile and focused runs.
- Every E2E and soak Python file passes `python3 -m py_compile`.
- Every E2E shell file and `run-e2e-tests` passes `bash -n`.
- `./run-e2e-tests --list` resolves all 14 suites at their centralized paths.
- No tracked references remain to the previous hardware-test locations:
  - `tools/io/`
  - `tools/api/`
  - `tools/developer/`
- The FTP harness's early `Failure("device unreachable")` path returns a non-zero status.
- ViviPi's focused non-hardware tests for the imported network harness pass **113/113**.

The four manual E2E suites remain opt-in by design.

In particular, `freeze-menu` must not enter the automatic sweep until the FPGA decoder fix is available.


chrisgleissner:
Hi @GideonZ,

The file browser staleness issue we discussed earlier is now fixed. I have also added an E2E regression test to prevent this or similar issues from returning.

As discussed, I also moved all E2E tests into a consistent hierarchy under `tests/e2e`. 

I also added the soak tests from my [ViviPi](https://github.com/chrisgleissner/vivipi) health monitoring project which helped to find and fix many network issues to `tests/soak/network`. That way, they are easy to find and can continue to evolve with this project instead of being in a separate repository.

Thanks for reviewing and merging this PR.

Best wishes,
Christian

PR #739, opened 2026-07-29
Add E2E test for the SID-mirroring freeze wedge

chrisgleissner:
## Summary

Test-only extraction from #734 to test fix of SID-mirroring freeze wedge.

## Context

As discussed, the firmware workaround from #734 will not be merged. It can address only DMA-based accesses, including the freezer, `machine:readmem`/`writemem`, and the monitor. It cannot solve the general case because a real 6510 `POKE` or `PEEK` to the same unclaimed address stalls the C64 core in exactly the same way, with no firmware involved in that path.

The correct fix belongs in the address decoder of the FPGA core, maintained in the separate `ult64` repository. The `$D400-$D7FF` region must acknowledge accesses based on the address range rather than on a successful decoder hit. An unclaimed SID address should therefore return open bus instead of leaving the bus stalled.

This PR preserves only the regression coverage from #734 so that it is not lost. It contains `tools/io/c64/freeze_menu_test.py`, unchanged from that branch, and makes no firmware changes.

## Wiring

The suite is registered in `run-e2e-tests` as `freeze-menu` in **manual** mode.

Until the FPGA decoder fix is available, the "mirroring off" scenario reproduces the actual failure: the DMA access stalls and recovery requires JTAG. It therefore must not be included in a normal `run-e2e-tests` sweep.

The suite can be run explicitly with `-s freeze-menu` or as part of `-a`. Once the FPGA core fix has shipped, it can be changed to `auto` and retained as a permanent regression guard.

## Testing

The following checks pass:

* `bash -n run-e2e-tests`
* `python3 -m py_compile`

Also verified that the suite:

* appears as `manual` in `--list`
* is excluded from a bare `run-e2e-tests` invocation
* can be selected with `-s freeze-menu`

The test was not run against real hardware for this PR. The purpose of `manual` mode is specifically to prevent automated test sweeps from triggering the wedge scenario before the FPGA decoder fix is available.

### DIFF
--- a/run-e2e-tests
+++ b/run-e2e-tests
@@ -16,6 +16,7 @@ TIMEOUT="30.0"
 SELECTED=()
 LIST_ONLY=false
 STOP_ON_FAIL=false
+RUN_MANUAL=false
 
 C_BLUE='\033[1;34m'
 C_GREEN='\033[1;32m'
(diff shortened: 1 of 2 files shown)


chrisgleissner:
Hi @GideonZ ,

I raised this PR as a replacement for https://github.com/GideonZ/1541ultimate/pull/734 . It preserves the E2E of that other PR since it will still be useful for when the VHDL-based fix is implemented.

As soon as the VHDL-side fix that you mentioned is done, I will close the other PR.

Thanks for merging.

Best
Christian

chrisgleissner:
Superseded by #738. The SID freeze-wedge E2E commit is included there as a separate, test-only commit alongside the centralized tests/e2e relocation, avoiding conflicting run-e2e-tests changes while keeping the responsibilities independently reviewable.

Issue #740, opened 2026-07-30
CTRL_CMD_LOAD_REU never returns and leaves the UCI wedged until a power cycle


barryw:
# CTRL_CMD_LOAD_REU never returns and leaves the UCI wedged until a power cycle

## Summary

`CTRL_CMD_LOAD_REU` (`$08`) on the control target (`$04`) never completes. The UCI
command interface stays BUSY indefinitely and cannot be released by the documented
abort sequence — only a power cycle recovers it.

The cause appears to be in `software/io/command_interface/control_target.cc`: the
filename is read from the **reply** buffer (`data_message`) rather than from the
command buffer.

## Environment

- Ultimate 64 Elite
- Firmware 3.14d, FPGA 122, Core 1.49
- REU: reproduced with `RAM Expansion Unit` both **Enabled** (16 MB) and **Disabled**

## Reproduction

Driven entirely over the REST API, so no 6502 code is needed —
`machine:readmem`/`machine:writemem` are DMA cycles on the cartridge bus and reach
`$DF1B-$DF1F` directly.

```bash
HOST=<ultimate-ip>

# push command bytes one at a time into the $DF1D FIFO
# (writemem writes an ascending span, so a single multi-byte write would
#  land on $DF1E and beyond)
for b in 04 08 <filename bytes>; do
  curl -X PUT "http://$HOST/v1/machine:writemem?address=df1d&data=$b"
done

# ControlPushCommand
curl -X PUT "http://$HOST/v1/machine:writemem?address=df1c&data=01"

# poll status ($DF1C only -- reading a span across $DF1E/$DF1F pops those FIFOs)
curl "http://$HOST/v1/machine:readmem?address=df1c&length=1" | xxd -p
```

### Observed

`$DF1C` settles at `$11` (Busy | NewCommandSet) and never leaves it. After the
client gives up it reads `$15` (Busy | AbortSet | NewCommandSet).

Recovery attempts, none of which release Busy:

| Write to `$DF1C` | Result |
|---|---|
| `$04` ControlAbort | `$1D` — AbortSet latches, Busy remains |
| `$08` ControlClearError | `$15` — error bit clears, Busy remains |
| `$02` ControlDataAccept | no change |
| `$00` strobes low | no change |

Only a power cycle restores `$DF1B-$DF1F` to `0b 00 c9 00 00`.

Every other UCI command tested on the same machine behaves correctly and returns to
[...]

PR #741, opened 2026-07-31
Fix UCI wedge on LOAD_REU/SAVE_REU and SoftIEC reply framing

chrisgleissner:
Fixes https://github.com/GideonZ/1541ultimate/issues/740.

While investigating the bug reported in that issue, I also reviewed the rest of `software/io/command_interface` for the same class of problem and found three additional defects. This PR fixes all four.

Three fixes were demonstrated failing and then passing on an Ultimate 64 Elite by reverting exactly one change at a time. The fourth fixes a SoftIEC heap buffer overrun. That defect cannot be made observably red on hardware, for the reasons explained in section 4, but is included anyway to ensure code correctness and to prevent future issues.

The firmware change is 13 added and 5 removed lines across three files, including comments. Everything else is test coverage.

## 1. `CTRL_CMD_LOAD_REU` and `CTRL_CMD_SAVE_REU` never return (#740)

### Cause

`struprt()` in `control_target.cc` never advanced its pointer:

```c
char *next = str;
while (*str != '\0')
    *str = toupper((unsigned char)*str);   // str is never incremented
````

`LoadREU()` and `SaveREU()` write a status string into the reply buffer, after which `parse_command()` uppercases that string in place. Because the loop repeatedly reads the same byte, it never terminates for any non-empty string.

`CommandInterface::run_task()` calls the target handler inline, so the failure wedges the entire UCI server task rather than only the current command:

* `HANDSHAKE_ACCEPT_COMMAND` is never written.
* `CMD_NEW_COMMAND` is never cleared.
* The task never reaches the code that services an abort.

This explains both parts of the report: `$DF1C` remains at `$11`, and ControlAbort, ControlClearError, and ControlDataAccept cannot release it because they are handled by the same loop that is spinning.

The issue report narrowed the defect to the correct seven lines. Its register traces and four-condition table made the problem quick to confirm.

The report interpreted `data_message.message + 4` as a filename, which is understandable from `control_target.cc` alone but is not what the argument represents. `REUPreloader::LoadREU(char *status)` receives an output buffer for status text and reads the image path from the "REU Preload Image" setting. Passing the reply buffer at offset 4 is therefore deliberate: the reply contains a 4-byte little-endian transferred count followed by status text. The parameter name is only visible in `reu_preloader.h`, so the call site otherwise looks like a filename is being passed.

The report is correct that the documented `<FILENAME>` argument never reaches the handler. That is a genuine but separate gap and is not changed here.

### Fix

Advance the pointer in `struprt()`.

### Verification against the issue reproduction

The `issue-740-matrix` scenario exercises the same four conditions listed in the report. It sends the same command through the same registers and requires every case to leave Command Busy and return to Idle.

**Red**, with only the `struprt` fix reverted:

```text
[06] issue-740-matrix: REU Enabled, image absent, name at offset 2 ... FAIL
     command 04 08 52 45 55 2e 49 4d 47 never left Command Busy after 15s:
     $11 (Command Busy, CMD_BUSY)
     command interface left in $15 (Command Busy, CMD_BUSY|ABORT_P), wanted Idle
```

These are the same values recorded in the issue: `$11` while polling, then `$15` after the client gives up and requests an abort. The suite teardown attempts DATA_ACC, CLR_ERR, and ABORT. None releases Busy, reproducing the issue's recovery table. The run stops at that point because the interface is wedged for every target.

**Green**, with all six commands completing in under 0.2 seconds each:

```text
[06] REU Enabled,  image absent,  name at offset 2  -> b"\xff\xff\xff\xffREU LOAD: FAILED TO OPEN /TEMP/UCI_E2E_ABSENT.REU: FILE DOESN'T EXIST"  85,REU FILE CANNOT BE OPENED
[07] REU Enabled,  image absent,  name at offset 4  -> same
[08] REU Disabled, image absent,  name at offset 2  -> b'\xfe\xff\xff\xff'  84,REU NOT ENABLED
[09] REU Disabled, image absent,  name at offset 4  -> same
[10] REU Enabled,  image present, name at offset 2  -> b'\x00@\x00\x00REU LOAD: LOADED 16384 BYTES TO $000000 FROM /TEMP/UCI_E2E_PRESENT.REU'  00,OK
[11] REU Enabled,  image present, name at offset 4  -> same
```

Rows 10 and 11 also verify the command's intended behaviour: a 16384-byte image produces a reported transfer count of exactly 16384 bytes. This is the first successful REU load exercised by the suite.

### Why a freshly restarted device may not reproduce the hang

With the REU disabled immediately after a firmware restart, `LoadREU()` returns before writing a status string. The reply buffer then contains fresh heap memory whose first byte happens to be zero, so the broken uppercase loop exits immediately.

The device used for the issue report had already run other commands, leaving text in the buffer. That exposes defect 2 below. The `load-reu-disabled` scenario reproduces it deterministically by dirtying the buffer first.

### Fixture differences from the issue report

The transport is identical to the issue reproduction:

* Command bytes are written to `$DF1D`, one `machine:writemem` call at a time.
* The command is pushed with `$DF1C = $01`.
* Only `$DF1C` is polled, for the reasons given in the report.
* No 6502 code is involved on either side.

The differences are limited to the fixture:

|                | Issue #740                                                                           | This suite                                                            | Reason                                                                                                                                                                                          |
| -------------- | ------------------------------------------------------------------------------------ | --------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Image file     | Present in the working directory                                                     | A 16384-byte file uploaded to `/Temp` over FTP and deleted afterwards | The run must not depend on existing device contents and must leave nothing behind. FTP is used because the REST API can create disk images but cannot delete a file.                            |
| REU size       | 16 MB                                                                                | 128 KB, preload offset 0                                              | The transferred count then depends only on file size, allowing an exact assertion of 16384. The code path is unchanged; the size only changes how many bytes `LoadREU()` requests.              |
| Recovery probe | ControlAbort, ControlClearError, and ControlDataAccept tried manually after the hang | No separate test step                                                 | After the fix, the command completes and one DATA_ACC returns the interface to Idle. On a red run, teardown still tries all three and reports the interface stuck at `$15`, matching the issue. |
| Status names   | Busy, NewCommandSet, AbortSet                                                        | Command Busy, CMD_BUSY, ABORT_P                                       | These are the same bits, named as in the register API document.                                                                                                                                 |

The report also exercised DOS, ECHO, and control IDENTIFY to show that only LOAD_REU failed. Equivalent coverage comes from the `transport` and `control-target` groups, which test IDENTIFY, an unregistered target, an unimplemented command, and rejection paths. The `interface-usable-after` group repeats IDENTIFY after the REU commands have completed.

## 2. The REU reply length comes from the previous command

### Cause

When the REU is disabled, `LoadREU()` and `SaveREU()` return `-2` without writing a status string. `parse_command()` nevertheless called `strlen()` on that buffer and used the result as the reply length.

As a result, both the reply length and any trailing reply content came from whichever command had most recently used the same buffer.

### Fix

Clear the first byte of the status area before calling the handler. A handler that writes no status text now produces a 4-byte reply containing only the little-endian return value.

### Verification

**Red**, with the `struprt` fix present and only the buffer clear removed:

```text
[21] 04 28 00                    data b'Ultimate 64 Elite'
[22] 04 08 52 45 55 2e 49 4d 47  data b'\xfe\xff\xff\xffMATE 64 ELITE'
     FAIL: expected a 4-byte reply, got 17 bytes (stale text from the previous command)
```

`MATE 64 ELITE` is the product name from offset 4, uppercased by the same conversion. If the leftover text is longer, the reported reply length grows with it.

**Green**: `04 08 ...` and `04 09 ...` both return exactly `b'\xfe\xff\xff\xff'`, which is `-2` in little-endian form.

## 3. SoftIEC replies are announced as Data More

### Cause

`SoftIECTarget::data_message.last_part` was never assigned in `softiec_target.cc`. The object has static storage, so the flag initially remained `false`, causing `copy_result()` to write `HANDSHAKE_VALIDATE_MORE`.

Every single-part SoftIEC reply, including LOAD_SU, was therefore announced as Data More.

A client following the register API accepts the data and waits for another block. The next call reaches `get_more_data()` without an input channel and returns `03`, meaning not in data mode. The interface recovers, but the client consumes one spurious status byte for every affected command.

The separate `direct_message` object is the path that genuinely streams. `prepare_data()` assigns its flag for each block. Only `data_message` lacked an initial value, and nothing in the file later sets it to `false`.

### Fix

Set `data_message.last_part` to `true` in the constructor.

### Verification

**Red**, with the first two fixes present and only this assignment removed:

```text
[27] softiec-single-part-reply: LOAD_SU on a missing file completes in one part ... FAIL
     command 05 10 00 00 00 00 00 00 4e 4f 53 55 43 48 46 49 4c 45 replied in state
     $70 (Data More, STAT_AV); this reply is sent in one part, so the state has to
     be Data Last
```

**Green**: the same command returns status `01` in state `$60`, Data Last. GET_FATNAME returns `b'/buffer'`, also in Data Last.

## 4. The SoftIEC reply buffer is smaller than the bound used by its writes

### Cause

`cmd_get_fatname()` fills the reply buffer in two places using:

```c
strncpy((char *)data_message.message, work.c_str(), CMD_MAX_REPLY_LEN);   // n = 896
```

The constructor allocated only 512 bytes:

```c
data_message.message = new uint8_t[512];
```

`strncpy()` does not stop after the source terminator. It pads the destination with zeroes up to `n`. Every GET_FATNAME call therefore writes 896 bytes into a 512-byte heap block, overrunning it by 384 bytes regardless of whether the resolved path is long or short.

### Fix

Allocate `CMD_MAX_REPLY_LEN` bytes, matching the bound already passed by both `strncpy()` calls. `HttpTarget` already sizes its reply buffer this way.

This restores the invariant directly: the buffer size equals the maximum length used by every write to it.

### Why there is no red run

The other three defects have failing runs recorded above. This one does not, because the overrun is silent rather than because it was not checked.

`strncpy()` copies the correct path first and then writes zero padding past the allocation. The reply read by the client is therefore still correct. Nothing changes observably on the C64 side or through REST. The corruption affects whatever heap allocation follows this block and may only surface later as an apparently unrelated failure.

I also checked whether `data_message.length = work.length()` could expose the problem as an over-long reply. It cannot: `ConstructPath()` bounds the result well below the buffer size. On hardware, an 880-character request produced a 53-byte reply:

```text
name  880 -> cmd 883B, reply 53B, status b'\x00'
   head b'/Temp/AAAAAAAAAAAAAAAAAAAAAAAA' ... tail b'AAAAAAAAAAAA.???'
```

The length assignment is therefore unchanged. Only the allocation is corrected.

Verification for this fix consists of:

* A successful firmware build.
* Rechecking the affected reply paths on hardware.
* Confirming that the full suite still passes.

```text
05 22 02 54 45 53 54          data b'/Temp/TEST.???'            status b'\x00'  Data Last, back to Idle
05 22 02 4c 4f 4e 47 ...      data b'/Temp/LONGISHNAMEHERE.???'  status b'\x00'  Data Last, back to Idle
05 22 02 23                   data b'/buffer'                    status b'\x00'  Data Last, back to Idle
05 23 02 54 45 53 54          data b'\x00\x02TEST'               status b'\x00'  Data Last, back to Idle
```

The first two commands exercise the `strncpy()` path changed here. The `#` case in the suite uses a `sprintf()` path and does not cover it, which is why the affected paths were also run separately.

## 5. Misspelled register name in the register dump

`dump_registers()` printed `CMD_IF_REPSONSE_LEN_H` while reading `CMD_IF_RESPONSE_LEN_H`.

The correction changes one word and has no behavioural effect, but it ensures that anyone searching the source for the register name shown in the log can find it.

## Test suite

`tests/e2e/io/command_interface/uci_targets_test.py` is registered as `uci-targets`. It performs 36 checks in 8 groups and takes about 33 seconds.

| Group                                    | Coverage                                                                                                                                                                                                                                          |
| ---------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `transport`                              | Empty command; unregistered target `$0F`; the no-reply bit; ERROR set by pushing while a reply is pending and cleared by CLR_ERR; verification that the rejected push leaves nothing in the command queue; ABORT returning the interface to Idle. |
| `control-target`                         | IDENTIFY; an unimplemented command; both REU commands below the five-byte minimum; GET_HWINFO with a device number the product does not have.                                                                                                     |
| `issue-740-matrix`                       | Defect 1 across the four conditions from the issue, using both filename offsets.                                                                                                                                                                  |
| `save-reu-offset-past-end`               | Defect 1 on the SAVE path, using the branch that reports the offset before accessing storage.                                                                                                                                                     |
| `load-reu-disabled`, `save-reu-disabled` | Defect 2, requiring a 4-byte reply after deliberately dirtying the buffer.                                                                                                                                                                        |
| `softiec-single-part-reply`              | Defect 3 for LOAD_SU and GET_FATNAME.                                                                                                                                                                                                             |
| `interface-usable-after`                 | Confirms that the interface still serves other targets after the REU commands.                                                                                                                                                                    |

### Test design notes

* Apart from the 16384-byte image described above, every scenario is read-only with respect to device storage. LOAD_REU targets a path whose absence is verified through `files:info` rather than assumed. SAVE_REU is used only on the two paths that return before any rename or write, so the suite never creates a multi-megabyte image. GET_FATNAME targets the `#` buffer stream, which is resolved without accessing a partition, making the result independent of the device's SoftIEC setup.

* The ERROR check strobes PUSH_CMD directly instead of queueing bytes first. The FPGA advances the command pointer on every write to `$DF1D`, even when it rejects the push, while only the data and status queues are reset. Queueing bytes before the rejected push would therefore leave them in front of the next command. A subsequent GET_HWINFO verifies that nothing was left behind.

* The suite records the five settings it changes, restores them, reads them back for confirmation, deletes the uploaded file, and returns the interface to Idle. If accepting the data is insufficient, teardown escalates to ABORT. Any failure in cleanup or restoration fails the run.

* The suite opens one connection for each register access, roughly 900 connections per run. This can briefly encounter a full HTTP connection pool. Hardware measurements show that about four connections are served concurrently and that a stuck connection is reaped after roughly 15 seconds.

  Requests that can be repeated without changing device state, namely reads of `$DF1C` plus configuration and file queries, are retried up to three times at two-second intervals, with each retry logged. Writes to the command queue and reads that pop `$DF1E` or `$DF1F` are never retried because a timeout does not reveal whether the first request took effect. This behaviour surfaced once as a single `readmem` timeout at the end of a complete eleven-suite sweep.

* Expected values were derived from the firmware and confirmed on hardware. The manuals under `doc/` were used as background, but they have drifted from the implementation in places and are therefore not treated as authoritative.

## Verification

All hardware runs used an Ultimate 64 Elite with FPGA 122, core 1.4B, and firmware built from this branch. Each build was deployed over JTAG.

For each observable defect, the red run was produced by reverting exactly one change, rebuilding, deploying, and running the same suite. The corresponding green run restored that change and repeated the process.

```text
red 1  (struprt reverted)        issue-740-matrix FAIL at $11, teardown stuck at $15
red 2  (buffer clear reverted)   load-reu-disabled FAIL
red 3  (last_part reverted)      softiec-single-part-reply FAIL
green  (all three)               uci_targets_test: OK (36 checks)

section 4 (reply buffer size)    no red possible; see "Why there is no red run";
                                 affected reply paths rechecked on hardware,
                                 suite still 36/36
```

The complete automatic sweep through `run-e2e-tests` passes on the final build.

## Found but not fixed here

The review of `software/io/command_interface` found two additional issues. Neither can currently be demonstrated failing on hardware, so neither is changed in this PR. They are documented here so that the findings and the decision to defer them are explicit.

The SoftIEC buffer overrun originally belonged in this list, but is now fixed in section 4. It is the deliberate exception to the "no red, no change" rule because leaving it would permit heap corruption rather than merely produce an incorrect reply.

### 1. `HttpTarget::cmd_exchange()` does not set `data_message.last_part` on the object path

This is the same defect as section 3 in a different target, but it is masked in normal use. `cmd_header_create()`, which a client must call before an exchange, leaves the flag set to `true`.

Reaching the faulty state requires a client to abandon a raw exchange part way through and then begin an object exchange. I could not turn that sequence into a reliable test, so the code is unchanged. The eventual fix would be one line.

### 2. `CommandInterface::copy_result()` does not bound the reply length

`data->length` and `status->length` are passed directly to `memcpy()` operations targeting fixed-size FPGA queues.

Defect 2 was one path to an unbounded length. With that fixed, and with `ConstructPath()` measured to keep the SoftIEC reply well below its buffer, I found no remaining reachable trigger. Adding a bound here would therefore be hardening rather than a demonstrated bug fix, and there is no failing case to justify including it in this PR.

It remains worth considering separately because this shared choke point could contain any future target that reports a length larger than the data it can safely provide.

### DIFF
--- a/run-e2e-tests
+++ b/run-e2e-tests
@@ -46,6 +46,11 @@ SUITES=(
   "prg-load-path-trim|tests/e2e/api/prg_load_path_trim_test.sh|-H @HOST@ -p @PASS@"
   "temp-auto-cleanup|tests/e2e/filemanager/temp_auto_cleanup_test.sh|-H @HOST@ -p @PASS@"
   "printer|tests/e2e/io/printer/printer_test.py|-H @HOST@ -p @PASS@"
+  # Drives the UCI control and SoftIEC targets over $DF1B-$DF1F; reproduces #740.
+  # A #740 regression leaves the interface stuck in Command Busy for every target
+  # until the firmware restarts, so run it after the suites that drive the menu
+  # and the file system.
+  "uci-targets|tests/e2e/io/command_interface/uci_targets_test.py|-H @HOST@ -p @PASS@ -t @TIMEOUT@ --test all"
   # Long telnet run; drives the machine-code monitor with its own long flags.
   "machine-code-monitor|tests/e2e/monitor/monitor_test.py|--host @HOST@ --password @PASS@ --timeout @TIMEOUT@|manual"
   # Needs passwordless "sudo -n ip" plus an --iface/--victim-ip for this LAN,
--- a/software/io/command_interface/command_intf.cc
+++ b/software/io/command_interface/command_intf.cc
@@ -214,7 +214,7 @@ void CommandInterface :: dump_registers(void)
     printf("CMD_IF_STATUS_END      %b\n", CMD_IF_STATUS_END    );
     printf("CMD_IF_STATUS_LENGTH   %b\n", CMD_IF_STATUS_LENGTH );
     printf("CMD_IF_RESPONSE_LEN_L  %b\n", CMD_IF_RESPONSE_LEN_L);
-    printf("CMD_IF_REPSONSE_LEN_H  %b\n", CMD_IF_RESPONSE_LEN_H);
+    printf("CMD_IF_RESPONSE_LEN_H  %b\n", CMD_IF_RESPONSE_LEN_H);
     printf("CMD_IF_COMMAND_LEN_L   %b\n", CMD_IF_COMMAND_LEN_L );
     printf("CMD_IF_COMMAND_LEN_H   %b\n", CMD_IF_COMMAND_LEN_H );
     printf("CMD_IF_IRQMASK         %b\n", CMD_IF_IRQMASK       );
--- a/software/io/command_interface/control_target.cc
+++ b/software/io/command_interface/control_target.cc
@@ -30,9 +30,8 @@ static Message c_status_reu_not_saved       = { 31, true, (uint8_t *)"86,REU OFF
 
 char *struprt(char *str)
 {
-    char *next = str;
-    while (*str != '\0')
-        *str = toupper((unsigned char)*str);
+    for (char *next = str; *next != '\0'; next++)
+        *next = toupper((unsigned char)*next);
     return str;
 }
 
@@ -299,6 +298,10 @@ void ControlTarget :: parse_command(Message *command, Message **reply, Message *
             *reply = &c_message_empty;
             if (command->length >= 5) {
                 *reply  = &data_message;
+                // LoadREU/SaveREU return without writing a status string when the REU is
+                // disabled, so clear it here; otherwise the reply length is taken from
+                // whatever the previous command left in this buffer.
+                data_message.message[4] = 0;
                 if (command->message[1] == CTRL_CMD_LOAD_REU) {
                     retVal = reu_preloader->LoadREU((char *)data_message.message + 4);
                 } else {
--- a/software/io/command_interface/softiec_target.cc
+++ b/software/io/command_interface/softiec_target.cc
@@ -23,7 +23,12 @@ static Message c_status_invalid_dir         = {  1, true, (uint8_t *)"\x09" };
 SoftIECTarget :: SoftIECTarget(int id)
 {
     command_targets[id] = this;
-    data_message.message = new uint8_t[512];
+    // cmd_get_fatname() fills this with strncpy(..., CMD_MAX_REPLY_LEN), and strncpy
+    // pads the destination all the way to n, so the buffer has to be that size.
+    data_message.message = new uint8_t[CMD_MAX_REPLY_LEN];
+    // No command in this target replies in more than one part; direct_message is
+    // the one that does, and it sets its own flag in prepare_data().
+    data_message.last_part = true;
     status_message.message = new uint8_t[80];
     input_channel = NULL;
     input_length = 0;
--- a/tests/e2e/README.md
+++ b/tests/e2e/README.md
@@ -20,7 +20,7 @@ protocol used to drive it:
 | `api/` | `software/api/` | REST contracts for input, menu screen, memory, and PRG runners |
 | `filemanager/` | `software/filemanager/`, `software/userinterface/` | Browser actions, change notification, and managed `/Temp` lifecycle |
 | `filesystem/` | `software/filesystem/` | Filesystem implementations, including the remote FTP filesystem |
-| `io/` | `software/io/` | Device-facing I/O subsystems; nested by production package (`c64/`, `printer/`) |
+| `io/` | `software/io/` | Device-facing I/O subsystems; nested by production package (`c64/`, `command_interface/`, `printer/`) |
 | `monitor/` | `software/monitor/` | Machine-code monitor behavior over the normal Telnet UI |
 | `network/` | `software/network/` | Network service and connection lifecycle |
 
--- a/tests/e2e/io/command_interface/uci_targets_test.py
+++ b/tests/e2e/io/command_interface/uci_targets_test.py
@@ -1,0 +1,940 @@
+#!/usr/bin/env python3
+# E2E: Verifies UCI transport state machine, command completion, and reply framing.
+
+"""End-to-end check of the Ultimate Command Interface over the cartridge registers.
+
+Regression guard for GideonZ/1541ultimate#740: CTRL_CMD_LOAD_REU ($08) and
+CTRL_CMD_SAVE_REU ($09) never returned, so the command interface stayed BUSY for
+every target until the machine was power-cycled.
+
+It also covers the transport state machine, the control target's rejection paths,
+and reply framing on the SoftIEC target, whose single-part replies were announced
+as "Data More" and left a client waiting for a block that is never sent.
+
+Every expected value here was taken from the firmware and confirmed against a
+real device. The manuals under doc/ ("Ultimate Command Interface - Register API",
+and the per-target summaries) describe the intended design and are useful
+background, but they have drifted from the implementation in places, so they are
+not used as the source of truth. Where the two differ, this suite follows the
+firmware and says so at the assertion.
+
+The registers live at $DF1B-$DF1F and are reached through REST
+machine:readmem / machine:writemem, which perform DMA cycles on the cartridge
+bus. No 6502 code is involved.
+
+  $DF1C  write: control (bit 0 PUSH_CMD, bit 1 DATA_ACC, bit 2 ABORT, bit 3 CLR_ERR)
+         read:  status  (bit 0 CMD_BUSY, bit 1 DATA_ACC, bit 2 ABORT_P, bit 3 ERROR,
+                         bits 4-5 state, bit 6 STAT_AV, bit 7 DATA_AV)
+  $DF1D  write: command byte queue;  read: identification ($C9, or $49 on IRQ)
+  $DF1E  read:  response data queue
+  $DF1F  read:  status data queue
+
+Command bytes go into $DF1D one REST call at a time on purpose: machine:writemem
+writes an ascending span, so a single multi-byte write would land on $DF1E onwards.
+Only $DF1C is polled for the same reason -- a span read starting there would pop
+the response and status queues.
+
+Supported on any Ultimate whose FPGA provides the command interface. The suite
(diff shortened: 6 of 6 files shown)


chrisgleissner:
Hi @GideonZ ,

thanks for reviewing and merging this PR which contains 3 UCI bug fixes and one typo fix.

Thanks
Christian

PR #743, opened 2026-08-02
Keep long file extensions when copying between filesystems

### LINKED ISSUE #652: File browser mangles filenames with dots, enforces three-char extensions
**Describe the bug**

When using the C64 Ultimate's built-in file browser to copy some files from one storage medium to another (for example going from USB0 to SD), if a filename contains a dot, the portion after the dot gets truncated to three characters as if it's a traditional filename extension, which results in a file copy error if there are multiple files in the copy list that would end up with the same names when so-mangled.

Only the "extension" after the dot is affected -- the part preceding the dot is preserved.

If the filenames do not contain any dots/extensions, their names remain intact when copied.

This appears to be directly related to the old "filename.prg" naming method once demanded by some emulators.

**To Reproduce**

1. by any means you see fit, create two files on a USB thumbdrive:  "foo.bar1234" and "foo.bar5678"
2. plug in a second storage device, say an SD card or maybe just another thumbdrive
3. use the file browser to attempt to copy both files from one to the other

Result:

The first file that gets copied to the destination drive will end up named "foo.bar" without any of the trailing digits, both in the file browser and in the C64 context.  The second file will just generate an error when copied.

On the other hand, if you add ".prg" extensions to the ends of those file names, making them "foo.bar1234.prg" and "foo.bar5678.prg", then those files copy over just fine, and the extra ".prg" extension is hidden from the C64's context, as one might expect.

**Expected behavior**

Filenames should be carried over to the destination drive exactly as the appear on the source drive, regardless of the filename's contents (unless the destination drive imposes some restriction).

I get that some C64 programs may require their files to appear as `PRG` filetype in the `$` directory, and hence needing the ".prg" extension, but this isn't universal, and enforcing that naming method in the file copier is really not a good idea, especially if the user will be frequently working with files that defy this convention.

**Applies to:**

 - Hardware type: C64 Ultimate, maybe also Ultimate64 and/or 1541U
 - Firmware Version:  3.14 as supplied on the machine
 - Git Hash:  b169667b  
 
 (note: you have a typo in your issue template, it has "Gash" when you meant to write "Hash")

### LINKED ISSUE #725: DOS_CMD_COPY_HOME_PATH ($17) always reports "00,OK" even when the home directory doesn't exist / can't be resolved
**Describe the bug**
`DOS_CMD_COPY_HOME_PATH` (UCI DOS1 command `$17`) is supposed to change into the
user-configured "Home Directory" (Settings -> User Interface Settings -> Home
Directory) and report success/failure like `DOS_CMD_CHANGE_DIR` does. In practice
it can never report failure, and in some cases the cached home path it uses is
stale from boot. Both issues make the command unusable for client software that
wants to reliably jump to (or detect the absence of) the user's home directory.

### Bug 1: the CHANGE_DIR failure status is discarded (fallthrough overwrites it)

In `software/filemanager/dos.cc`, the `$17` case calls the shared `cd()` helper
and then deliberately falls through into the `DOS_CMD_GET_PATH` case:

```cpp
case DOS_CMD_COPY_HOME_PATH:
    destination = (char*) HomeDirectory::getHomeDirectory();
    strncpy((char*) command->message + 2, destination, strlen(destination) + 1);
    command->length = 2 + strlen(destination) + 1;
    cd(command, reply, status);
    // fallthrough

case DOS_CMD_GET_PATH:
    strcpy((char *) data_message.message, path->get_path());
    data_message.length = strlen((char *) data_message.message);
    data_message.last_part = true;
    *reply = &data_message;
    *status = &c_status_ok;   // <-- unconditionally overwrites whatever cd() set
    break;
```

`cd()` (same file, a few lines above) correctly detects a missing/invalid
directory, reverts `path` to the previous location, and sets
`*status = &c_status_no_such_dir`. But the fallthrough into `DOS_CMD_GET_PATH`
immediately clobbers that status back to `c_status_ok` regardless of whether the
`cd()` actually succeeded. The data reply is `path->get_path()` taken *after*
`cd()` has already reverted on failure, so the client gets `00,OK` plus a path
string that's silently just wherever it already was -- there is no way to tell
success from failure over the wire.

### Bug 2: the cached home path is only ever resolved once, at boot, and never refreshed

In `software/userinterface/home_directory.cc`, `HomeDirectory::getHomeDirectory()`
returns a static (file-scope) buffer that is populated exactly once, in the
constructor:

```cpp
HomeDirectory::HomeDirectory(UserInterface *ui, TreeBrowser *browser)
{
    ...
    path = fm->get_new_path("HomeDirectory");
    path->cd(ui->cfg->get_string(CFG_USERIF_HOME_DIR));
    setHomeDirectory(path->get_path());
    ...
}
```

The `HomeDirectory` object is constructed once during UI init and then
self-deletes at the end of its background `poll()` task (`delete(this);`).
`poll()` itself only calls `browser->cd(path->get_path())` for the one-time
on-screen "enter home at startup" jump -- it never calls `setHomeDirectory()`
again. Since no other code path updates the static buffer, **any change to the
Home Directory setting made after boot (e.g. via the C=+HOME "save current
location as home" shortcut) has no effect on `DOS_CMD_COPY_HOME_PATH` until the
device is rebooted** -- the command keeps handing back whatever was resolved at
the previous boot.

**To Reproduce**
1. Boot the device, then change the Home Directory setting at runtime (e.g. via
   Settings -> User Interface Settings, or the C=+HOME shortcut) without
   rebooting.
2. From a cartridge or client sending raw UCI DOS1 commands, send target `$01`,
   command `$17` (`DOS_CMD_COPY_HOME_PATH`), no payload.
3. Observed behavior: status is always `00,OK`, and the returned path is
   unaffected by the just-made setting change -- it reflects whatever was
   cached at boot (bug 2), and separately, even a genuinely failed `cd()` inside
   the command would be unreportable since the fallthrough always forces
   `00,OK` (bug 1).

This was observed via a throwaway single-byte test on a client cartridge (only
the command byte sent to the UCI was changed, $17 in place of a known-working
$12 GET_PATH) -- confirming the discrepancy is on the firmware side, not the
client. The specific mechanisms above (the status fallthrough, and the
one-shot uncached-refresh constructor) are from reading the current `master`
source, not from runtime instrumentation/debugging.

**Expected behavior**
Home is saved and GET_PATH API will be able to return it.

 - Hardware type: C64 Ultimate
 - Firmware Version 1.1

**Additional context**
## Suggested fix

- Bug 1: don't fall through past the `cd()` status -- either duplicate the
  `GET_PATH`-style reply logic without re-assigning `*status`, or `break`
  immediately when `cd()` failed and let the caller see
  `c_status_no_such_dir`.
- Bug 2: have `DOS_CMD_COPY_HOME_PATH` re-resolve
  `ui->cfg->get_string(CFG_USERIF_HOME_DIR)` live on each call instead of
  reading the boot-time-only cached buffer.

## Context

This looks related to the fragility already seen around this feature in
#332 ("This might fix HomeDir on startup", merged 2023-04-25), where an
unrelated commit had accidentally disabled the home-directory startup task.
`DOS_CMD_COPY_HOME_PATH` was introduced in DOS API v1.1 and appears not to have
had its failure path exercised since.

barryw:
Fixes https://github.com/GideonZ/1541ultimate/issues/652.

## Overview

Copying a file with a long extension in the browser mangles its name. Rebased onto `test-merge` and narrowed to this one issue; the `$17` status fix that was bundled here is now #747.

User-visible failures before this change:

- Copying `foo.bar1234` produces `foo.bar` at the destination.
- Copying a second file whose name differs only after the third character of the extension fails with `FR_EXIST`, because both names collapse onto one.

## Bug

`fcopy()` unconditionally rewrote the destination extension from `FileInfo::extension`. That field means two different things depending on where the file came from:

| Source | `FileInfo::extension` | In the filename? |
|---|---|---|
| FAT | derived from the name, truncated to 3 chars by `get_extension()` | yes |
| D64 / T64 / ISO | the CBM file type (`PRG`, `SEQ`, `REL`, ...) | no |

The call is needed for the second case. For the first it truncates a name that was already complete.

## Fix

- Add `ensure_extension()`, which appends the extension only when the name does not already end in it.
- Call it from `fcopy()` in place of `set_extension()`.

### Compatibility

The rename-during-copy callers (`filetype_bin.cc` ×4, `filetype_crt.cc`, `filetype_vpl.cc`) build the destination name with `truncate_filename()`, which already normalises to the source extension, so the **ROM and cartridge flashing paths are byte for byte unchanged**. There are regression tests for both.

Two deliberate differences:

- A CBM name containing a dot keeps it. `MY.DOC` of type PRG becomes `MY.DOC.prg` instead of `MY.prg`. The old result collided with `MY.TXT` of the same type, so this strictly reduces collisions.
- The extension keeps the case it had on disk instead of being lower-cased. This is invisible to file type detection: those factories compare against `FileInfo::extension`, and the FAT layer uppercases that on read regardless of the name (`filesystem_fat.h:28-31`). Also covered by a test.

`make test` — 168 assertions, no hardware and no FPGA toolchain.

| Suite | Assertions |
|---|---|
| `software/api/tests` | 38 |
| `software/components/tests` | 12 |
| `software/filemanager/tests` | 28 |
| `software/io/usb/tests` | 90 |

Written test-first; each red state was confirmed to reproduce the reported symptom before the fix went in.

Three supporting changes:

- **`record_failure()` dropped non-fatal failure messages.** A failing `EXPECT_*` showed up as nothing but a bump in the failure count, with no way to tell which assertion failed. Fixed first, because otherwise none of the red states above were diagnosable.
- **The fake filesystem moved out of `temp_auto_cleanup_test.cpp`** into `fake_filesystem.h` so more than one test can use it. That accounts for most of the diff size — it is a move, not new code. It also models the FAT/CBM extension difference described above, since that is what the code under test reacts to.
- **`software/api/tests` needed `pattern.cc` on its link line.** It does not currently build on `test-merge` — `url_encode()` is an undefined symbol. Fixed here so it can join the gate rather than be excluded from it.

### CI

The host tests were only reachable by running `make` by hand in each directory, so nothing ran them. Added a `test` job covering all four suites; the hardware builds now `needs: test`, so an hour of FPGA work is not spent on a tree whose logic is already known to be broken.

Placement follows [`tests/README.md`](tests/README.md): these are isolated logic tests, so they live beside the production component rather than under `tests/e2e/`.

Not yet. I have a JTAG programmer arriving and will follow up with `tests/e2e/filemanager/` cases that reproduce the reported interaction directly — copy a long-extension file in the browser, check the destination name — once I can run the suite against a device.

## Notes

Unrelated to this PR, but noticed while working nearby:

- `fcopy()` does `strncpy(dest_name, dest_filename, 100)` with no NUL guarantee.
- `make u64ii` does not currently build on `test-merge`: `c64_crt.cc:19` declares `extern uint8_t __cart_rom_start;` while `c64.h:420` declares it as `uint8_t __cart_rom_start[1024*1024]`, which is a conflicting declaration. Verified against a clean checkout of `test-merge` with none of this branch applied, using the xPack RISC-V GCC 11.3.0 that `docker/Dockerfile` pins. Happy to open a separate issue.


chrisgleissner:
Hi @barryw ,

thanks for your contribution. Much appreciated!

A bunch of comments:
- I see that the build failed. Thanks for investigating. I suggest running the build for u64e2, u64, and u2 (riscv) locally before raising a PR to catch teething issues.
- It's great you added tests. We now have an extensive E2E test infrastructure. I am currently still working on refining and modularizing them in another PR. It would be great if you could add an E2E test once my PR will be merged to `test-merge`, or ideally extend an existing one.
- Please declare `test-merge` as the target branch since all current work targets that branch (which will be used to create 3.15) instead of `master`.
- Please reduce churn (i.e. changes) in the PR to ensure it is highly targeted and only addresses the concrete bugs, plus is consistent with the existing code base. I see that there are a lot of low-level unit tests that you built, and this is great and needed in the medium term, but the benefit of E2E tests is that they do not extend the core app code base and instead interact with the application via its external contract, i.e. the REST/FTP/Telnet APIs.
- If any user-reported issues are closed, it would be great if the E2E tests that you write precisely replicate the user interactions (e.g. open menu, navigate to a certain item, perform a certain activity, etc.). This makes it much easier to assess whether the reported issues were addressed. The two issues which you referenced described such user interactions, and would be easy to turn into E2E tests.

The consolidated E2E tests now live in the `test-merge` branch. Please feel free to extend them as needed. There are various reusable libraries in `tests/lib` and `/tests/e2e/lib`, and you could model it on already existing E2E tests for maximum consistency.

Many thanks
Christian

barryw:
Closing this one. Taking the feedback on board — this PR reached too far for a first contribution, bundling a bug fix, a test harness and a CI change together.

Thanks for the detailed review, Christian. It was genuinely useful, and it turned up things worth knowing regardless:

- `software/api/tests` does not currently build on `test-merge` — `url_encode` is an undefined symbol because the suite does not link `pattern.cc`.
- `make u64ii` does not build on `test-merge` either: `c64_crt.cc:19` declares `extern uint8_t __cart_rom_start;` while `c64.h:420` declares it as `uint8_t __cart_rom_start[1024*1024]`. Verified against a clean checkout with none of this branch applied, using the xPack RISC-V GCC 11.3.0 pinned in `docker/Dockerfile`.

Happy to raise either of those as its own issue if useful.

The underlying bug (#652) is still real and I'll come back to it in a smaller, more targeted form once I have the E2E suite running against real hardware. The branch stays on my fork in the meantime.

PR #747, opened 2026-08-03
Report failure when COPY_HOME_PATH cannot enter the home directory

barryw:
Fixes https://github.com/GideonZ/1541ultimate/issues/725.

## Overview

`DOS_CMD_COPY_HOME_PATH` ($17) reports `00,OK` even when it did not change directory. A client asking to jump to an unreachable home directory is told the command succeeded, and is handed the current path instead of the one it asked for.

Split out of #743 so each change stands on its own.

## Bug

The `$17` case calls `cd()`, which sets the status correctly, and then falls through into `DOS_CMD_GET_PATH`:

```c
case DOS_CMD_COPY_HOME_PATH:
    ...
    cd(command, reply, status);
    // fallthrough

case DOS_CMD_GET_PATH:
    ...
    *status = &c_status_ok;
```

`DOS_CMD_GET_PATH` overwrites the status unconditionally, so the failure from `cd()` is discarded.

## Fix

Break out of the case when `cd()` failed, so the real status reaches the caller. The success path is unchanged.

The failure status is now identical to what `DOS_CMD_CHANGE_DIR` already returns for the same directory — the command this one is documented to mirror.

`make -C software/filemanager/tests dos-command`

```
3 test(s) passed
```

Added to `software/filemanager/tests/dos_command_test.cpp`:

- `ReportsOkWhenHomeDirectoryResolves` — control; the success path must stay green
- `ReportsFailureWhenHomeDirectoryIsMissing`
- `MatchesChangeDirFailureStatus` — pins `COPY_HOME_PATH` to the status `DOS_CMD_CHANGE_DIR` already returns

Confirmed red before the fix. Reverting only the `break` in `dos.cc` gives:

```
2 test(s) failed
```

The root filesystem is enough to tell the two cases apart — `/` always resolves and a name below it never does — so the tests need no fixture filesystem.

`software/test/hoststubs/` makes `dos.cc` host-buildable. `dos.h` → `c1541.h` → `browsable_root.h` drags in the whole UI and the lwIP stack; stubbing at `c1541.h` collapses that subtree in one move. Only headers that genuinely cannot build on a host are replaced — everything else is the real firmware source.

One of those stubs is worth flagging: **`c64.h` cannot be compiled for a 64-bit host.** `get_cartridge_max_rom()` casts pointers to `uint32_t`, which is a hard error off the 32-bit target:

```
c64.h:426:17: error: cast from 'uint8_t*' to 'uint32_t' loses precision [-fpermissive]
```

`dos.cc` only uses `isMP3RamDrive()` and `getSizeOfMP3NativeRamdrive()` from it, neither on any path these tests reach, so the stub declares just those. Happy to change the casts to `uintptr_t` in a separate PR instead if you would rather the real header stayed host-compilable — on the 32-bit target that generates identical code.

Not yet. Per `tests/README.md` this behaviour belongs beside its component as a host test, which is what the above is; an `e2e/io/command_interface/` case driving the same command over the wire would be complementary rather than a replacement. I have a JTAG programmer arriving and will follow up with one.

### DIFF
--- a/software/filemanager/dos.cc
+++ b/software/filemanager/dos.cc
@@ -420,6 +420,11 @@ void Dos::parse_command(Message *command, Message **reply, Message **status) {
                 strlen(destination) + 1);
         command->length = 2 + strlen(destination) + 1;
         cd(command, reply, status);
+        if (*status != &c_status_ok) {
+            // The home directory could not be entered. Report that, rather than
+            // falling through and overwriting the status with "00,OK".
+            break;
+        }
         // fallthrough
 
     case DOS_CMD_GET_PATH:
--- a/software/filemanager/tests/Makefile
+++ b/software/filemanager/tests/Makefile
@@ -2,11 +2,32 @@ CXX ?= g++
 CXXFLAGS := -std=c++14 -g -I. -I.. -I../../components -I../../filesystem -I../../infra -I../../system -I../../userinterface -I../../io/usb/tests
 OUTPUT_DIR := output
 TEMP_AUTO_CLEANUP_TEST := $(OUTPUT_DIR)/tempAutoCleanupTest
+DOS_COMMAND_TEST := $(OUTPUT_DIR)/dosCommandTest
 
-.PHONY: all temp-auto-cleanup
+SOURCES := ../../io/usb/tests/host_test_main.cpp ../filemanager.cc ../path.cc \
+           ../../filesystem/file_system.cc ../../filesystem/filesystem_root.cc \
+           ../../filesystem/directory.cc ../../filesystem/partition.cc \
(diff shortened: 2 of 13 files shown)


barryw:
Closing this alongside #743 while I scope down to something more appropriate for a first contribution.

The fix itself is small — `DOS_CMD_COPY_HOME_PATH` falls through to `DOS_CMD_GET_PATH`, which overwrites the status with `00,OK`, so a failed home-directory change reports success. But it carries a test harness with it that is more than the change warrants, and it is better raised once I can demonstrate it through the E2E suite the way you suggested.

The branch stays on my fork. Thanks again for the review.

Issue #590, opened 2026-01-01
Operator Precedence Error in Modem ATH Command


gglessner:
# Bug: Operator Precedence Error in Modem ATH Command

**File:** `software/io/acia/modem.cc`  
**Line:** 620  
**Severity:** High (Logic Error)

## Description

The ATH (hang up) command has an operator precedence bug that prevents busy mode from working correctly.

## Code

```cpp
case 'H': // hang up
    sscanf(cmd->command + i + 1, "%d", &temp);
    while(i < cmd->length && (isdigit(cmd->command[i+1])))
        i++;
    busyMode = (temp > 0);
    keepConnection = false;
    connectionStateChange = 1 + (busyMode) ? 2 : 0; // if busy mode is selected, return 3
    break;
```

## Problem

The ternary operator `?:` has lower precedence than addition `+`, so the expression:

```cpp
1 + (busyMode) ? 2 : 0
```

Is parsed as:

```cpp
(1 + busyMode) ? 2 : 0   // Always evaluates to 2!
```

Since `1 + busyMode` is always non-zero (either 1 or 2), the result is **always 2**, never 1 or 3.

## Impact

- The comment states the intent: "if busy mode is selected, return 3"
- But `connectionStateChange` never equals 3
- ATH1 (hang up and set busy mode) doesn't work correctly
- Incoming callers won't receive the busy signal

## Fix

```cpp
connectionStateChange = busyMode ? 3 : 1;
```

Or with explicit parentheses:

```cpp
connectionStateChange = 1 + (busyMode ? 2 : 0);
```

PR #749, opened 2026-08-03
Fix operator precedence in the modem ATH handler

barryw:
Fixes https://github.com/GideonZ/1541ultimate/issues/590.

## Overview

`ATH1` drops the connection without ever sending the busy message. An incoming caller gets a silent disconnect instead of the configured busy file.

## Bug

`software/io/acia/modem.cc:627`:

```cpp
connectionStateChange = 1 + (busyMode) ? 2 : 0; // if busy mode is selected, return 3
```

The conditional operator has lower precedence than addition, so this parses as `(1 + busyMode) ? 2 : 0`. Both `1` and `2` are truthy, so the expression is **always 2** — the value 3 named in the comment is never produced.

That matters because the caller only relays the busy file on 3 (`modem.cc:322`):

```cpp
connChange = ExecuteCommand(&modemCommand);
if (connChange) {
    if (connChange == 3) { // ATH1
        RelayFileToSocket(cfg->get_string(CFG_MODEM_BUSYFILE), socket, "The modem you are connecting to is currently busy.\n");
    }
    break;
}
```

So that branch is unreachable today. The hangup itself still works — `keepConnection = false` handles it and any non-zero value breaks the loop — but the busy notification is lost.

## Fix

Write the intent directly:

```cpp
connectionStateChange = busyMode ? 3 : 1;
```

`busyMode` false gives 1, matching `ATZ`; true gives 3, which is what the comment and the caller expect.

## Tests

The precedence claim, demonstrated rather than asserted:

```cpp
for (int busyMode = 0; busyMode <= 1; busyMode++) {
    int old_expr = 1 + (busyMode) ? 2 : 0;
    int new_expr = busyMode ? 3 : 1;
    printf("busyMode=%d   old=%d   new=%d\n", busyMode, old_expr, new_expr);
}
```

```
busyMode=0   old=2   new=1
busyMode=1   old=2   new=3
```

No host test is included deliberately. `modem.cc` reaches FreeRTOS, the socket layer, the config store and the ACIA layer, so covering it on the host would mean a stub harness far larger than the change — which is the kind of churn you asked me to avoid. The behaviour is already covered on hardware by `tests/soak/network/modem_probe.py`, which inspects responses for `busy`.

Verified to compile: `make u64_no_esp` on a clean checkout of this branch, no new warnings in `modem.cc`.

### DIFF
--- a/software/io/acia/modem.cc
+++ b/software/io/acia/modem.cc
@@ -624,7 +624,7 @@ int Modem :: ExecuteCommand(ModemCommand_t *cmd)
                 i++;
             busyMode = (temp > 0);
             keepConnection = false;
-            connectionStateChange = 1 + (busyMode) ? 2 : 0; // if busy mode is selected, return 3
+            connectionStateChange = busyMode ? 3 : 1; // if busy mode is selected, return 3
             break;
         case 'A': // answer
             if (keepConnection) {


chrisgleissner:
@GideonZ This change seems reasonable to me and is very well argued in the PR description, do you agree?

PR #750, opened 2026-08-03
Guard BYTE_ORDER against newlib's definition

barryw:
## Overview

The U64 build emits 17 `"BYTE_ORDER" redefined` warnings. `software/portable/nios/endianness.h:19` defines `BYTE_ORDER` unconditionally, and newlib's `<machine/endian.h>:20` has already defined it in every translation unit that sees both.

## Fix

Guard the definition:

```c
#ifndef BYTE_ORDER
#define BYTE_ORDER LITTLE_ENDIAN
#endif
```

## Why this is safe

For `nios2-elf` the two definitions resolve to the same value, so taking whichever is already defined cannot change behaviour. Rather than argue that from the headers, here it is measured.

`-dM` shows how both sides resolve:

```
#define _LITTLE_ENDIAN 1234
#define _BYTE_ORDER    _LITTLE_ENDIAN     <- newlib, derived from __IEEE_LITTLE_ENDIAN
#define LITTLE_ENDIAN  _LITTLE_ENDIAN
#define BYTE_ORDER     LITTLE_ENDIAN      <- this header, unguarded
```

A translation unit that includes both headers and asserts the value:

```c
#include <stdint.h>
#include <machine/endian.h>
#include "endianness.h"
#if !defined(BYTE_ORDER)
#error BYTE_ORDER_UNDEFINED
#endif
#if BYTE_ORDER != 1234
#error BYTE_ORDER_NOT_1234
#endif
int ok = BYTE_ORDER;
```

```
WITHOUT guard: BYTE_ORDER == 1234 (compiles)
WITH guard:    BYTE_ORDER == 1234 (compiles)
object files identical: YES
```

The `#error` would fire on either side if the value moved. It doesn't, and the two object files are byte-identical.

Warning count on `make u64_no_esp`: **17 before, 0 after**.

## Scope

`software/portable/microblaze/endianness.h` has the same pattern but asserts `BIG_ENDIAN`, where a guard could genuinely change the value. That build needs Xilinx ISE, which I cannot run, so it is deliberately left alone.

## Note

While verifying this I found the firmware build is not reproducible — two `make u64_no_esp` runs from identical source produce different `ultimate.app` hashes:

```
959e1f1bd54a2bf8078390b7acb476fb
ba95800bc346aa33de1d3f4a535f615b
```

Not addressed here, but happy to open a separate issue if it is of interest.

### DIFF
--- a/software/portable/nios/endianness.h
+++ b/software/portable/nios/endianness.h
@@ -16,6 +16,11 @@ __inline uint16_t cpu_to_16le(uint16_t p)
 #define le16_to_cpu(x)  cpu_to_16le(x)
 #define le_to_cpu_32(x) cpu_to_32le(x)
 
+// newlib's <machine/endian.h> already defines BYTE_ORDER, and for nios2-elf it
+// resolves to little endian as well, so defining it unconditionally only
+// produced a redefinition warning in every translation unit that saw both.
+#ifndef BYTE_ORDER
 #define BYTE_ORDER LITTLE_ENDIAN
+#endif
 
 #endif /* ENDIANNESS_H */


chrisgleissner:
@barryw Thanks for that. Could you please add a link to the precise line in the CI build that exhibits the error, and another link to the line which shows the error is got with your fix?

barryw:
Good question, and checking it changes my answer: **I can't give you that link, because the warning does not appear in the CI build at all.**

I searched the log of the `test-merge` run you pointed me at earlier (run 30853818335, `e2264deb`) for `BYTE_ORDER` and got zero hits. The reason looks to be the toolchain:

```
/mnt/build-tools/quartus/18.1/quartus/adm/qenv.sh
```

CI builds with **Quartus 18.1**. I was building with **19.1**, whose Nios II EDS ships a newer GCC that emits the redefinition warning. So this is the same mistake I made on #751 — I verified the fix carefully and never verified that the problem exists in your environment.

Sorry for the repeat. I should have checked the CI log before opening it, and I'll check the premise first from now on.

The change is still safe if you want it for forward compatibility with a newer EDS — the value is provably unchanged (`BYTE_ORDER == 1234` both ways, object files byte identical) — but it fixes nothing you would ever see, so closing it is probably the right call. Your decision.

For what it's worth, the same search did turn up warnings that **are** in your CI log and might be worth a look. From that run:

```
  5  software/components/dict.h:115: 'printf' was not declared in this scope,
     and no declarations were found by argument-dependent lookup [-fpermissive]
  3  ../../../common/rules.mk:45: overriding recipe for target '.../riscv/lwip/result/liblwip.a'
  3  makefile:114: ignoring old recipe for target '.../riscv/lwip/result/liblwip.a'
  2  ../../../common/rules.mk:45: overriding recipe for target '.../nios2/lwip/result/liblwip.a'
  2  makefile:113: ignoring old recipe for target '.../nios2/lwip/result/liblwip.a'
  2  software/io/rtc/rtc_i2c.cc:405: lambda expressions only available with -std=c++11 or -std=gnu++11
```

The `liblwip.a` one is a duplicate rule: `target/libs/*/lwip/makefile` defines `$(RESULT)/$(PRJ).a` and then includes `common/rules.mk`, which defines it again. The include wins, so the makefile's own `$(LINK)` prerequisite is silently dropped.

Happy to look into any of those instead, if they'd be more useful than warnings only I can see.

GideonZ:
Side note: Since the microblaze was ditched and replaced with a proper little endian Risc-V, there are no longer any oddball big endian CPUs in the project.

barryw:
@GideonZ Thanks — that's useful context, and it means the caveat in my description was unnecessary caution. I left `software/portable/microblaze/endianness.h` alone because it asserts `BIG_ENDIAN` and I had no way to build or verify it. If there is no big endian target left, that concern doesn't apply.

It also lines up with the workflow: CI builds `u2_rv`, `u2plus`, `u2pl_no_esp`, `u64_no_esp`, `u64ii_no_esp` and `esp32`, but never `make u2`, so nothing under `target/u2/microblaze/` is exercised.

On this PR itself — as above, the warning it fixes doesn't appear in your CI at all, only under Quartus 19.1. Happy to close it; just say the word, or close it yourself if that's easier.

If the MicroBlaze tree is genuinely retired and you'd ever want it removed, I'm happy to take that on as a separate, self-contained PR. Entirely your call though, and I'd rather not touch it uninvited.

chrisgleissner:
@barryw @GideonZ I think that given the warning does not occur in CI, this can be closed, just as you suggested. Thanks for having submitted it, your research is much appreciated.

PR #751, opened 2026-08-03
Remove unused cart ROM declarations that break the RISC-V builds

barryw:
## Overview

Every RISC-V target fails to compile on `test-merge`:

```
software/io/c64/c64_crt.cc:19:16: error: conflicting declaration 'uint8_t __cart_rom_start'
software/io/c64/c64.h:420:24: note: previous declaration as 'uint8_t __cart_rom_start [1048576]'
make: *** [c64_crt.o] Error 1
```

Reproduced on a clean checkout of `e2264deb` with the xPack RISC-V GCC 11.3.0 pinned in `docker/Dockerfile`.

## Bug

`c64_crt.cc:19-20` declares two symbols at file scope:

```cpp
extern uint8_t __cart_rom_start;
extern uint8_t __cart_rom_limit;
```

Neither is referenced anywhere in the file — those two lines are the only occurrences of either name. The one place that needs the cartridge ROM size calls the accessor instead (`c64_crt.cc:666`):

```cpp
work->initialize(mem, C64::get_cartridge_max_rom());
```

But the scalar declaration of `__cart_rom_start` conflicts with `c64.h:420`, which declares it as an array inside `get_cartridge_rom_addr()`:

```cpp
extern uint8_t __cart_rom_start[1024*1024];
```

`__cart_rom_limit` is scalar in both places and does not conflict; it is removed only because it is equally unused.

These came in with `eaa7751b` (large cartridge support) and are not on `master`.

## Fix

Delete both declarations.

## Tests

**Before**, clean `test-merge`:

```
$ make -C target/u64ii/riscv/ultimate
c64_crt.cc:19:16: error: conflicting declaration 'uint8_t __cart_rom_start'
make: *** [c64_crt.o] Error 1     (exit 2)
```

**After**, both affected targets:

```
target/u64ii/riscv/ultimate     exit 0, 0 errors
target/u2plus_L/riscv/ultimate  exit 0, 0 errors
```

**No behavioural change.** On nios2 the file compiles both ways, so the object files can be compared directly. Same-source rebuilds were first confirmed byte-identical, so the comparison is meaningful:

```
code disassembly identical:   YES
sections differing:           .debug_info only, 0x7f85 -> 0x7f6d
```

The emitted code is byte identical. The only difference is 24 bytes of DWARF — the records describing the two removed declarations.

### DIFF
--- a/software/io/c64/c64_crt.cc
+++ b/software/io/c64/c64_crt.cc
@@ -16,9 +16,6 @@ extern uint8_t _eapi_65_start[768];
 #define CRTCHP_LOAD    0x0C
 #define CRTCHP_SIZE    0x0E
 
-extern uint8_t __cart_rom_start;
-extern uint8_t __cart_rom_limit;
-
 const struct C64_CRT::t_cart C64_CRT::c_recognized_c64_carts[] = {
     {  0, 0xFF, CART_NORMAL,    "Normal cartridge" },
     {  1, 0xFF, CART_ACTION,    "Action Replay" }, // max 4 banks of 8K


chrisgleissner:
@barryw Thanks for raising this. The `test-merge` branch is building, otherwise it could not produce any firmware images. Please have a look at https://github.com/GideonZ/1541ultimate/actions/runs/30853818335/job/91820024722

It is possible that there are warnings, but certainly no errors that prevent firmware creation. Could you pleae double check whether you are building the firmware the same way as our CI/CD build does?

barryw:
You're right, and my claim was wrong — thanks for pushing back.

I checked the run you linked. It builds, and it compiles `c64_crt.cc` for every target. The difference is the compiler:

```
/mnt/build-tools/riscv/bin/../lib/gcc/riscv32-unknown-elf/10.2.0/libgcc.a
```

CI uses **riscv32-unknown-elf GCC 10.2.0** from the bind-mounted `/opt/build-tools`. I was building with **11.3.0** — the version `docker/Dockerfile:21` installs:

```dockerfile
curl -L .../xpack-riscv-none-elf-gcc-11.3.0-1-linux-x64.tar.gz | tar -xzC /opt/riscv
for file in /opt/riscv/bin/riscv-none-elf-*; do
  ln -s "$file" "/usr/local/bin/riscv32-unknown-elf-${file##*riscv-none-elf-}"
```

GCC 10.2.0 accepts the conflicting declaration; 11.3.0 rejects it. `nios2-elf-gcc 8.3.1` accepts it too, which is why the Nios targets were never affected.

So `test-merge` is not broken, and I should have verified against the CI toolchain before saying so. Sorry for the noise.

That leaves a smaller, real observation, which you may or may not think is worth acting on: **the toolchain in `docker/Dockerfile` cannot build the RISC-V targets.** The container installs 11.3.0 and symlinks it into `/usr/local/bin`, but the `/opt/build-tools` mount shadows it on PATH, so CI never exercises what the Dockerfile provides. Anyone building from the documented container image hits this error.

The change itself is still safe if you want it — the two declarations are unused (they are the only occurrences of either name in the file, and the cartridge size comes from `C64::get_cartridge_max_rom()`). On nios2, where the file compiles both ways, the emitted code is byte identical before and after; only `.debug_info` shrinks by the 24 bytes those declarations occupied.

But it is no longer a build fix, so it is entirely your call. Happy to close it, or to retitle it as forward-compatibility with newer GCC. If the Dockerfile/`build-tools` divergence is worth tracking I can raise that separately instead.

chrisgleissner:
@barryw I do think that it would be great if anyone trying to build this project locally could do so without any warnings.

@GideonZ When I first tried to build and contribute to this repo, I also ran into a few issues. As a result, I built up a set of CLI tools which I have been using to drive all of my builds for U64, UE2, and U2L for the last few months.

They can be found at https://github.com/chrisgleissner/1541ultimate-tools and if there is interest, I could raise a PR for these to be merged into this repo.

PR #753, opened 2026-08-05
Remove duplicate liblwip.a rule from the lwip makefiles

barryw:
## Overview

Every CI build emits ten `overriding recipe` warnings for `liblwip.a`, five per architecture. From run [30913782624](https://github.com/GideonZ/1541ultimate/actions/runs/30913782624) on `8efdffe6`:

```
3  makefile:114: warning: ignoring old recipe for target '/mnt/project/target/libs/riscv/lwip/result/liblwip.a'
3  ../../../common/rules.mk:45: warning: overriding recipe for target '/mnt/project/target/libs/riscv/lwip/result/liblwip.a'
2  makefile:113: warning: ignoring old recipe for target '/mnt/project/target/libs/nios2/lwip/result/liblwip.a'
2  ../../../common/rules.mk:45: warning: overriding recipe for target '/mnt/project/target/libs/nios2/lwip/result/liblwip.a'
```

## Bug

Both lwip makefiles define the archive rule and then include `common/rules.mk`, which defines the same target again:

```make
$(RESULT)/$(PRJ).a: $(LINK) $(OBJS_C)
	@echo "Creating Archive $@"
	$(AR) -rc $@ $(ALL_OBJS)

include ../../../common/rules.mk        # rules.mk:45 defines $(RESULT)/$(PRJ).a again
```

The include comes last, so make keeps `rules.mk`'s recipe and discards the local one.

## Fix

Delete the shadowed rule from both makefiles.

## Why this changes nothing

`rules.mk`'s recipe is already the one that executes. The two differ in quoting, and the CI log shows the quoted form running:

```
ar -rc "/mnt/project/target/libs/riscv/lwip/result/liblwip.a" ./output/sys_arch.o ...
```

`rules.mk:45` is `$(AR) -rc "$@" $(ALL_OBJS)`; the lwip makefiles' is unquoted. So the local recipe was already dead, and removing it cannot change what runs.

The only other difference is the `$(LINK)` prerequisite, which make merges even when it discards the recipe:

- **riscv** — `LINK` is empty (`makefile:102`), so the local rule expands to exactly `rules.mk`'s. Identical in every respect.
- **nios2** — `LINK = ./mb_app.lds` (`makefile:101`), a MicroBlaze linker script. It is not an input to `ar`, and the nios2 link step uses `-T'.../nios_appl_bsp/linker.x'` instead.

## Tests

**nios2**, building `target/libs/nios2/lwip` before and after:

```
BEFORE   warnings: 2   archive members md5: 35eea8fcb4fded7169e1cf997ab588d3
AFTER    warnings: 0   archive members md5: 35eea8fcb4fded7169e1cf997ab588d3
```

Same archive, same members.

**riscv**, at parse time via `make -n`, so no cross toolchain is needed:

```
BEFORE: 2 warnings
AFTER:  0 warnings
```

with the `ar -rc` rule still resolving afterwards.

Note that the wording differs by make version — 3.81 says `overriding commands`, 4.x says `overriding recipe` — so a grep for one phrasing will miss the other.

### DIFF
--- a/target/libs/nios2/lwip/makefile
+++ b/target/libs/nios2/lwip/makefile
@@ -109,9 +109,4 @@ VPATH   += $(OUTPUT) $(RESULT)
 
 all: $(OUTPUT) $(RESULT) $(RESULT)/$(PRJ).a
 
-$(RESULT)/$(PRJ).a: $(LINK) $(OBJS_C)
-	@echo "Creating Archive $@"
-	$(AR) -rc $@ $(ALL_OBJS)
-
-
 include ../../../common/rules.mk
--- a/target/libs/riscv/lwip/makefile
+++ b/target/libs/riscv/lwip/makefile
@@ -110,9 +110,4 @@ VPATH   += $(OUTPUT) $(RESULT)
 
 all: $(OUTPUT) $(RESULT) $(RESULT)/$(PRJ).a
 
-$(RESULT)/$(PRJ).a: $(LINK) $(OBJS_C)
-	@echo "Creating Archive $@"
-	$(AR) -rc $@ $(ALL_OBJS)
-
-
 include ../../../common/rules.mk


chrisgleissner:
Thanks @barryw . I can verify that the error is gone for U2; I have not checked the others. 

Just to ensure there are no regressions and since I understand that you did not JTAG-deploy or flash the outcome of your build on a real device, would you mind comparing the size and content of the resulting binaries for all target devices before and after your change, then reporting the results here in the chat?

barryw:
Thanks — done. **All binaries are byte identical before and after, on every target.**

| Artifact | Size | MD5 (before = after) |
|---|---:|---|
| `target/u64/nios2/ultimate` | 1185344 | `6a25d3627d36e00074f7c763c060b1cd` |
| `target/u2plus/nios/ultimate` | 1003780 | `f1f6b8308892897f93ba3f939a60d348` |
| `target/u2/riscv/ultimate` | 879344 | `abb2ac97421d5990f3ec12dc71a1058c` |
| `target/u2plus_L/riscv/ultimate` | 989504 | `c47d779973bc5dddc78a6c05466e6f32` |
| `target/u64ii/riscv/ultimate` | 1155828 | `445d32e395d37c2545b7a3737c8762ec` |
| `target/libs/nios2/lwip/result/liblwip.a` | 1375608 | `17231f09361fc0441efcc5b30a7ff746` |
| `target/libs/riscv/lwip/result/liblwip.a` | 3632396 | `c1d07ebe04e1a0cde7edf3bc320209e8` |

`ultimate.app` for each device, plus both lwip archives since those are what the change touches. Same commit both runs (`v3.15-30-g06a21735`), change applied as an uncommitted working-tree edit so nothing else varied.

## One thing worth knowing: the build is not reproducible by default

My first attempt at this showed every binary **+24 bytes** and I nearly reported that as a real difference. It was not. `target/common/rules.mk:27-31` compiles build metadata in:

```make
echo '#define APP_VERSION_TAG    "'`git describe --tags`'"'  >>$(OUTPUT)/gitinfo.h
echo '#define APP_VERSION_BRANCH "'`git describe --all`'"'   >>$(OUTPUT)/gitinfo.h
echo '#define APP_BUILD_DATE     "'`date +"%F %R"`'"'        >>$(OUTPUT)/gitinfo.h
echo '#define APP_BUILD_MACHINE  "'`hostname`'"'             >>$(OUTPUT)/gitinfo.h
```

So two builds of identical source differ, and a before/after comparison is meaningless unless you control for it. I verified this directly — two clean builds of the same tree gave different hashes.

For the table above I pinned all four: same commit and branch name (so `--tags` and `--all` match), same machine, and a stub `date` earlier in `PATH` returning a fixed string. With those pinned, two clean builds of unmodified source are byte identical:

```
A = 3df0709804d593be351f17f4c5df2985
B = 3df0709804d593be351f17f4c5df2985
```

That is what makes the result above meaningful rather than coincidental. Might be useful if you ever want to do this kind of check yourself.

## Toolchains

Built with your CI's toolchain, so these should be directly comparable to yours:

- **nios2** — Quartus Prime Lite **18.1.0.625**, `nios2-elf-gcc (Altera 18.1 Build 625) 5.3.0`. Same build that produced `external/u64.sof`.
- **riscv** — GCC **10.2.0**, matching the `riscv32-unknown-elf/10.2.0` in your CI logs (xPack build, symlinked to that prefix).

You are right that I have not JTAG-deployed or flashed this.

On that note, installing GCC 10.2.0 also let me verify your point on #751 directly: `target/u64ii/riscv/ultimate` builds clean on unmodified `test-merge` with GCC 10.2.0, exit 0, no errors. The error I reported there only appears with GCC 11.3.0, which is what `docker/Dockerfile` installs. So `test-merge` is fine and I was wrong; thanks for catching it.

PR #754, opened 2026-08-05
Report firmware application space in one CI table

chrisgleissner:
## Firmware application space

Every firmware image is linked into a fixed flash partition. An image that outgrows its partition cannot be flashed, so the build measures each one and refuses to publish an oversized set.

### Limits come from the flash partition tables

The permitted size of each image is the `FLASH_ID_APPL` entry of its `t_flash_address` table in `software/io/flash/w25q_flash.cc` and `software/io/flash/at45_flash.cc`. `tools/app_space.py` parses those entries at run time and holds no size constants of its own, so the budget it enforces is the same number the firmware uses to address flash. Each variant names only the partition table it is budgeted from:

| Variant | Partition table | Parsed limit |
| --- | --- | ---: |
| U2 | `flash_addresses` | `0xC6000` |
| U2+ | `flash_addresses_u2p` | `0x140000` |
| U2+L | `flash_addresses_u2pl` | `0x160000` |
| U64 | `flash_addresses_u64` | `0x170000` |
| U64E2_50T | `flash_addresses_u64ii_50t` | `0x1E0000` |
| U64E2_100T | `flash_addresses_u64ii_100t` | `0x1C0000` |

`flash_addresses` is defined in both source files, once per flash device. The parser takes the smaller of the two sizes, because the U2 image has to fit whichever device the board carries. If a source file cannot be read, or a variant's table has no `FLASH_ID_APPL` entry, the tool reports which variant and which table it could not resolve and exits 2 rather than guessing a limit.

### Measuring does not stop a build; the report does

`tools/app_space.py measure <variant>` runs inside each firmware build target, directly after the image is linked. It prints that image's row and returns success even when the image is over its limit, so the build target finishes and the next firmware still gets built and measured. An image that is over the limit prints a line naming the overrun and stating that the final report will fail the build. The one condition that stops a build target here is a missing artifact, which is a build fault rather than a size fault.

`tools/app_space.py report` is the single size gate. It reads every variant's artifact, prints one table, and exits non-zero if any image it found is over its limit. A variant with no artifact is listed with its limit and marked `n/a`, and does not fail the gate; the build target that was supposed to produce it fails on the missing artifact instead.

CI therefore builds and measures all six firmware images in one run and fails afterwards, in the dedicated `Check Firmware Application Space` step, with the complete picture in one place. This is that step's output on the current head of this branch:

```
Firmware application space (limit = FLASH_ID_APPL partition size)

  Firmware           Limit         Size         Free    Free %  Usage                 Status
  ───────────────────────────────────────────────────────────────────────────────────────────
  U2             792.0 KiB    861.5 KiB    -69.5 KiB    -8.8 %  ████████████████████  FAIL
  U2+           1280.0 KiB    980.3 KiB    299.7 KiB    23.4 %  ███████████████░░░░░  PASS
  U2+L          1408.0 KiB    966.6 KiB    441.4 KiB    31.4 %  █████████████░░░░░░░  PASS
  U64           1472.0 KiB   1157.6 KiB    314.4 KiB    21.4 %  ███████████████░░░░░  PASS
  U64E2_50T     1920.0 KiB   1131.3 KiB    788.7 KiB    41.1 %  ███████████░░░░░░░░░  PASS
  U64E2_100T    1792.0 KiB   1131.3 KiB    660.7 KiB    36.9 %  ████████████░░░░░░░░  PASS
  ───────────────────────────────────────────────────────────────────────────────────────────
  FAIL 1  PASS 5
```

Rows are coloured green, yellow, red and grey by status, and the usage bar is coloured to match. The step carries `if: always()`, so the table is produced even when an earlier step failed for an unrelated reason. It emits one `::error::` annotation per oversized image and one `::warning::` per image near its limit, and appends a markdown copy of the table to `$GITHUB_STEP_SUMMARY` so the numbers are on the run summary page without opening a log. Writing that summary needs `GITHUB_STEP_SUMMARY` passed into the container and the file it points at bind-mounted; if the write fails the tool warns and carries on rather than failing the step.

### Status thresholds

An image passes with more than 1% of its partition free. Below 1% free, including exactly at the limit, it passes with a warning. Over the limit it fails. The bar and the free percentage give a graduated reading between those points, so a firmware approaching its limit is visible on every run rather than only once it crosses the threshold.

### Local builds

`make app_space` prints the table and applies the same gate. `make all` runs it as the last line of its recipe, which is a recipe line rather than a prerequisite so that it still runs last under `make -j`. A single-variant build such as `make u64` prints that image's row as it builds. Build targets carry only the variant name, and `tools/app_space.py` resolves the artifact path, so a path is written down once.

`build-tool` drives the same top-level make targets, so builds through it are measured and gated without any change to that script.

### Output encoding and colour

The table uses box drawing and block characters when the output stream can encode them, and falls back to `-`, `#` and `.` when it cannot. This matters because the CI build container can run under a POSIX locale, where writing those characters would otherwise raise `UnicodeEncodeError` and fail the step for a reason unrelated to firmware size. Colour is enabled on a terminal, or when `GITHUB_ACTIONS=true`, or when `FORCE_COLOR` is set, and is suppressed by `NO_COLOR`.

### Verification

`tools/test_app_space.py` holds 36 tests covering the partition table parser, the status thresholds, column alignment for every row type, the usage bar including its over-limit and rounds-to-zero cases, the markdown summary, the colour and encoding decisions, both subcommands with their exit codes, the GitHub annotations, and the job summary write including its failure path.

Three of those tests are the load-bearing ones for this design. One asserts that `measure` returns 0 for an image that is over its limit, which is what lets a run finish and report on every firmware. One asserts that no firmware build target invokes a gating subcommand. One asserts that none of the parsed limits appears as a literal in `tools/app_space.py`, the `Makefile` or the workflow, so a size constant cannot creep back in alongside the parsed one.

The parser was also checked against a perturbed copy of the real sources: editing the U64 `FLASH_ID_APPL` size to `0x100000` in a temporary copy makes `load_limits` return `0x100000` for U64, confirming the value is read rather than restated.

The two exit paths were demonstrated on a real image. Padding `target/u64/nios2/ultimate/result/ultimate.app` 4096 bytes past its limit makes `measure u64` print the row with status `FAIL` and exit 0, and makes `make app_space` print the table, emit `::error::U64 is 4096 bytes over its application space limit (1511424 of 1507328 bytes)` and exit 2. Restoring the artifact, verified byte-identical by SHA-256, makes the same command exit 0.

### U2 does not fit its partition

The U2 image measures 882,220 bytes against an 811,008 byte partition, which is 71,212 bytes too large. That single row is why CI is red on this branch. It is a property of the U2 image and not of the tooling in this PR, and the image has to shrink before the branch can go green. The other five images have between 21% and 41% of their partitions free.

### Alternatives considered

Running the report at the end of every firmware target was considered, so that a single-variant build would also print the full table. It would print the table six times with partial data during `make all` and during the CI job, ahead of the complete one, so the per-target output is one row instead.

Making the size check fail its own build target was considered. It stops the run at the first oversized image, which means a single run cannot report how much room the remaining firmwares have. Deferring the failure to the report costs the remaining build time but produces the complete table, which is the point of the step.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -134,6 +134,14 @@ jobs:
           -v /opt/build-tools:/mnt/build-tools:ro \
           -v $GITHUB_WORKSPACE:/mnt/project:rw my_docker_image /bin/bash -c "cd /mnt/project && make u64ii_no_esp && mv update.ue2 update_${{ env.GIT_VERSION }}.ue2 && mv update.cfw update_${{ env.GIT_VERSION }}.cfw"
 
+      - name: Report Firmware Application Space
+        if: always()
+        run: |
+          docker run --rm -e GITHUB_ACTIONS -e GITHUB_STEP_SUMMARY \
+          -v /opt/build-tools:/mnt/build-tools:ro \
+          -v $GITHUB_STEP_SUMMARY:$GITHUB_STEP_SUMMARY:rw \
+          -v $GITHUB_WORKSPACE:/mnt/project:rw my_docker_image /bin/bash -c "cd /mnt/project && make app_space"
+
       - name: Upload User Artifacts
         uses: actions/upload-artifact@v7
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -134,7 +134,7 @@ jobs:
           -v /opt/build-tools:/mnt/build-tools:ro \
           -v $GITHUB_WORKSPACE:/mnt/project:rw my_docker_image /bin/bash -c "cd /mnt/project && make u64ii_no_esp && mv update.ue2 update_${{ env.GIT_VERSION }}.ue2 && mv update.cfw update_${{ env.GIT_VERSION }}.cfw"
 
-      - name: Report Firmware Application Space
+      - name: Check Firmware Application Space
         if: always()
         run: |
           docker run --rm -e GITHUB_ACTIONS -e GITHUB_STEP_SUMMARY \
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,54 +83,54 @@ jobs:
       - if: steps.cache-u2.outputs.cache-hit != 'true'
         name: Build U2 including FPGAs
         run: |
-          docker run --rm --net=custom_network --mac-address=4c:cc:6a:06:b3:79 \
+          docker run --rm -e GITHUB_ACTIONS --net=custom_network --mac-address=4c:cc:6a:06:b3:79 \
           -v /opt/build-tools:/mnt/build-tools:ro \
           -v $GITHUB_WORKSPACE:/mnt/project:rw my_docker_image /bin/bash -c "cd /mnt/project && make u2_rv && mv update.u2r update_${{ env.GIT_VERSION }}.u2r"
 
       - if: steps.cache-u2.outputs.cache-hit == 'true'
         name: Build U2 using cached FPGAs
         run: |
-          docker run --rm --net=custom_network --mac-address=4c:cc:6a:06:b3:79 \
+          docker run --rm -e GITHUB_ACTIONS --net=custom_network --mac-address=4c:cc:6a:06:b3:79 \
           -v /opt/build-tools:/mnt/build-tools:ro \
           -v $GITHUB_WORKSPACE:/mnt/project:rw my_docker_image /bin/bash -c "cd /mnt/project && make u2_rv_swonly && mv update.u2r update_${{ env.GIT_VERSION }}.u2r"
 
       - if: steps.cache-u2p.outputs.cache-hit != 'true'
         name: Build U2+ including FPGAs
         run: |
-          docker run --rm --net=custom_network --mac-address=4c:cc:6a:06:b3:79 \
+          docker run --rm -e GITHUB_ACTIONS --net=custom_network --mac-address=4c:cc:6a:06:b3:79 \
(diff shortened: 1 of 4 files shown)


chrisgleissner:
@GideonZ As discussed, the U2's maximum firmware size limit is now indeed breached. I updated this PR with the numbers you gave me. It fails now at https://github.com/GideonZ/1541ultimate/actions/runs/31032274439/job/92395643016#step:11:942

I also adjusted  the escalation steps:
- Less than 1% available: build warns
- Less than 0% available (i.e. too large): build fails

This PR can thus only be merged once the U2 issue is resolved. The PR itself is meant to fail its build right now as this is its main job: Increase visibility for large issues, similar to how a failing unit/e2e test should fail a build.

Thanks

chrisgleissner:
Closing in favor of #761, which merges this branch's app-space CI gate in directly and adds the fix for the U2 (original) overrun this gate reports (`FAIL 1` above). Carrying the fix on top of the gate lets the gate's own CI run confirm the fix directly, rather than having two PRs racing the same pre-existing U2-over-budget number.

#761 also fixes `.github/workflows/build.yml` triggering two builds per commit (`push` + `pull_request` firing for the same commit on any same-repo branch with an open PR) — spotted while working the same CI step.

No functional disagreement with this PR's design (the parser, the report/measure split, the pass/warn/fail thresholds, and the test suite are all carried forward unchanged) — this is purely to avoid running the gate on two branches in parallel.

Issue #755, opened 2026-08-06
Ultimate64 Elite-II PDsid type setting is not being saved.


Retro1968:
Hello Gideon,
as of yesterday, I have installed two PDsids in my Ultimate64 Elite-II board (with FW 3.14d).

After downloading and extracting the latest PDsid firmware zip file from here:
https://uni64.com/shop/index.php?system=1&m=70&item_id=7536

I successfully flashed firmware V1.88 onto both PDsids using the `pdsid_fw_188.prg` program.

While it is possible to configure the PDsids from a Windows PC using the `PDsid_Ultimate_Tool_Win` (also available at the download link above) and then manually save the configuration file on the Ultimate64 Elite-II, the previously selected PDsid type (6581/8580) is unfortunately not saved.

The same issue occurs when changing the emulation mode to 6581 or 8580 in the Ultimate64 Elite-II menu and saving the settings. This is a shame, as the menu allows you to assign specific types to each PDsid - for example, Socket 1 to 6581 and Socket 2 to 8580.

To date (August 6, 2026), the PDsid type setting (6581 or 8580) can only be permanently changed using the "pdsid_detect.prg" program by pressing the space bar (followed by F1 to reload). However, this method only allows the SID type to be changed for both PDsids simultaneously (provided two PDsids are installed in the Ultimate64 Elite-II).

Kind Regards
Retro1968

Retro1968:
We discuss this issue in Forum 64:
https://www.forum64.de/index.php?thread/164986-ultimate64-elite-ii-pdsid-typeneinstellung-6581-8580-wird-nicht-gespeichert/&postID=2462097#post2462110

Retro1968:
The Saving of the  PDsid-Type (6581/8580) Settings in this Menue on the Ultimate64 Elite-II (FW3.14d) does not work:

<img width="4096" height="1864" alt="Image" src="https://github.com/user-attachments/assets/8c934f4d-fae7-4453-a1dd-bfa3f670beaf" />

<img width="4096" height="1864" alt="Image" src="https://github.com/user-attachments/assets/da23df08-0507-421b-b558-12f6ba045896" />

chrisgleissner:
Thanks for the detailed report, @Retro1968. Much appreciated.

A few follow-up questions:

- Did this ever work with your **current PDsid firmware**?
- Did it ever work with an **older PDsid firmware** or an **older PDsid hardware version** if there is one?
[...]

PR #757, opened 2026-08-09
Fix reversed 6581/8580 selection for PDsid

barryw:
Fixes the reversed 6581/8580 selection reported in #755.

## The bug

`Emulation Mode` for PDsid is a `CFG_TYPE_ENUM` over `{ "6581", "8580" }`, so `it->getValue()` is **0 or 1**:

https://github.com/GideonZ/1541ultimate/blob/test-merge/software/u64/sid_device_pdsid.cc#L15-L19

The change hook passed that value straight into `SetSidType()`:

```c
obj->SetSidType(it->getValue());
```

But `SidDevice::SetSidType()` takes **1 = 6581, 2 = 8580**. That is the convention every other implementation follows — `sid_device_swinsid.cc`, `sid_device_armsid.cc` and `sid_device_fpgasid.cc` all open with the same line:

```c
type--; // receive 1 = 6581, 2 = 8580 => map to 0 and 1
```

and it is the same encoding `U64Config::MapSid()` passes down through `t_sid_definition::sidType` (`{ "None", "6581", "8580", "Either" }`).

So the hook is off by one, and the two options are swapped:

| Menu selection | `getValue()` | `SetSidType()` | `base[31]` | Device becomes |
|---|---|---|---|---|
| 6581 | 0 | 0 | `0x01` | **8580** |
| 8580 | 1 | 1 | `0x00` | **6581** |

## Why it also looked like the setting was not being saved

Reopening the menu calls `at_open_config()`, which reads the mode back from the device and displays it. So the menu was showing the chip the device had actually been switched to — the opposite of the one just selected. From the outside that is indistinguishable from the selection being discarded, which is how it was originally reported.

That is consistent with the reporter's own findings in #755: he independently arrived at the workaround "select 6581 to get 8580, and vice versa", and separately answered "no" to whether the selection survives leaving and reopening the menu. One bug, both symptoms.

## The fix

```c
obj->SetSidType(it->getValue() + 1);
```

`SetSidType()` and `at_open_config()` are already consistent with each other and with the register encoding (`0x00` = 6581, `0x01` = 8580) — `SetSidType(1)` writes `0x00`, and `at_open_config()` maps `base[31] & 1 == 0` back to enum index 0, `"6581"`. Only the call site disagreed, so only the call site is changed. `SetSidType()` semantics are untouched.

## Verification

`sid_device_pdsid.cc` is referenced by exactly two target makefiles — `target/u64/nios2/ultimate` and `target/u64ii/riscv/ultimate` — so those are the only two builds this can affect. Both were built from a clean tree and completed with exit status 0:

| Target | Toolchain | Result |
|---|---|---|
| `make u64` | Quartus Prime Lite 18.1.0.625, `nios2-elf-gcc 5.3.0` | exit 0 |
| `make u64ii` | `riscv32-unknown-elf-gcc 10.2.0`, ESP-IDF v5.3.1 | exit 0 |

Those are the CI toolchain versions, so the result should be directly comparable to yours.

I do **not** have PDsid hardware, so I have not confirmed this on a device. The reporter in #755 does, and has been very willing to test. Happy to have him verify before this is merged.

## Not addressed here

Two further problems from #755 are real but are separate from this one, and I would rather not fold them into this change:

1. **PDsid settings are absent from saved `.cfg` files.** `ConfigIO::S_write_to_file()` skips any store without a flash page, and `PdSidConfig` is constructed with `ConfigStore(NULL, ...)`. I checked the `.cfg` files the reporter attached to #755 — there is no `[SID Socket n: PDsid]` section in any of them. This guard is generic and also covers SidKick, ArmSid and the RTC stores, so changing it is a decision about intended behaviour rather than a bug fix.
2. **`PdSidConfig::effectuate()` is empty**, so even with the section present, loading a `.cfg` would not push the mode to the device.

I will raise both on #755 rather than guess at the intent here.

One thing noticed while tracing this, unrelated to #755 and not changed: `SidDeviceSidKick::SetSidType()` does not do the `type--` step the other implementations do, so the auto-mapping path in `U64Config::MapSid()` would select the wrong chip for a SidKick too. I have not tested that and it is a different code path — mentioning it in case it is useful.

### DIFF
--- a/software/u64/sid_device_pdsid.cc
+++ b/software/u64/sid_device_pdsid.cc
@@ -68,7 +68,7 @@ SidDevicePdSid::~SidDevicePdSid()
 int SidDevicePdSid::PdSidConfig:: S_cfg_pdsid_type(ConfigItem *it)
 {
     SidDevicePdSid *obj = (SidDevicePdSid *)it->store->get_hook_object();
-    obj->SetSidType(it->getValue());
+    obj->SetSidType(it->getValue() + 1); // menu enum is 0 = 6581, 1 = 8580; SetSidType takes 1 = 6581, 2 = 8580
     return 0;
 }
 


chrisgleissner:
@barryw Thanks for the fix! That's great news. 

Also thanks for the detailed PR description and for having kept this PR very targeted whilst describing issues you encountered along the way.

@GideonZ I see that the build failed, but that is unrelated to the PR's change. Either another Github issue like we had a few days ago or maybe some runner degradation. A rerun just now did not fix the issue.

barryw:
Thanks @chrisgleissner. A bit more detail on the CI failure in case it helps narrow it down — I don't think this one is GitHub-side.

Both attempts failed in the same place, a few seconds in, before checkout finished and long before anything was compiled:

```
Checkout repository
##[error]error: could not lock config file /home/github-runner/_work/_temp/<uuid>/.gitconfig: No such file or directory
##[error]The process '/usr/bin/git' failed with exit code 255
```

Attempt 1 also failed earlier than that, during `Set up job`:

```
##[error]The file '/home/github-runner/_diag/pages/acdfa68f-..._1.log' already exists.
```

So two separate runner-owned directories — `_diag/pages/` and `_work/_temp/` — both refused writes in the same job. `actions/checkout` never got its per-job temp directory created, which is why git then could not write the config file it wanted.

That combination points at the runner's own filesystem rather than at GitHub: most likely `ryzen` is out of disk, or `/home/github-runner` has gone read-only. `df -h` on the runner, and clearing `_work/_temp` and `_diag`, would be the first things I would check.

For completeness: the failures on `agent/firmware-app-space` a few days ago were a different problem — those were real build failures (`U2 is 71212 bytes over its application space limit`), not this.

barryw:
@chrisgleissner — thanks, but before this gets merged I want to be straight about how far the evidence actually goes, because I think I stated it too confidently above.

There are two claims in this PR and they are not equally solid.

**Solid: the call site is wrong.** This needs no hardware. `getValue()` returns 0 or 1, `SetSidType()` takes 1 or 2, and — the part that settles it — `SetSidType()` and `at_open_config()` currently contradict each other, so the menu cannot round-trip its own selection no matter what the device does. Something here is definitely broken.

**Not solid: which direction to correct it.** My change assumes the PDsid register uses `0x00` = 6581 and `0x01` = 8580. My only sources for that are the two code paths in `sid_device_pdsid.cc` itself, and the commit that introduced them is labelled *"Add support for PDsid (untested)"*. There is no PDsid protocol documentation in the repo, and I have no PDsid to check against.

If that encoding is actually inverted, this patch makes the menu self-consistent while silently mislabelling both options — the device would end up on the opposite chip from the one displayed. That would be harder to spot than the current bug, not easier.

I should also flag that the reporter's evidence in #755 cannot distinguish the two cases, even though it looks like it can. `at_open_config()` reads `base[31] & 1` — the same bit either way — so "pick 6581, reopen, see 8580" happens identically under both encodings. The workaround he found is real and useful, but it is circular as proof of direction.

What makes me think the direction is right, roughly: every other SID device in this codebase puts 6581 at the lower value (`SidKick #define SID_TYPE_6581 0`, FpgaSID `0x40` vs `0xFF`, ArmSid and SwinSid `'6'` vs `'8'`), and the two PDsid code paths were written consistently with each other, so the encoding presumably came from somewhere real. That is a strong prior, not proof.

I have posted a test on #755 that settles it in about a minute on stock 3.14d, with no beta firmware and nothing to flash. Happy to have this held until that comes back — and if it comes back the other way, the correct fix is in `at_open_config()` instead, and I will redo it.

chrisgleissner:
@barryw Thanks for clarifying.

I found an independent PDsid API test from April 2026 which uses the same `$D41D/$D41E/$D41F` + `P`/`D` handshake as `sid_device_pdsid.cc` and explicitly documents `0 = 6581`, `1 = 8580`:

https://oldmachinery.blogspot.com/2026/04/pdsid.html

That strongly supports the direction used by this PR. One caveat: the test predates a later PDsid firmware release mentioning an “API Fix”, so testing current v1.88 is still the cleanest final confirmation.

barryw:
Update: the uncertainty I raised above is resolved, and it went the right way.

@Retro1968 ran the check on #755 ([comment](https://github.com/GideonZ/1541ultimate/issues/755#issuecomment-5232415746)). He set the mode with `pdsid_detect.prg`, power cycled, and read the menu without touching it:

| Set with `pdsid_detect.prg` | Menu shows |
|---|---|
| 6581 | 6581 |
| 8580 | 8580 |

So the readback in `at_open_config()` is correct in both directions, which pins the register encoding to `0x00` = 6581 and `0x01` = 8580 — exactly what `SetSidType()` already assumes. The call site was the only thing wrong, and `+ 1` is the correct correction.

The power cycle is what makes it conclusive: `PdSidConfig::read()` and `effectuate()` are both empty, so the firmware writes nothing to the PDsid at startup. What the menu displayed was the device's own state, which is the independent reference I could not obtain myself.

No change to the patch — it stands as originally submitted. Withdrawing my request to hold it.

Worth repeating one thing from the description, since the issue title is about saving and this PR is not: this fixes the reversed selection only. PDsid settings still do not appear in saved `.cfg` files, and loading a `.cfg` would not apply the mode to the device. I have set both out on #755 for @GideonZ, as the first one depends on a decision about page-less config stores that also affects ArmSid, SidKick and the RTC.

PR #761, opened 2026-08-10
Drop machine-code monitor and FTP client from U2 (original) build

chrisgleissner:
## Summary

The U2 (original) application partition is a fixed 811,008-byte (`0xC6000`) flash region, the same on both the AT45 and W25Q flash variants of the original board, so it is a hardware limit, not a config choice. Recent additions (the interactive machine-code monitor, FTP client/server support, and other features) pushed the U2 build past that limit.

This PR removes two U2-only features that are dead-code-safe to drop, merges in #754's firmware application-space CI gate so the fix gets checked against a real gate instead of a projection, and fixes a CI workflow bug that built every commit twice.

## Remove the machine-code monitor and FTP client from U2 (original)

The interactive 6502 machine-code monitor/debugger (`software/monitor/*`) has one call site, a weak symbol (`get_machine_monitor_task`, `software/monitor/monitor_init.h:13`) guarded with `if (get_machine_monitor_task) { ... }` in `tree_browser.cc`. Removing the translation units from the U2 `Makefile` leaves that check false and the rest of the UI unaffected.

The FTP client that lets the U2 file manager browse a remote FTP server as a filesystem (`software/network/ftp_client.cc`, `software/filesystem/filesystem_ftp.cc`) is separate from the FTP *server* daemon (`ftpd.cc`), which stays in the U2 build. The client self-registers into the file manager root via a static `InitFunction`; no other file references `FileSystemFTP`, `FTPClient`, or that registration function.

Only `target/u2/riscv/ultimate/Makefile` changes for this part (10 `SRCS_CC` lines removed). Every other hardware target (U2+, U2+L, U64, U64-II) keeps its own independent `SRCS_CC` list and is untouched.

**Measurements.** Rebuilt locally with the Docker image tag the self-hosted CI runner uses (`my_docker_image:latest`), `make -C target/u2/riscv/ultimate`, comparing `riscv32-unknown-elf-size` on the resulting `.elf` before and after:

| Build | `ultimate.bin` |
|---|---:|
| `test-merge` tip, unmodified | 904,788 |
| − machine-code monitor only | 841,756 |
| − monitor + FTP client (this PR) | 824,976 |

Both link cleanly with no undefined-symbol errors. Local absolute numbers run larger than what CI reports for the same source, a known toolchain drift between the locally cached image and whatever the self-hosted runner actually pulls (the Docker tag is not a content pin). What transfers reliably across that drift is the delta from a controlled single-variable rebuild: applying the measured 79,812-byte delta to the most recent CI-reported number for a size-equivalent commit (882,220 bytes) projects to roughly 802,408 bytes, about 1% under the 811,008-byte partition limit. The merged-in gate below gives a real, non-projected number instead of relying on that estimate.

## Firmware application-space CI gate (merged from #754)

Every firmware image links into a fixed flash partition; an image that outgrows its partition cannot be flashed. `tools/app_space.py` measures each build's image against its partition and fails CI on an oversized one. It holds no size constants of its own: the limit for each variant is parsed at run time from the `FLASH_ID_APPL` entry of the `t_flash_address` tables in `software/io/flash/w25q_flash.cc` and `software/io/flash/at45_flash.cc`, the same numbers the firmware itself uses to address flash, so the budget can't drift out of sync with the hardware.

`tools/app_space.py measure <variant>` runs inside each firmware build target right after linking. It prints that image's row and always exits 0, even when the image is over its limit, so every remaining firmware still gets built and measured in the same run. `tools/app_space.py report` is the single gate: it reads every variant's artifact, prints one table, and exits non-zero if any image is over its limit. This runs as a dedicated `Check Firmware Application Space` CI step with `if: always()`, so the table appears even if an earlier step failed for an unrelated reason. It also emits one `::error::` annotation per oversized image and one `::warning::` per image near its limit, and appends a markdown copy of the table to the job summary.

An image passes with more than 1% of its partition free, passes with a warning below 1% free (including exactly at the limit), and fails only when it is actually over. `make app_space` runs the same check locally, and `make all`/`build-tool` run it automatically as the last build step.

`tools/test_app_space.py` holds 36 tests covering the partition parser, the status thresholds, the table rendering, both subcommands, the GitHub annotations, and the job summary write. Three are load-bearing for the design: one confirms `measure` always exits 0 so a run finishes and reports on every firmware, one confirms no build target invokes a gating subcommand directly, and one confirms none of the parsed size limits is duplicated as a literal anywhere else in the tree, so a stale constant can't creep back in beside the parsed one.

## CI: build once per PR, once per merge

`.github/workflows/build.yml` declared `on: [push, pull_request]` unconditionally, which fires both a `push` build and a `pull_request` (synchronize) build for the same commit whenever a branch lives in this repo and has an open PR — confirmed directly on this branch, where an early commit triggered two full builds. Restricting `push` to specific branches was considered and rejected, since people often push work-in-progress branches with no PR open yet, and those still need to build.

The fix keeps `on: [push, pull_request]` unrestricted, so every branch still builds on push whether or not it has a PR. The `build` job instead skips itself on a `pull_request` event when the PR head is this same repo, since `push` already built that exact commit; it still runs on `pull_request` when the head is a fork, since a fork's push never reaches this repo and `pull_request` is the only signal for those. Verified on this branch: the latest commit produced one `push` run that built, and the matching `pull_request` run for the same commit completed as `skipped`.

## Test plan

- [x] `make -C target/u2/riscv/ultimate` links cleanly for U2 (original) with both features removed, no undefined symbols.
- [x] Only `target/u2/riscv/ultimate/Makefile` changed for the size fix; U2+/U2+L/U64/U64-II Makefiles are untouched.
- [x] `python3 -m pytest tools/test_app_space.py`, 36/36 pass after the merge.
- [x] Confirmed the build-once fix on this branch's own CI runs: one `push` run built, the paired `pull_request` run for the same commit shows `skipped`.
- [ ] Real CI confirmation of the final U2 image size against the merged-in gate, this PR's own `Check Firmware Application Space` step is the first place that gets checked against the actual self-hosted-runner toolchain rather than the delta-transfer projection above.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -134,6 +134,14 @@ jobs:
           -v /opt/build-tools:/mnt/build-tools:ro \
           -v $GITHUB_WORKSPACE:/mnt/project:rw my_docker_image /bin/bash -c "cd /mnt/project && make u64ii_no_esp && mv update.ue2 update_${{ env.GIT_VERSION }}.ue2 && mv update.cfw update_${{ env.GIT_VERSION }}.cfw"
 
+      - name: Report Firmware Application Space
+        if: always()
+        run: |
+          docker run --rm -e GITHUB_ACTIONS -e GITHUB_STEP_SUMMARY \
+          -v /opt/build-tools:/mnt/build-tools:ro \
+          -v $GITHUB_STEP_SUMMARY:$GITHUB_STEP_SUMMARY:rw \
+          -v $GITHUB_WORKSPACE:/mnt/project:rw my_docker_image /bin/bash -c "cd /mnt/project && make app_space"
+
       - name: Upload User Artifacts
         uses: actions/upload-artifact@v7
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,12 +1,14 @@
 name: Build
 
-on:
-  push:
-    branches: [master, test-merge]
-  pull_request:
+on: [push, pull_request]
 
 jobs:
   build:
+    # A same-repo branch's push already builds via the push event above; only run the
+    # pull_request event when the PR head is a fork, whose pushes never reach this repo
+    # and so would otherwise never build at all. This is what avoids building the same
+    # commit twice, without restricting which branches build on push.
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     runs-on: self-hosted
     steps:
       - name: Checkout repository
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -134,7 +134,7 @@ jobs:
           -v /opt/build-tools:/mnt/build-tools:ro \
           -v $GITHUB_WORKSPACE:/mnt/project:rw my_docker_image /bin/bash -c "cd /mnt/project && make u64ii_no_esp && mv update.ue2 update_${{ env.GIT_VERSION }}.ue2 && mv update.cfw update_${{ env.GIT_VERSION }}.cfw"
 
-      - name: Report Firmware Application Space
+      - name: Check Firmware Application Space
         if: always()
         run: |
           docker run --rm -e GITHUB_ACTIONS -e GITHUB_STEP_SUMMARY \
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,9 @@
 name: Build
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [master, test-merge]
+  pull_request:
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


chrisgleissner:
Hi @GideonZ,

As discussed, I removed MCM and the FTP Client from the original U2 to keep the firmware within its size limits.

This also improves visibility of future size-limit breaches, so they are clearly flagged in CI before we merge changes that would negatively affect a device.

One detail worth mentioning: the `GITHUB_ACTIONS` environment variable passed into the Docker builds is intentional. It lets the size checks distinguish CI output from locally redirected/batched builds. This gives us colored warnings in GitHub Actions while keeping ANSI color codes out of local non-interactive output.

Thanks for the review and merge.

Christian

PR #765, opened 2026-08-10
Apply only groups loaded from CFG files

chrisgleissner:
## Overview

Loading a CFG file applies settings only from stores that successfully loaded an item from that file. A one-group file therefore cannot effectuate SID, network, drive, UI, or other unrelated settings.

## Implementation

`ConfigIO::S_read_from_file()` records each distinct store that successfully loads an item. `FileTypeCfg` effectuates only those stores. The existing effectuation and flash dirty flags retain their ownership, so unrelated pending stores are neither applied nor cleared. An invalid section clears the active store.

## Regression coverage

`cfg-single-group` is registered in `run-tests`. It loads a browser-selected CFG containing only `[Audio Mixer]`, saves the device debug log, and asserts that exactly that store is effectuated. It restores the modified setting and removes its fixtures. Navigation reuses shared quick-seek and overlay routing.

### Hardware red/green

```sh
python3 tests/e2e/filemanager/cfg_single_group_test.py -H u64 --mode overlay
```

```text
# Red
cfg_single_group_test: FAIL (Expected ["Audio Mixer"] after CFG load, got unrelated stores...)

# Green
[01] load a one-group CFG file through the browser ... OK
[02] only the CFG group is considered after loading ... OK
cfg_single_group_test: OK (2 checks, 6.443s)
```

A host unit test was not added: ConfigIO has no host harness, and a faithful test requires flash, file-manager, UI, stream-log, and config-store scaffolding. The hardware regression covers the real parser-to-effectuation path with less test-only code.

## Validation

- `BUILD_TOOL_ALLOW_PARTIAL=1 ./build-tool -s u64` — pass
- JTAG deployment of the PR-head U64 image `update_6fd9cdc9.u64` — pass
- `BUILD_TOOL_ALLOW_PARTIAL=1 ./build-tool -s u64 u64ii u2 u2pl` — U64/U64-II pass; U2/U2+L applications compile and their updater builds stop at known absent software-only FPGA artifacts (`rv700dd.chk`, `u2p_ecp5_impl1.bit`).

### Full E2E gate

Executed against JTAG-deployed `agent/cfg-single-group-upstream` commit `6fd9cdc9`:

```sh
./run-tests -H u64
```

Result: **19/19 suites passed in 698 seconds**, including `cfg-single-group`, `machine-code-monitor`, and `assembly64`.

[Full 54,553-byte transcript](https://gist.github.com/chrisgleissner/4f23105878dd1d6fdbef08d72f4be284)

### DIFF
--- a/run-tests
+++ b/run-tests
@@ -129,6 +129,8 @@ SUITES: Sequence[Suite] = (
           "-H @HOST@ -p @PASS@ --mode @MODE@"),
     Suite("e2e", "temp-auto-cleanup", "tests/e2e/filemanager/temp_auto_cleanup_test.py",
           "-H @HOST@ -p @PASS@ --mode @MODE@"),
+    Suite("e2e", "cfg-single-group", "tests/e2e/filemanager/cfg_single_group_test.py",
+          "-H @HOST@ -p @PASS@ -t @TIMEOUT@ --mode @MODE@"),
     Suite("e2e", "printer", "tests/e2e/io/printer/printer_test.py", "-H @HOST@ -p @PASS@"),
     # Drives the UCI control and SoftIEC targets over $DF1B-$DF1F; reproduces #740.
     # A #740 regression leaves the interface stuck in Command Busy for every target
--- a/software/filetypes/filetype_cfg.cc
+++ b/software/filetypes/filetype_cfg.cc
@@ -79,7 +79,8 @@ SubsysResultCode_e FileTypeCfg :: execute(SubsysCommand *cmd)
     StreamTextLog log(8192);
 
     if(file) {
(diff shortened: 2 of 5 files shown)


chrisgleissner:
Hi @GideonZ,

here is the fix for the bug we discussed today.

Thanks for review / merge.

Best
Christian

GideonZ:
@chrisgleissner I merged your PR as is, because you have proven it to work with your red/green test. However, as you can see that there is a function called "need_effectuate". This simply should return false for stores that are not loaded. If this would be fixed, you wouldn't need the IndexedList of loaded stores, am I right?

PR #767, opened 2026-08-11
Stabilize the machine-code monitor and make U2 a runner target

chrisgleissner:
## Summary

Three things, in one branch because the second is what proved the first and the
third is what makes either diagnosable after the fact:

1. **Machine-code monitor**: memory editing is correct on an Ultimate II+L, a
   typed range means the same thing in every command, `C=+O` opens the monitor
   from anywhere in the menu, and the Help screen is one readable page.
2. **Test harness**: the Ultimate II+L and the C64 Ultimate are runner targets
   the standard gate drives directly, alongside the Ultimate 64.
3. **Observability**: a run keeps enough about itself to be diagnosed without
   the device, as one Markdown report over JSONL, console logs, captured
   screens, the devices' own log and an optional recording.

The firmware changes are the monitor's, the global shortcut, and three items in
the syslog path. Everything else is the harness.

## Monitor: memory editing on an Ultimate II+L

A hex edit at an ordinary main-RAM address did not reach C64 memory. While the
cartridge's freezer holds the machine it banks its own scratch RAM over part of
the address range, and the gap the frozen-Ultimax handling left open started at
`$8000`, above everything a test or a user edits. It starts at `$1000`, which
is where the freezer's window actually begins. Three call sites carried the
same constant: `peek()`, `poke()` and `dma_transfer_frozen()`.

Proven by editing through the monitor, leaving the monitor completely, and
reading the byte back through the device afterwards. While the freezer is up,
the device's own view of that range is the freezer's, not the C64's, so it is
not an oracle.

## Monitor: a range includes both of its ends

`Fill`, `Hunt` and `Save` read `C000-CFFF` as including `$CFFF`. `Transfer` and
`Compare` did not: `T C000-CFFF,2000` copied 4095 bytes and left the last one
behind, and `C C000-CFFF,2000` never compared it. Both now use the same rule,
and `start == end` is a one-byte range rather than an error. The lengths are
32-bit, so `0000-FFFF` is the whole 65536 bytes.

## Monitor: `C=+O` is a global shortcut

`C=+O` was a case in `TreeBrowser::handle_key`, so the file browser opened the
monitor and no other screen did: from the task menu, the settings screens and
the system-information screen the key was read and nothing happened. It is
handled once now, in `UserInterface::keymapper`, the one function every UI
context passes its keys through, so no screen has to implement it.

A popup, a string box, an editor and the monitor itself own the screen and are
polled outside the UI object stack, so there is nothing there that could
restore what the monitor drew over. Those bracket themselves and the shortcut
is inert while one is up, which is also what keeps `C=+O` inside the monitor
meaning what it already meant.

## Monitor: the Help screen

One page, two grids, aligned on fixed columns, keys in the brighter colour and
descriptions subdued. `D` is reserved rather than a second name for the
Assembly view; `A` is the Assembly view. `U` is `Undoc/Case`, `O` is
`CPU Bank`, `SH+O` is the VIC bank.

## Harness: three machines, one gate

A target is a host, or `cartridge@computer` for a cartridge under test in the
computer that supplies its C64 keyboard and video, written `u2@c64u`. One
module answers where every surface of a device is; another identifies the
machine once from `/v1/info` and answers what it is and what its firmware line
lacks, so a check that cannot run on a machine skips with the reason rather
than failing.

## Observability

`-o DIR` keeps the whole run: every check, suite, health sweep and device
request as JSONL, each suite's console log, the screens the suites read, the
state captured when a check failed, and with `--syslog` the devices' own log.
`python3 tools/e2e_report.py DIR` turns that into `runs/index.md` with no
device: a greppable status line, then every failing check with the screen, the
log tail and the command that runs it again. `--record` adds a video of the
harness's screen beside the device's, subtitled per check.

Nothing here may change a verdict or an exit status. An output directory that
cannot be created is reported once and the run continues; a log file that
cannot be written is found when the collector starts rather than on the first
datagram; and an interval a component could not observe anything is recorded as
a `gap` with a start and an end, or with no end when it was still open, and
shown on the timeline beside the suite that was running.

124 device-free cases cover it, against a loopback device double. They run as
`make observability_test` and in the build CI.

## CLI

`-j/--jsonl-dir` is now `-o/--output-dir`, with no alias: the directory holds
every artifact a run keeps, not only JSONL. `./run-tests --help` is grouped by
what a reader is looking for rather than by the order the options were added.
`--color auto|always|never` is one implementation for every program here, and
verdicts are coloured in GitHub Actions, which renders the escapes but is not a
terminal.

## Validation

Every device-free check runs as `make observability_test`: 124 cases, green in
under 30s on the build CI's own image, against a loopback device double.
`runner_policy_test` and `check_transport_usage` are green beside it, and the
monitor's host tests cover the lexical space of every command prompt.

On hardware, the complete default suite on each machine, sequentially and then
with all three driven from one invocation. Every run kept and rendered to
`index.md`.

The observability layer was audited against real recorded runs rather than only
through its own tests: the output directory's records, console logs, screens,
captures and device log cross-checked against each other and against the
console; the video's duration, frame count, chapters and subtitles cross-checked
against the check records; the burned-in timecode compared pixel for pixel with
what the report tells a reader to seek to; and every still reproduced pixel for
pixel by the video frame at its moment.

That audit found seven defects, each fixed with a test that fails against the
previous behaviour:

- the timecode burned into every frame was late by the title card's length, so
  a screenshot did not seek back to itself;
- a check that reported its own verdict was recorded twice, the second record
  contradicting the first, and a suite of 39 checks reported 43;
- three suites left no verdict in their own records, which is what a killed
  suite looks like;
- a failing menu suite showed only the C64 screen behind it, while the window
  that caused the failure was in the spool and nowhere in the report;
- a non-empty `syslog-unmapped.txt` went unexplained;
- three record kinds were written and documented nowhere;
- the observability suite itself failed inside the gate, because it inherited
  the gate's `E2E_SYSLOG_OWNED` and its scripted runs then started no
  collector.

The last of those only appears when several targets are driven at once, which
is the case the concurrent run exists to cover.

Recording health on hardware: `frames_shed` 0, `frames_padded` 0,
`packets_dropped` 0, `packets_malformed` 0, `frames_reordered` 0, the audio
track carrying signal rather than silence, and the video timing detected rather
than assumed, `ntsc` on the Ultimate 64 and `pal` on the U2+L in its C64
Ultimate. `frames_lost` is 0 on a single-suite recording and 12729 of 69535
across a full 23-suite sweep, because several suites take the video stream for
their own captures and the recorder receives nothing while they hold it. The
capture record carries that as the re-arm count, which is what tells a reader
the run fought for the stream. The file itself stays complete: 1064.8s at 10
frames a second is 10648 frames, with one chapter per suite run.

Enabling the observability features does not change a verdict: the monitor
suite on `u64` reaches the same result with and without `--record` and
`--syslog`.

## Firmware changes outside the monitor

Nine files under `software/` are touched outside `software/monitor/` and the
test trees. Each is listed with what it does and why the change belongs there
rather than in the monitor.

**`userinterface/userinterface.{cc,h}`, +122 and +18 lines.** `C=+O` is
dispatched in `keymapper`, the one function every UI context passes its keys
through, and a modal depth counter keeps the shortcut inert while a popup,
string box, editor or the monitor itself owns the screen. This is the change
that makes the shortcut global; no smaller placement reaches every context. The
dispatch is compiled out where the application has no monitor to open, which is
what keeps the updater and recovery images linking.

**`userinterface/tree_browser.cc`, net removal.** Deletes the local
`KEY_CTRL_O` case that the global dispatch replaces, and leaves a comment
saying where the key is handled now.

**`userinterface/ui_elements.{cc,h}`, +54 and +26 lines.** Adds
`UIStringEditPolicy`, an optional per-field input policy for string boxes:
which candidate buffers may be typed, how a printable key is rewritten before
insertion, and one extra cancel key. Every member is optional and a field that
supplies no policy behaves as before, so the many existing string fields are
unaffected. The monitor's address and hex fields are the only user. The
alternative was a second string-box implementation inside the monitor, which
would have duplicated the editing, drawing and cancel handling.

**`io/c64/c64.{cc,h}`, +41 and +8 lines.** Three things, all about the machine
rather than the monitor. The frozen Ultimax bus gap starts at `$1000` rather
than `$8000`, which is the memory-editing fix above. `begin_stopped_session` and
`end_stopped_session` move here from `U64Machine` so that a cartridge in a host
computer gets them too; `u64/u64_machine.{cc,h}` is the matching deletion, so
this pair is a move, not new code. An accessor pair reads and writes the frozen
CIA2 port A value, which is where the VIC bank lives while the machine is
frozen.

**`network/syslog.{cc,h}`, +155 and +28 lines.** The device's own log is one of
the run's evidence streams, and it was previously unable to say whether it had
lost anything. It now counts datagrams the stack refused and buffer overflows,
exposes a flush, takes the safe section around the buffer write so a
half-written line cannot be forwarded, and frees its 16 KB buffer when nothing
is configured to drain it. The write path stays a single bounded store under
the safe section.

**`api/routes.cc`, +7 lines.** `/v1/info` reports `syslog_failed_sends` and
`syslog_overflows`. A device logging where nothing listens is silent and
harmless to a run, so without these counters a lossy link and a quiet device
look identical from the host. The run's health sweep reads them.

**`system/assert.c`, +23 lines.** An assertion failure used to print inside the
critical section, where the syslog task can no longer run, so the one message
worth having never left the machine. It prints and flushes before interrupts go
off. `syslog_flush` is a weak symbol because `assert.c` links into applications
that have no syslog, and there the null pointer is the correct answer. The cost
is stated in the comment: printing outside the critical section can interleave
with another task's output.

**`application/ultimate/ultimate.cc`, +25 lines.** Defines the `syslog_flush`
that `assert.c` calls, and releases the forwarding buffer when no log target is
configured, which matters most on the U2 where the heap is tightest.

A reviewer went back over these looking for anything that could be dropped. The
syslog work lost a configurable silence threshold, a boot-marker scan, and two
collector helpers with no caller. What stayed is what a test or the report
depends on, and the full gate was re-run on hardware afterwards.

## Documentation

`doc/machine_code_monitor.md` states the one range rule and where `C=+O` works.
`tests/e2e/doc/self-hosted-runner.md` is new: what a machine carrying the `e2e`
label has to provide, why the syslog setting is device boot-time state rather
than run configuration, and how the firmware under test gets onto the devices
before a run, by JTAG (which writes no flash and works on a wedged device) or
through the device's own updater (which writes flash and can leave a device
that does not boot).

## Known limitations

**The Ultimate II (non-plus) build is over its application space** by about
17 KiB. It predates this branch and the app-space check reports it rather than
hiding it. Nothing here reduces the image.

**The C64 Ultimate runs a separate firmware line** that does not have the
monitor revision this suite drives, so `./run-tests c64u -s machine-code-monitor`
reports SKIP with the machine and version in the reason. That is the accurate
result, not a gap in the gate.

**The C64 Ultimate stops answering after sustained driving, and only mains
power brings it back.** Twice during this gate: once its UI task blocked in a
modal that RUN/STOP would not release while every listener still answered, and
once it stopped answering REST altogether, both after hours of continuous
driving. `machine:reboot` returns 200 and clears neither. The runner does the
right thing with it, abandoning the run rather than failing every suite left
for the same reason, and it is why the concurrent run's third target needed a
restart the other two did not.

**That machine also answers about ten times slower than the Ultimate 64.** The
health sweep's `machine:readmem` of `$D012` takes a median of 271, 274 and
276ms across three complete `u2@c64u` sweeps, against 26ms on the Ultimate 64,
and a different timing-sensitive suite failed on each of those sweeps. Two
symptoms of the same path were characterised rather than papered over: an
injected keystroke is lost occasionally, caught on video as `J C003` reaching
the monitor as `$0C03`, which `MonitorSession.goto` now fails at rather than
three assertions later; and a byte written through the monitor while frozen
occasionally does not land, an Assembly-view edit committing its opcode and not
its operand, seen once in a sweep and never in 25 repetitions of the same check
in isolation.

Neither is fixed here. Keys reach the cartridge by being tapped into that
host's keyboard matrix, and nothing on this branch touches that host's
firmware.

**The E2E gate workflow has still never run.** It needs a self-hosted runner
carrying the `e2e` label on the device LAN, which is a decision about a machine
rather than a repository change. Everything it calls was run by hand for this
branch; `tests/e2e/doc/self-hosted-runner.md` is what somebody standing that
machine up needs.


chrisgleissner:
Closing this in favour of two PRs. The work here was two separable things on one branch, and reviewing them together was harder than it needed to be.

- #784 `feat/e2e-observability`: everything that keeps and reads a run. The run directory, the report generator, the recorder, the device-log collector, the shared target and machine layer, both GitHub workflows, and the five firmware files in the syslog path. Based on `test-merge`.
- #785 `fix/mcm-stability`: the machine-code monitor, its host tests, its E2E suite and its manual, plus the `c64.cc` freeze-range fix and the global `C=+O` shortcut. Based on #784, because the monitor suite is driven through the harness that lands there. Merge #784 first and GitHub retargets it.

The split was done by content rather than by commit, since the commits here interleave. Together the two branches reproduce this branch's tree exactly, apart from the three `test-merge` commits that landed after this branch was cut.

Both were verified on hardware after the split: firmware built from `fix/mcm-stability`, deployed to the Ultimate 64 over JTAG and to the Ultimate II+L through its updater, then `./run-tests c64u u64 u2@c64u -o runs/ --syslog --record`. All 69 suite runs passed across the three targets with no recoveries.

This branch is left in place, so everything here stays readable.

Issue #775, opened 2026-08-12
Proposal: an outstanding-allocation tracker to go with machine:heap


barryw:
`GET /v1/machine:heap` (#766) tells you a leak exists. It does not tell you
where, and finding that has been the slow part of every leak fixed since.

I have been running a tracker on a debug branch to close that gap. It found
four of the five leaks in #770, #772 and #773, and none of them were reachable
by reading code -- each one had correct-looking ownership on every path around
it. Raising this before sending a PR, because it adds a build option and two
endpoints, and I would rather agree the shape first. Both @GideonZ and
@chrisgleissner asked what the tooling was, so this writes it up properly.

## What it does

Records every live allocation with the address that asked for it, and reports
them grouped by that address.

- A table written from `pvPortMalloc` and cleared from `vPortFree`, holding
  pointer, size and `__builtin_return_address(0)`.
- Slots addressed directly from the pointer rather than searched, so record and
  forget stay constant-time. They sit in the allocator path; a linear scan
  there is not affordable once every allocation is tracked.
- Four ways per set. A plain direct-mapped table lost a noticeable number of
  records to collisions on real runs. When a set is genuinely full the oldest
  entry is dropped and counted, so an under-reporting table says so rather than
  quietly lying.
- `PUT /v1/machine:heapblocks_reset`, `GET /v1/machine:heapblocks`.

`operator new` and `malloc` both funnel through `get_mem()` in
`memory_wrap.cc`, so it re-attributes each block to the real caller there;
without that every allocation reads as the wrapper.

## The part that actually finds leaks

The tracker alone is not enough, because a live block is not a leaked block.
The workflow is differential: reset, run a suite twice, compare per-caller
counts. Genuinely live objects hold steady; leaked ones grow by the same amount
each round.

That is what separated four stranded `FileInfo` copies per run from the seven
that were simply the directory listing on screen (#773), and it is what stopped
me attributing that leak to the disk image mount cache -- the totals looked
right for it and were not. I would send a small host-side script for this
alongside, since the raw endpoint invites exactly that mistake.

## Cost, and keeping it out of release images

Everything is behind `HEAP_TRACK`, undefined by default, with the entry points
compiling to nothing when it is off -- no table, no allocator-path work. The
default table is 8192 slots at 16 bytes, so 128 KB of BSS on a diagnostic
build, overridable at build time, as is a minimum block size for when you are
chasing one big buffer rather than many small objects.

[...]

PR #773, opened 2026-08-12
Free the directory listings that get_directory hands out

barryw:
`FileManager::get_directory()` fills the caller's list with `new FileInfo`
copies, so the list and its contents have to be freed separately. Three callers
freed only the list:

- `FileManager::delete_recursive()`
- the directory branch of `FileManager::fcopy()`
- `SystemInfo::storage_info()`, whose list is a local, so it frees nothing at all

Every entry stranded a `FileInfo` and the name buffer it carries, about 170
bytes each.

The rule is already written down twice in the tree -- `vfs_closedir()` and the
MPS printer both delete the entries before the list, the latter with a
`// FileInfo not needed anymore` comment, and `BrowsableDirEntry::getSubItems()`
notes that it passes ownership on instead. These three just missed it, so the
two in `filemanager.cc` now share a helper that states the rule once.

`SystemInfo::storage_info()` is the one that matters outside a test run: it
leaks one `FileInfo` per root entry every time the System Information page is
opened.

## How it was isolated

An allocation tracker, reset before the run, then the `browser-filesystem-refresh`
suite run twice. The `FileInfo` copies went from 7 live to 11 while every other
allocation site stayed flat, so four per run were being stranded rather than
merely being the listing on screen. Four copies at 172 bytes is 688, and the
measured slope was 696 including one block header.

## Measurement

Ultimate 64 Elite, warm slope over `GET /v1/machine:heap`, settling before each
sample:

| build | bytes per run |
|---|---:|
| `test-merge` | 16,364 / 16,648 / 16,656 |
| with #772 | 704 / 688 / 692 |
| with #772 and this | 28 / -8 / -24 |

The first step is #772; this is the second.

## The soak suite

Nothing measured this path. The e2e suite asserts that Menu, Telnet and FTP
converge on the same file system, which it did just as happily while the
browser leaked 16.6 KB per run -- it never looks at the heap.

`browser-refresh-leak` runs the matrix as a subprocess, the way the runner does,
and requires the free heap back. Proven in both directions on hardware:

| build | verdict |
|---|---|
| #772 + this fix | 0 bytes over 3 runs, **passes** |
| #772 alone | 688 bytes/run against a 250-byte tolerance, **fails** |

**It depends on #772.** Without it the entry a browser drops on removal is
stranded and this suite fails at about 16.6 KB per run, so this PR should land
after that one.

One rough edge worth knowing: the underlying matrix suite can fail on a device
that has just booted, and a failed run yields no heap verdict, so this suite
[...]

PR #772, opened 2026-08-12
Reload the listing when a browsed entry is removed

barryw:
`TreeBrowser::checkFileManagerEvent()` unlinked the `Browsable` for a removed
node and left it:

```c
b = st->node->findChild(event->newName.c_str());
if (b) {
    printf("Removing %s\n", b->getName());
    st->node->children.remove(b);
}
```

`IndexedList::remove()` only drops the pointer out of the array. A
`BrowsableDirEntry` owns a `FileInfo`, a `FileType`, its generated FAT name and
a `FileManager` path reference, so every file that disappeared under an open
browser stranded all of it -- around 300 bytes, and one path handle from a
bounded pool, which is the part that worries me more than the bytes.

This is not test-only: deleting files over FTP, from a C64 program or from a
second Telnet session while a browser sits on that directory is ordinary use.

## Why not just delete it

Deleting the `Browsable` at that point trades a leak for a use-after-free. This
state's `under_cursor` still names it, and on the path where the event names the
browser's own directory a deeper state can still hold it as its `node`.
`level_up()` unwinds those states only on the other branch.

Marking the state for reload frees the children through the existing
`killChildren()` path and rebuilds the list from the filesystem, and
`do_refresh()` runs the reload before anything reads `under_cursor` again. It is
also exactly what `eNodeAdded` and `eRefreshDirectory` do with the same list
twenty lines above, so the event paths now behave the same way.

## Measurement

Ultimate 64 Elite, warm slope over `GET /v1/machine:heap`, settling six seconds
before each sample. The `browser-filesystem-refresh` suite deletes files over
FTP and checks the menu notices, so it is the one that provokes this.

| build | bytes per run |
|---|---:|
| before | 16,364 / 16,648 / 16,656 |
| after | 704 / 688 / 692 |

An allocation tracker on the before build attributed it to 56 `FileInfo`
copy-constructions, 55 `BrowsableDirEntry` allocations from
`BrowsableDirEntry::getSubItems()` and their strings, for 56 removals in a run,
with no other site accounting for it.

The same bug showed up in `prg-context-menu`, which removes its fixtures and
drives at teardown: with the disk image mount cache bounded (#769) that suite
sat at 3,020 bytes per run, and with this fix as well it reads 80 and 40.

A residual of about 695 bytes per run remains in `browser-filesystem-refresh`
after this change. It is consistent across runs, so it is a real and separate
thing rather than noise, and it is not this bug; it is being chased separately.

## Testing

[...]

PR #770, opened 2026-08-12
Fix four heap leaks in the Assembly 64 search and the tree browser

barryw:
Running the Assembly 64 search leaked about 31.4 KB of heap per pass through
its E2E suite. Chasing that down found four causes; the last one is not
Assembly 64 code at all and costs every config menu open 60 bytes.

Measured on an Ultimate 64 Elite, warm slope over `GET /v1/machine:heap`,
settling six seconds before each sample:

| stage | bytes per `assembly64` run |
|---|---|
| before | 31,440 / 31,520 |
| after the query-path fixes | ~780 |
| after the browser-state fix | 0 / 64 / -20 |

What is left oscillates in both directions around a fixed free-heap figure, so
it is allocator fit rather than accumulation.

## The response buffers

`collect_in_buffer()` allocates a 16,392-byte `t_BufferedBody` per request and
parks it in the request's user context. `convert_buffer_to_json()` copies every
key and value into the JSON tree -- the parser builds `JSON_Object(true)`, so
keys are owned, and `JSON_String` holds an `mstring` -- so the buffer has no
owner once the tree exists.

Only the query path deleted it. `request_entries()` dropped one on every entry
opened, along with the whole entries JSON tree, and `get_presets()` held one
for the lifetime of the box. `Assembly::take_response_json()` now frees the
body itself, which puts it in one place instead of asking each caller to
remember. `request_binary()` is unaffected: its user context is a
`TempfileWriter` the caller still owns.

An outstanding-allocation tracker confirmed this before the fix: one live
block of 16,400 bytes after a single suite run.

## The query results

`BrowsableQueryResults` had no owner either. It is handed to
`AssemblyResultsView` as its node, and `TreeBrowserState` only ever kills a
node's children, never the node, so both it and the `BrowsableQueryResult`
entries inside it leaked per query. The search form holds it now: it outlives
the results view, which deletes its `previous` last, so freeing it from
`~AssemblySearchForm` is ordered safely against the base destructor's use of
`node`. Its own destructor frees the entries, which live in a list of their own
rather than in `Browsable::children`.

## The browser state, which is the general one

`TreeBrowser`'s constructor makes a plain `TreeBrowserState`, and three
subclasses then assign over it with a state of their own: `AssemblySearch`,
`ConfigBrowser` and `FormUI`. The one the constructor made is never freed. That
is 60 bytes every time one of these browsers opens -- measured at 61 bytes per
Assembly 64 form open, and the same on every config menu.

[...]

PR #776, opened 2026-08-12
Outstanding-allocation tracker behind a build flag

barryw:
`machine:heap` (#766) says a leak exists. This says where. It found four of the
five leaks in #770, #772 and #773, none of which were reachable by reading
code -- each had correct-looking ownership on every path around it.

## Shape

- `software/system/heap_track.{c,h}` -- the table and the report. All of it
  behind `HEAP_TRACK`; with the flag undefined the entry points are macros that
  expand to nothing, no table is reserved and the allocator does no extra work.
- `heap_4.c` -- four guarded lines, no logic. I put the implementation in its
  own file specifically so the FreeRTOS vendor diff stays trivial.
- `memory_wrap.cc` -- one call. `operator new` and `malloc` both funnel through
  `get_mem()`, so without re-attributing there, every allocation reads as that
  wrapper rather than as the code that asked.
- `route_machine.cc` -- `GET /v1/machine:heap_allocations` and
  `PUT /v1/machine:heap_allocations_reset`, compiled only with the flag.
- Makefiles -- `$(EXTRA_DEFINES)` appended to `OPTIONS` (the hook #763 proposed,
  which did not land with #766) plus `heap_track.c` in the source list, for the
  six actively built targets.

Slots are addressed directly from the pointer rather than searched: record and
forget sit in the allocator path and a linear scan there is not affordable once
every allocation is tracked. Four ways per set, because a plain direct-mapped
table lost a noticeable number of records to collisions on real runs. A full
set drops its oldest entry and counts it in `evicted`, so a table that is
under-reporting says so rather than quietly lying.

## Gating, demonstrated

```
flag off:  update.u64 5,374,124 bytes,  no machine:heap_allocations in the image
flag on:   update.u64 5,375,552 bytes,  both endpoints present
```

The 128 KB table is BSS, so it does not appear in the image either way. Both
configurations build clean on the CI-matched Quartus 18.1 nios2 toolchain.

## The part that actually finds leaks

`tests/tools/heap_diff.py`. A live block is not a leaked block, and on real
firmware most outstanding blocks are legitimate -- the directory on screen, the
current session, cached structures. Reading the endpoint once and treating the
big entries as leaks is how you fix the wrong thing.

So it runs the same work twice and reports which callers grew between rounds.
Steady callers are holding live objects; growing ones are leaking, by the
amount they grow. That is what separated four stranded `FileInfo` copies per
run from the seven that were merely the listing on screen in #773 -- and what
stopped me attributing that leak to the disk image mount cache, whose totals
looked right for it and were not.

```
tests/tools/heap_diff.py -H <device> -- ./run-tests -H <device> -s prg-context-menu
```

It goes through `tests/lib/rest.py` like everything else; `transport-usage`
passes.

## Status and what I want to hear

**Not yet verified on hardware in this form.** The tracker it descends from has
been in daily use all week, but this is a rewrite -- four-way sets, sorted
report, compiled-out gating -- and I have not yet run *this* code on a device.
Happy to do that before it leaves draft; I did not want to build on a shape
nobody has agreed to.

The open questions are in #775, and the two that matter:

1. **Vendor hooks or not.** Completeness argues for `pvPortMalloc`/`vPortFree`
   as here. Hooking only `memory_wrap.cc` would touch no FreeRTOS code and
   would still have caught every leak found so far, at the cost of missing task
   stacks, queues and direct `pvPortMalloc` callers.
2. **Whether you want a debugging tool in the shipping tree at all**, gated but
   present. If the answer is that it should stay on a branch, that is a fine
   answer and nothing is wasted -- it lives on one today.

@chrisgleissner -- this touches no part of the e2e runner, and `heap_diff.py`
is standalone, so it should not collide with your multi-device work.

### DIFF
--- a/software/FreeRTOS/Source/MemMang/heap_4.c
+++ b/software/FreeRTOS/Source/MemMang/heap_4.c
@@ -84,6 +84,7 @@ task.h is included from an application file. */
 
 #include "FreeRTOS.h"
 #include "task.h"
+#include "heap_track.h"
 
 #undef MPU_WRAPPERS_INCLUDED_FROM_API_FILE
 
@@ -294,6 +295,7 @@ void *pvReturn = NULL;
 	#endif
 
 	configASSERT( ( ( ( uint32_t ) pvReturn ) & portBYTE_ALIGNMENT_MASK ) == 0 );
+	heap_track_record( pvReturn, xWantedSize, __builtin_return_address( 0 ) );
 	return pvReturn;
 }
 /*-----------------------------------------------------------*/
@@ -305,6 +307,8 @@ BlockLink_t *pxLink;
 
 	if( pv != NULL )
 	{
+		heap_track_forget( pv );
+
 		/* The memory being freed will have an BlockLink_t structure immediately
 		before it. */
 		puc -= xHeapStructSize;
--- a/software/api/route_machine.cc
+++ b/software/api/route_machine.cc
@@ -654,11 +654,13 @@ API_CALL(GET, machine, heap_allocations, NULL, ARRAY( {  }))
     char buf[64];
 
     int n = heap_track_report(callers, HEAP_TRACK_REPORT_CALLERS, &totals);
+    JSON_List *caller_list = JSON::List();
+    resp->json->add("caller", caller_list);
     for (int i = 0; i < n; i++) {
         // small_printf has no %lu, so every field is printed as %d or %p.
         sprintf(buf, "%d allocations %d bytes ra %p", (int)callers[i].blocks,
                 (int)callers[i].bytes, callers[i].caller);
-        resp->json->add("caller", buf);
+        caller_list->add(buf);
     }
     resp->json->add("live_allocations", (int)totals.live_blocks);
     resp->json->add("live_bytes", (int)totals.live_bytes);
--- a/software/api/route_machine.cc
+++ b/software/api/route_machine.cc
@@ -635,19 +635,19 @@ API_CALL(GET, machine, heap, NULL, ARRAY( {  }))
 // build made with EXTRA_DEFINES=-DHEAP_TRACK=1; see software/system/heap_track.h.
 //
 // Read these differentially: reset, exercise the device, read, exercise it
-// again, read again. A block that is live is not a block that is leaked, and
+// again, read again. An allocation that is live is not one that is leaked, and
 // only the growth between two identical rounds separates them.
 #include "heap_track.h"
 
 #define HEAP_TRACK_REPORT_CALLERS 48
 
-API_CALL(PUT, machine, heapblocks_reset, NULL, ARRAY( {  }))
+API_CALL(PUT, machine, heap_allocations_reset, NULL, ARRAY( {  }))
 {
     heap_track_reset();
     resp->json_response(HTTP_OK);
 }
 
-API_CALL(GET, machine, heapblocks, NULL, ARRAY( {  }))
+API_CALL(GET, machine, heap_allocations, NULL, ARRAY( {  }))
 {
     static HeapTrackCaller_t callers[HEAP_TRACK_REPORT_CALLERS];
     HeapTrackTotals_t totals;
@@ -656,11 +656,11 @@ API_CALL(GET, machine, heapblocks, NULL, ARRAY( {  }))
     int n = heap_track_report(callers, HEAP_TRACK_REPORT_CALLERS, &totals);
     for (int i = 0; i < n; i++) {
         // small_printf has no %lu, so every field is printed as %d or %p.
-        sprintf(buf, "%d blocks %d bytes ra %p", (int)callers[i].blocks,
+        sprintf(buf, "%d allocations %d bytes ra %p", (int)callers[i].blocks,
                 (int)callers[i].bytes, callers[i].caller);
         resp->json->add("caller", buf);
     }
-    resp->json->add("live_blocks", (int)totals.live_blocks);
+    resp->json->add("live_allocations", (int)totals.live_blocks);
     resp->json->add("live_bytes", (int)totals.live_bytes);
     resp->json->add("distinct_callers", n);
     // Non-zero means the table dropped records and is under-reporting.
--- a/software/api/route_machine.cc
+++ b/software/api/route_machine.cc
@@ -629,3 +629,42 @@ API_CALL(GET, machine, heap, NULL, ARRAY( {  }))
     resp->json->add("total", (int)configTOTAL_HEAP_SIZE);
     resp->json_response(HTTP_OK);
 }
+
+#ifdef HEAP_TRACK
+// Where the leak is, as opposed to whether there is one. Present only in a
+// build made with EXTRA_DEFINES=-DHEAP_TRACK=1; see software/system/heap_track.h.
+//
+// Read these differentially: reset, exercise the device, read, exercise it
+// again, read again. A block that is live is not a block that is leaked, and
+// only the growth between two identical rounds separates them.
+#include "heap_track.h"
+
+#define HEAP_TRACK_REPORT_CALLERS 48
+
+API_CALL(PUT, machine, heapblocks_reset, NULL, ARRAY( {  }))
+{
+    heap_track_reset();
+    resp->json_response(HTTP_OK);
+}
+
+API_CALL(GET, machine, heapblocks, NULL, ARRAY( {  }))
+{
+    static HeapTrackCaller_t callers[HEAP_TRACK_REPORT_CALLERS];
+    HeapTrackTotals_t totals;
+    char buf[64];
+
+    int n = heap_track_report(callers, HEAP_TRACK_REPORT_CALLERS, &totals);
+    for (int i = 0; i < n; i++) {
+        // small_printf has no %lu, so every field is printed as %d or %p.
+        sprintf(buf, "%d blocks %d bytes ra %p", (int)callers[i].blocks,
+                (int)callers[i].bytes, callers[i].caller);
+        resp->json->add("caller", buf);
+    }
+    resp->json->add("live_blocks", (int)totals.live_blocks);
+    resp->json->add("live_bytes", (int)totals.live_bytes);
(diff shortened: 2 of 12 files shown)


chrisgleissner:
Hi @barryw ,

Thanks for raising that and for sharing more of the amazing and very powerful infrastructure you have built in such a short time. 

It is clearly powerful as it has found multiple issues with ease, so whether or not it is useful is IMHO not a question. In terms of the concrete approach taken - there may be alternatives out there when it comes to static code analysis, runtime monitoring, etc. However, an easy, readily available solution that is part of the repo and does not require separate tool installations always wins in my book unless it smells of "reinventing the wheel", which this one does not in my opinion.

A few thoughts after looking at this alongside the REST API we already have:

* I normally favour developer-facing diagnostics being available in standard builds rather than hidden behind flags. In this case, though, I think the flag is justified: the tracker reserves 128 KB and adds work to the allocator path even when nobody is calling the REST endpoint. That's materially different from something lightweight like `machine:heap`.

* On the API shape, I originally thought we should make this more idiomatic REST with nested resources. Having gone back through the existing API, I'm less convinced. We already consistently use the `route:command` model throughout - `machine:reset`, `drives/{drive}:mount`, `streams/{stream}:start`, etc. At this point I think consistency and least surprise within *our* API matter more than making these two endpoints REST-by-the-book.

* I would, however, avoid adding two commands for what is really one concept. Something like this feels cleaner to me:

```text
GET /v1/machine:heap_allocations
PUT /v1/machine:heap_allocations_reset
```

`GET` reports the tracked outstanding allocations; `PUT` resets the tracking state. I at first thought about `DELETE` without the `_reset` being better, but that makes it sound as though the caller would delete heap allocations, which is wrong and would be alarming. 

That also follows the precedent of endpoints such as `machine:debugreg`, where the HTTP method determines what happens to the same conceptual resource.

I'd also prefer `allocations` over `heapblocks`. What we're exposing here isn't really FreeRTOS heap blocks as an implementation detail - it's outstanding allocations, grouped by their allocating caller.

--- 

More generally, I'd like us to avoid gradually accumulating lots of narrowly named endpoints for individual firmware internals. 

Developer visibility is extremely useful, but ideally the API should still feel cohesive, predictable and extensible a few years from now. `machine:heap` plus one related allocation-tracking endpoint feels reasonably contained to me.

I don't think we've reached this point yet. If we feel we are getting close, we may also consider exposing more internals via a dedicated `/metrics` Prometheus endpoint which can be done with very minimal overhead; something @GideonZ and I discussed before.

barryw:
@chrisgleissner Thanks — renamed, and I took your naming point further than the URL.

You are right that "blocks" named the allocator's internal structure rather than the thing being measured, and the payload carried the same wrong noun, so both moved:

```
GET /v1/machine:heap_allocations
PUT /v1/machine:heap_allocations_reset
```

`live_blocks` is now `live_allocations`, and each caller line reads `"3 allocations 1024 bytes ra 0x…"`. `heap_diff.py` parses that line, so it moved with it. The tracker's internals still say blocks — inside the allocator they genuinely are blocks; it was only the API surface that misdescribed itself.

Both configurations rebuilt on the CI-matched Quartus 18.1 nios2 toolchain after the rename:

```
flag off:  update.u64 5,374,124 bytes,  no machine:heap_allocations in the image
flag on:   update.u64 5,375,552 bytes,  both endpoints present
```

Worth noting the `#ifdef HEAP_TRACK` means the flag-off build never compiles the renamed code, so the flag-on build is the only one that could have caught a mistake in the `API_CALL` registration. That is the one that matters here and it is clean.

On the flag: agreed, and for the reason you give. `machine:heap` is three counters read on demand; this reserves 128 KB and adds work to every allocation. If it were free I would argue for always-on too.

On `route:command` over nested resources — agreed, and thank you for going back through the existing API rather than taking the first answer.

One thing I would put back to you. You wrote that you would avoid two commands for what is really one concept, and cited `machine:debugreg` as the precedent where the method decides. But `debugreg` really is one command name with two methods (`route_machine.cc:264` and `:272`), whereas `heap_allocations` plus `heap_allocations_reset` is still two names sharing a prefix. Your principle seems to point at:

```
GET /v1/machine:heap_allocations     # report
PUT /v1/machine:heap_allocations     # reset
```

I have implemented your written version rather than guess at which you meant. The argument for keeping `_reset` is real — a bare `PUT` with no body that silently resets is less self-documenting, and `debugreg`'s `PUT` at least takes a required `value`. Happy either way; it is a two-line change and it is your call on which reads better across the API as a whole.

On endpoint sprawl and a `/metrics` endpoint — no argument from me, and it deserves its own discussion rather than being settled in the margins of this PR. For what it is worth I do not think this endpoint is a candidate to fold into it. A per-call-site allocation table is exactly the unbounded, dynamic label dimension a scrape target should not carry; coarse heap gauges belong in `/metrics`, and "which caller is leaking" belongs behind a debug flag. They answer different questions, and I think both are worth having.

This stays a draft either way, pending the design questions in #775 and the `$(EXTRA_DEFINES)` makefile hook from #763 that did not land with #766.

chrisgleissner:
Hi @barryw ,

great work on all the PRs.

One question - where do we stand with the various PRs and especially the draft PRs? Is there anything that you are waiting on from me or Gideon for you to move on with them and get them into a merge-ready state?

I see you currently have these PRs open:
- #791 
- #789 
- #788 
- #776

It would be good to understand what their dependencies are and what, if anything, blocks them from moving out of draft status, and then from being merged.

This also in regards to this PR dependency write-up from above:

"This stays a draft either way, pending the design questions in #775 and the $(EXTRA_DEFINES) makefile hook from #763 that did not land with #766."

Many thanks
Christian

barryw:
Hi Christian — thanks for asking. The short version is that **#789 is first in the chain**; #788 and #791 are waiting on it only so their full E2E gates can be rerun green. #776 is independent and is waiting on a design decision rather than another implementation PR.

### #789 — configuration effectuation lifecycle

- CI is green, the branch is mergeable, the focused regression passes on its first run after flashing, full hardware E2E is 22/22, and all application/space-gate builds pass.
- Your churn question was measured and resolved; the current direction is the smaller change.
- From my side the implementation is ready for review.
- The remaining ask is a sanity check from Gideon that the lifecycle boundary now matches the intent behind his comment on #787. If it does, I can take it out of draft immediately. If not, I need the preferred boundary called out.

### #788 — live network service switches

- The implementation, CI build, focused hardware regression, and cross-target builds are complete.
- Its only draft blocker is #787: the full gate is 22/23 because `cfg-single-group` fails independently of the network change.
- Once #789 lands, I will rebase #788, rerun the full hardware gate, and mark it ready. There is no additional design decision needed for this one.

### #791 — runtime palette UCI commands

- The implementation, host parser tests, CI, U64/U64-II builds, 45/45 UCI hardware checks, and the documented assembly/BASIC examples are complete.
- It has the same #787/#789 gate dependency: its full run is 21/22 only because of `cfg-single-group`.
- The companion documentation is in GideonZ/1541u-documentation#29. The code PR no longer depends on deciding whether all documentation should move into this repository; that broader discussion can remain separate. The companion docs still need review/merge.
- After #789 lands, I will rebase, rerun the full gate, and mark #791 ready.

### #776 — allocation tracker

- This one is independent of the three above. CI is green and both flag-off and flag-on builds are proven, but this rewritten tracker has not yet had its final hardware run.
- The quoted `$(EXTRA_DEFINES)` dependency is stale: #776 now includes that small hook itself, so it is **not blocked on #763**.
- The remaining Gideon decision is whether to accept the gated diagnostic tool in the shipping tree and whether the complete `pvPortMalloc`/`vPortFree` hooks in `heap_4.c` are acceptable, versus the narrower `memory_wrap.cc` coverage.
- One small API choice remains from our earlier exchange: keep the currently implemented `PUT /v1/machine:heap_allocations_reset`, or use `PUT /v1/machine:heap_allocations` so the HTTP method distinguishes report from reset. If you have a preference, I will use it; otherwise I will leave the explicit `_reset` form.
- Once those choices are settled, I will run the final hardware differential test and move it out of draft if it behaves as expected.

So the practical order is:

1. Review/settle #789.
2. I rebase and fully retest #788 and #791, then mark both ready.
3. Gideon settles the two #776 design points; I hardware-test that final shape and mark it ready.

There is no implementation work I am waiting for you to do. What would help most is review of #789, review of the palette documentation PR, and the two explicit decisions on #776.

chrisgleissner:
Thanks for the overview @barryw . I now asked a question on #789. It is sometimes hard to see what exactly a PR is waiting for, especially if a question was asked during the PR and the discussion then moved on.

Thus, to avoid that PRs sit in Draft status indefinitely, I recommend to occasionally check back here and see if any question you are waiting for has been answered, and then just ask it again in a new comment.

@GideonZ  FYI Barry's reply above shows how all his PR's are related and where he is waiting for your steer.

barryw:
@chrisgleissner I get emails when my PRs or issues change, so I keep on top of stuff. ;)

chrisgleissner:
@barryw Thanks for moving this and other PRs to merge ready (provided they are merge ready, i.e. all open questions resolved, tested. etc.) and resolving any merge conflicts. Keen to get stuff merged.

chrisgleissner:
Hi @barryw , 

the title of this PR says 'Draft: ' but the status of the PR itself is no longer 'Draft'. 

To avoid ambiguity, I suggest we only use the 'Draft' status of PRs (which clearly flags them as such even in the PR overview) and don't duplicate the word 'Draft: ' in the PR title.

Is this PR ready for review and merge?

Thanks
Christian

barryw:
> Hi @barryw ,
> 
> the title of this PR says 'Draft: ' but the status of the PR itself is no longer 'Draft'.
> 
> To avoid ambiguity, I suggest we only use the 'Draft' status of PRs (which clearly flags them as such even in the PR overview) and don't duplicate the word 'Draft: ' in the PR title.
> 
> Is this PR ready for review and merge?
> 
> Thanks Christian

Thanks, @chrisgleissner. Fixed. This is ready, but @GideonZ wants to noodle on it a bit

PR #784, opened 2026-08-15
Preserve E2E run evidence

chrisgleissner:
## Summary

This PR adds a complete observability layer around E2E runs so failures can be diagnosed from preserved artifacts rather than terminal scrollback.

Highlights:

- `run-tests -o DIR` preserves a complete structured run.
- Every REST, Telnet and FTP interaction is recorded.
- `tools/e2e_report.py` generates a standalone Markdown report with verdicts, failure captures, logs, machine state and reproduction commands.
- `--syslog` collects and attributes each device's firmware log.
- `--record` records device video, audio, test progress, interactions and subtitles.
- Failed suites are retried up to `--attempts N`, with every attempt preserved.
- Ultimate 64, C64 Ultimate and Ultimate II+L targets are supported, including `cartridge@computer` targets such as `u2@c64u`.
- The hardware E2E workflow runs from a detached worktree so edits to another checkout cannot affect a running gate.

This PR changes no firmware. The implementation is confined to the test harness and supporting tooling, apart from small Makefile, workflow and `.gitignore` changes.

### Sample U64 E2E recording

https://youtu.be/6PyTosV2zks

## Run artifacts

`-o DIR` writes `index.md` plus one directory per target.

| Artifact | Contents |
| --- | --- |
| `run.jsonl` | Run plan, health, verdicts and metadata |
| `<label>-<suite>.jsonl` | Checks and device actions |
| `<label>-<suite>.log` | Suite console output |
| `interactions.jsonl` | REST, Telnet and FTP interactions |
| `transcript.txt` | Human-readable interaction transcript |
| `screens.jsonl` | Screens read by suites |
| `screen-text.jsonl` | C64 screen text decoded from recorded video |
| `capture/` | Failure captures and recorder stills |
| `attempt-<n>/` | Artifacts from retried attempts |
| `syslog.txt` | Device firmware log |
| `video.mp4`, `video.srt` | Recording and test subtitles |

Record formats are documented in `tests/lib/README.md`; the detailed specification is in `tests/e2e/doc/observability-spec.md`.

Observability is deliberately non-invasive: failures in recording, logging or report generation are reported but do not change a suite's test verdict.

The runner also hashes the harness files at the start and end of a run. This detects changes to the code being executed while a run is in progress.

Hardware gates additionally execute from a detached worktree at the commit being tested, preventing edits in another checkout from affecting running suites.

## Report

`python3 tools/e2e_report.py DIR` generates a standalone Markdown report without contacting a device.

For each failure it includes:

- the screen at failure;
- the relevant console log;
- known machine state;
- nearby device-log entries;
- the exact reproduction command.

The report also distinguishes failed, incomplete, killed, skipped and retried runs.

Retries are shown explicitly so an intermittent pass cannot silently hide an earlier failure.

## Device logging

`--syslog` collects each target's firmware log.

The runner reads each device's configured syslog destination and listens on the corresponding UDP port. This allows logs to be attributed by destination port rather than relying only on source IP address.

That matters for devices with multiple interfaces. On the hardware run, the Ultimate 64 was addressed at `192.168.1.15` but sent almost all of its log from WiFi address `192.168.1.71`. Because its configured syslog port uniquely identified it, all 38,765 lines were still attributed correctly.

Across the three-target run:

| Target | Log lines |
| --- | ---: |
| C64 Ultimate | 34,640 |
| Ultimate II+L | 24,503 |
| Ultimate 64 | 38,765 |
| **Total** | **97,908** |

No lines were left unattributed.

The harness only reads syslog configuration. It never modifies device settings.

Syslog remains best-effort. The artifacts can show that no log arrived, but cannot reliably distinguish every possible firmware-side reason for that silence.

## Recording

`--record` captures:

- device video and audio;
- the harness view;
- suite and check state;
- subtitles and chapters;
- the interaction stream;
- frame timecodes.

Failure stills are tied to exact recorded frames and can be reproduced pixel-for-pixel from the recording.

The recorder also decodes the C64's 40x25 character screen from video into `screen-text.jsonl`, avoiding additional REST reads solely for observability.

## Multi-device runner

Targets can be standalone machines or `cartridge@computer` combinations:

```text
u64
c64u
u2@c64u
```

`tests/lib/targets.py` resolves REST, FTP, Telnet, keyboard, video, audio and logging endpoints.

Targets on different physical machines run concurrently. Targets sharing the same machine take turns.

This is particularly important for cartridge targets. Keyboard injection belongs to the computer hosting the cartridge, while other requests generally belong to the cartridge itself.

Three suites bypassed this routing by building URLs directly. They now use the shared routing helpers, and `tests/lib/runner_policy_test.py` prevents direct REST URL construction from being reintroduced.

## Retries

Every failing suite is now retried up to `--attempts N` total executions.

The default is:

```text
--attempts 3
```

so a failed suite may run up to three times.

`--no-retry` is equivalent to:

```text
--attempts 1
```

The attempt limit is absolute, including retries following device recovery.

Each additional attempt preserves its own console output and failure capture under `attempt-<n>/`, while JSONL records remain in the main per-suite file and carry an explicit `attempt` field.

Retries are deliberately visible:

- the console prints the attempt;
- the report includes a `## Retries` section;
- recordings and subtitles identify the attempt;
- the final result becomes `WARN` rather than `OK` if everything passes only after retry.

## Command line

`-j/--jsonl-dir` has been replaced by:

```text
-o, --output-dir
```

There is no compatibility alias because the directory now contains much more than JSONL.

`./run-tests --help` has also been reorganised around common usage and documents examples and exit statuses.

### Exit status

Exit codes form an ordered severity scale:

| Status | Meaning |
| ---: | --- |
| 0 | all suites passed first time |
| 1 | all passed, but at least one required a retry |
| 2 | all passed, but a device required recovery |
| 3 | at least one suite failed every attempt |
| 4 | a device could not be recovered |
| 64 | invalid command line |

This allows callers to use meaningful thresholds, for example:

```sh
[ $? -le 2 ]
```

to accept clean runs, retries and recoveries, but reject actual test failures.

## CI

`.github/workflows/build.yml` now runs:

```sh
make observability_test
```

before building firmware.

This catches harness and report regressions quickly without requiring hardware.

A separate `.github/workflows/e2e.yml` runs the hardware gate because it:

- requires a self-hosted runner on the device LAN;
- takes much longer than compilation;
- must not overlap with another run against the same hardware;
- should remain separate from firmware build failures.

The hardware workflow runs:

```text
c64u u64 u2@c64u
```

with both `--syslog` and `--record`.

It is scheduled rather than triggered on every push because the hardware is physical and shared.

Other changes outside the harness are limited to:

- a Makefile `observability_test` target;
- `.gitignore` ignoring generated `**/*.cfw` files.

## Artifact audit

The generated artifacts were audited against `tests/e2e/doc/observability-spec.md`.

That audit found 22 defects, including issues that a normal green E2E run would not expose because observability must not affect the test verdict.

Notable fixes include:

- password redaction no longer changes ordinary text or leaks secrets from `U64_PASS`, `curl -u` or prose;
- burned-in check numbers and subtitle check numbers now agree;
- interaction `seq` values are unique across processes;
- deduplicated interactions are flushed rather than being delayed indefinitely or lost when a process exits;
- runner warnings are included in the report;
- configuration changes and restored values are reported correctly;
- file creation/deletion operations are detected correctly;
- report acceptance tests can no longer silently turn into skips;
- interaction records consistently identify their target and attempt;
- structured request parameters are stored as JSON rather than Python `repr`;
- FTP data transfers are recorded;
- timelines contain usable wall-clock information;
- report paths resolve correctly;
- recording metadata includes all generated files;
- screen-text records use consistent check identifiers;
- transport errors are safely escaped in Markdown instead of being interpreted as HTML.

Six fixes were additionally verified by reverting them individually and confirming that their dedicated tests failed before restoring the fix.

### Device-free checks

| Check | Result |
| --- | --- |
| `make observability_test` | 171 cases, 2 skipped because no real encoder was available |
| `python3 tests/lib/runner_policy_test.py` | 71 checks |
| `python3 tests/lib/check_transport_usage.py` | 66 files checked |

### Firmware builds

CI run `31974122236` is green for:

- U2
- U2+
- U2+L
- U64
- U64-II

This PR changes no firmware, so these builds verify that the harness changes do not break the existing build.

### Syslog collection on hardware

A dedicated C64 Ultimate run was used to re-check the log collector:

```sh
./run-tests c64u -s ui-backend-smoke -o runs/collector --syslog
```

Result:

- suite passed;
- exit status 0;
- 582 syslog lines collected;
- no unattributed syslog file;
- all 582 lines attributed correctly by port.

### Three-target hardware gate

A complete gate was run against:

```sh
./run-tests c64u u64 u2@c64u -o runs/gate --syslog --record
```

It produced:

```text
RESULT: FAIL  targets=3  suites=72  ok=71  fail=1  warn=0  skip=0  recoveries=0  retried=1  exit=3
```

| Target | Result | Duration | Device log |
| --- | --- | ---: | ---: |
| c64u | 24/24 passed | 762s | 34,640 lines |
| u64 | 24/24 passed | 963s | 38,765 lines |
| u2@c64u | 23/24 passed | 1166s | 24,503 lines |

The failing `machine-code-monitor` suite exposed a real routing defect: it constructed URLs using the literal target token `u2@c64u`, which is not a hostname.

The fix routes requests through the shared target abstraction instead. On a cartridge target, keyboard input goes to the host computer while device operations go to the cartridge.

The routing change was subsequently verified on `u2@c64u`:

```sh
./run-tests u2@c64u -s machine-code-monitor -o runs/verify
```

During that run:

- all 18 keyboard requests returned HTTP 200;
- reset and menu requests returned HTTP 200;
- all 180 `menu_screen` requests returned HTTP 200.

The runner policy test now rejects suites that bypass the shared REST routing layer.

### Artifact validation

The three-target gate produced:

| Property | c64u | u2@c64u | u64 |
| --- | ---: | ---: | ---: |
| Interaction records | 6,466 | 7,640 | 8,201 |
| Duplicate sequence numbers | 0 | 0 | 0 |
| Missing body references | 0 | 0 | 0 |
| Invalid body hashes | 0 | 0 | 0 |
| Interactions missing target | 0 | 0 | 0 |
| Invalid screen sizes | 0 | 0 | 0 |
| Missing still pairs | 0 | 0 | 0 |
| Stills differing from recorded frames | 0 | 0 | 0 |

The generated report contained 13,827 lines and 1,524 referenced paths, all of which resolved. Regenerating it twice produced byte-identical output.

### Monitor suite on cartridge targets

After fixing REST routing, the current `machine-code-monitor` suite reaches the monitor on `u2@c64u` but fails its first screen assertion.

The test expects the Ultimate 64-class banking status line, while the cartridge monitor displays `CPU VIEW`.

This is a test limitation, not a firmware change introduced here. `software/monitor/` is untouched by this PR.

No capability skip is added because doing so would suppress cartridge coverage rather than make the test correctly support that layout.

### Two routing fixes lack hardware verification

The equivalent routing fixes in `telnet-stale-session` and `listener-soak` have not been exercised on hardware after the change.

They are covered by the same routing abstraction and by the runner policy test.

### `harness_changed=true`

Detection of a changed harness is covered by a device-free test, including modifying and restoring a tracked file.

No completed real hardware run has recorded `harness_changed: true`.

### Gate duration

The three-target gate took 1,928 seconds of wall-clock time.

Targets run concurrently where possible, but `u2@c64u` shares its physical machine with `c64u`, so those targets must take turns.

The runner itself remains lightly loaded; the limiting factors are device and network contention.


chrisgleissner:
Hi @GideonZ ,

This PR is now ready for review and merge. Full details in the PR description as usual. I tested it against both U64, C64U and U2.

---

Update 16 Aug: One thing I want to get rid of still is the churn around syslog. 

I introduced these changes since I noticed that a device that's both connected via Ethernet and WiFI sends its syslog packets via WiFI, not Ethernet. I changed it to prefer Ethernet in that case, but I think this change is better done in a dedicated PR if we decide to fix it, and I'll instead make the E2E framework tolerate either.

The fix for this should be done later this evening.

Many thanks
Christian

chrisgleissner:
Merged since after clean-up, all changes remaining were related to E2E test improvements, not firmware changes.

PR #785, opened 2026-08-15
Improve machine code monitor

chrisgleissner:
# Summary

This PR makes the machine-code monitor more reliable and consistent across the Ultimate 64 and Ultimate II+L, fixes several correctness and UI issues, and strengthens the E2E suite so transport and memory failures are detected rather than hidden by retries.

## Changes

The main changes are:

### All Devices

- Make assembled instructions and relocated operands atomic block writes.
- Make `Fill`, `Transfer`, `Compare`, `Hunt` and `Save` consistently use inclusive ranges.
- Add optional operand relocation to `Transfer`.
- Stabilize Assembly view across RAM, ROM and I/O boundaries.
- Make Back, Help, popups, command input and monitor exits consistent.
- Add structured `MCM` device logging for E2E diagnostics.
- Fix keyboard-injection batching exposed by the hardened E2E tests.
- Strengthen host and hardware coverage, including reliability tests where each input is sent exactly once.

### Ultimate II

- Fix unreliable main-memory writes on the Ultimate II while frozen.
- Add Ultimate II VIC-bank support and efficient block memory access.
- Fix larger `Transfer` operations on the Ultimate II.

### Related Changes

Most production changes remain inside `software/monitor`. The few cross-cutting changes are listed separately below with their justification.

## Related PRs

The following PR extends the `1541ultimate-documentation` repo with details about the machine code monitor, as well as details about a new REST API for keyboard injection and getting the current menu state:

https://github.com/GideonZ/1541u-documentation/pull/27

### Ultimate II+L memory writes

Hex edits to normal C64 RAM did not reliably reach main memory while the cartridge freezer was active.

There were two underlying problems.

First, the frozen Ultimax mapping was restored only from `$8000` upward, although the freezer's own window ends at `$0FFF`. The shared C64 access code now uses `$1000` as the boundary.

Second, writes could race the running 6510 while `C64_MODE` and `C64_DMA_MEMONLY` were being changed. Ultimate II+L reads and writes now use the same stopped-session mechanism, with synchronization around the FPGA-side mapping changes.

This fix is deliberately in the shared C64 layer rather than the monitor because the same defect also affected the device's own `machine:writemem` path.

Measured across six addresses on `u2@c64u`:

| Path | Before | After |
| --- | ---: | ---: |
| REST writes | 170/180 | 180/180 |
| Monitor hex edits | - | 36/36 |

Existing E2E failures at `$8000` and `$C000` now pass.

### Larger transfers on Ultimate II+L

`Transfer` previously alternated one-byte reads and writes.

On an Ultimate II+L, each access can require switching the frozen memory mapping. In practice, transfers longer than two bytes could therefore stop after the first two bytes.

Measured before the fix, at both `$1180` and `$C600`:

- 1-byte transfer: 1 byte copied
- 2-byte transfer: 2 bytes copied
- 4-, 8- and 16-byte transfers: only 2 bytes copied

`Transfer` now reads and writes chunks using block operations instead of alternating byte accesses.

The buffer is limited to one chunk rather than the complete range, so a full `$0000-$FFFF` transfer does not require a 64 KiB allocation.

Relocating transfers similarly operate on the copied range rather than repeatedly re-reading and rewriting individual instructions.

After the change, 16-byte, 64-byte and 256-byte transfers were verified on hardware at both low and high RAM addresses.

### Atomic instruction writes

Assembly edit previously wrote an instruction one byte at a time.

On a backend that must stop the C64 for each access, this could leave only an opcode or another partial instruction in memory.

Instructions are now committed through one `MemoryBackend::write_block()` call.

Wrapping writes at `$FFFF` retain the existing byte-wise fallback required by the API.

A host test verifies that assembling:

```text
LDA $1000
```

produces exactly one three-byte block write and no individual byte writes.

### Inclusive ranges

These commands all use the syntax `AAAA-BBBB`:

- `Fill`
- `Transfer`
- `Compare`
- `Hunt`
- `Save`

`Transfer` and `Compare` previously excluded the final address.

All five commands now consistently treat both endpoints as part of the range, including the complete `0000-FFFF` address space.

The overlapping-transfer check was also updated so a destination landing exactly on the source end uses the safe copy direction.

### Transfer operand relocation

`Transfer` can now optionally relocate absolute operands while moving code:

```text
T C000-C0FF,C100,C000-C07F
```

The fourth field identifies the source subrange containing code.

Absolute, absolute-indexed and indirect operands that point inside the copied source range are adjusted to point at the corresponding destination address.

The following remain unchanged:

- relative branches
- zero-page operands
- references outside the copied range
- incomplete instructions

Without the fourth field, `Transfer` behaves exactly as before.

Relocated operands are written as one two-byte block rather than as separate bytes.

Focused host tests cover the relocation algorithm and relevant failure modes.

### Assembly view

Assembly view is now deterministic when moving backwards and forwards through variable-length instructions and across different memory sources.

For `$D000-$DFFF`, RAM is still disassembled normally.

When I/O or character ROM is visible instead, the contents are shown as stable two-byte `DATA` rows:

```text
|D000 01 02        DATA 01 02       [I/O]|
|D002 03 04        DATA 03 04       [I/O]|
```

This avoids treating volatile I/O registers or character bitmap data as machine instructions.

The rule follows the actual banked source rather than simply the address.

DATA rows remain editable through the existing Assembly editor:

- each byte has its own edit position
- two hex digits complete one byte
- LEFT and RIGHT move byte-by-byte across row boundaries
- mnemonic input is ignored because there is no instruction to select
- writable I/O can still be edited
- character ROM remains read-only

Source indicators also have a fixed width:

```text
[RAM] [BAS] [KRN] [CHR] [I/O]
```

`D` remains intentionally unbound and reserved for a possible future Debug mode. `A` remains the Assembly shortcut.

### Back and exit behaviour

`RUN/STOP`, `ESC` and `←` now implement the same Back action.

They remove one interaction layer at a time:

1. Help
2. Number expression
3. Popup
4. Command prompt
5. Edit mode
6. Monitor

Where `←` is valid input, such as ASCII or Screen editing, it remains data.

A bare `X` no longer exits the monitor.

The monitor exits are:

- `C=+O`
- `RUN/STOP`
- `ESC`
- `←`

### Machine shortcuts

Two machine-level shortcuts are supported:

- `C=+R` resets the C64 and leaves the monitor.
- `C=+I` swaps the configured freeze-menu/HDMI-overlay interface and leaves the monitor.

On hardware without interface switching, `C=+I` reports:

```text
INTERFACE SWAP UNAVAILABLE
```

and changes nothing.

These shortcuts are handled below popup and editor layers so keys owned by an active control cannot accidentally reset the machine or switch interface.

Reset itself remains in `UserInterface::run_machine_monitor`, where monitor ownership can first be released in the correct order.

`MemoryBackend` only exposes the capability and does not perform the reset itself.

### Command input and Help

Command prompts now reject characters that cannot lexically occur in the command being entered.

Parsing and semantic validation on RETURN are unchanged.

The Help screen was reorganized into a single consistently aligned page and now reflects the actual mappings, including:

- `U` - `Undoc/Case`
- `O` - CPU bank
- `SH+O` - VIC bank
- no bare `X` Exit entry

The function key used to open Help is obtained from the application's key mapper rather than being hard-coded.

### Popup rendering

Popups are now rendered through one shared overlay path after the status row.

This ensures the popup owns every row it covers and prevents different redraw paths from disagreeing about layering.

## Ultimate II+L backend

`U2MemoryBackend` gains the capabilities required to make the monitor behave consistently with the U64 where the cartridge hardware allows it:

- block writes using one DMA operation
- stopped sessions around block reads and writes
- VIC-bank display and selection through CIA2 port A
- access to the freezer's saved CIA2 value

The Ultimate II+L has no monitor-selectable CPU bank, so its footer shows VIC-bank information instead.

## Device logging

The monitor now emits structured log entries beginning with `MCM` for significant actions, including:

- view and address changes
- fills, transfers, compares and hunts
- Number commits
- freeze changes
- CPU/VIC bank changes
- result-picker transitions

This gives E2E artifacts a device-side trace instead of requiring failures to be diagnosed only from screenshots.

Logging volume was checked on hardware.

A 437-second `u2@c64u` monitor run produced 96 `MCM` lines with no syslog overflow or failed send reported by the device.

## Keyboard injection and E2E reliability

The hardened monitor tests exposed two existing issues in the shared input path.

### Full 64-event batches

`POST /v1/machine:input` accepts up to 64 keyboard events.

The injected-key ring also had 64 slots, but because it used one slot to distinguish full from empty, it could hold only 63 events.

A valid 64-event API request therefore silently lost its final key while still returning HTTP 200.

The injected ring now has one additional slot while the public API limit remains unchanged at 64.

A unit test pushes all 64 events and verifies that all 64 can be read back.

Hardware verification also covered:

- complete 64-key batches
- a 640-key soak using batches of 64

No keys were lost.

### Queued keyboard timing

With the menu closed, injected taps are passed through the C64 keyboard matrix.

The previous timing held each key for three 20 ms ticks and inserted a two-tick gap, resulting in approximately 100 ms per key.

This is now:

- two ticks held
- one tick released

Measured on the Ultimate 64, this gives approximately 60 ms per key while still presenting the key to at least two keyboard scans.

A one-tick hold was also measured as lossless, but was deliberately rejected because it leaves less timing margin.

This change affects the firmware built by this PR. `u2@c64u` continues to use the keyboard implementation of the C64 Ultimate it is attached to and therefore remains at approximately 100 ms per key.

### Input batching

`machine:input` has two independent limits:

- at most 64 events
- at most 4096 bytes of JSON

These are not interchangeable because different key events serialize to different sizes.

The harness previously split only by a fixed event count. This could still generate requests larger than 4096 bytes.

Batching now uses the actual serialized request size and stops when either API limit would be exceeded.

`MachineApi.send_input` also refuses an oversized body before sending it, with an error that identifies the relevant limit.

A new device-free test suite covers both constraints.

### Input is no longer hidden by retries

The monitor E2E tests previously retried some keypresses until the expected screen appeared.

That could turn a dropped key into a passing test.

The suite now follows a stricter pattern:

1. Send the input once.
2. Poll for the resulting state.
3. Read command arguments back before RETURN.
4. Fail immediately on missing, duplicated or reordered input.

Preparation helpers may retry setup where necessary, but the dedicated input-reliability checks never retype input.

Stress coverage includes:

- repeated hex edits verified independently through `machine:readmem`
- one-, two- and three-byte assembled instructions
- command arguments sent exactly once

Operands change each round so stale memory cannot accidentally satisfy a check.

### Harness polling

Measurement also showed that E2E key handling was performing redundant screen reads.

The polling code now:

- reuses the screen already returned by the change detector
- reuses the final settled screen rather than capturing it again
- shortens poll sleeps when necessary so the required number of samples still fit within the existing time budget

The conditions used to declare a screen settled were not relaxed.

On an Ultimate 64, REST requests per key fell from about 9.3 to:

- about 7.8 when the screen did not change
- about 4.5 when each key changed the screen

The existing settle window remains unchanged because earlier calibration showed that shortening it is where reliability begins to degrade.

## Cross-cutting production changes

The production changes outside `software/monitor` are intentionally limited and live at the layer that owns the behaviour.

| Area | Change | Why it is outside the monitor |
| --- | --- | --- |
| `io/c64/c64.{cc,h}` | Frozen-memory boundary, synchronization, stopped sessions and frozen CIA2 access | These are shared C64 bus/freezer semantics also used by REST memory access |
| `io/c64/keyboard_usb.*` | Correct injected queue capacity and queued-tap timing | These belong to the shared input implementation exposed through `machine:input` |
| `userinterface/ui_elements.{cc,h}` | Optional `UIStringEditPolicy` | Reuses the existing string editor instead of introducing monitor-specific input code |
| `userinterface/userinterface.{cc,h}` | Policy-aware `string_box` and function-key lookup | Generic UI capabilities; existing callers retain their previous behaviour |
| `io/c64/keyboard.h` | Add `KEY_CTRL_X` | The value already existed; only the named constant was missing |

`begin_stopped_session()` / `end_stopped_session()` were moved into the shared C64 layer rather than duplicating equivalent handling for the cartridge backend.

The monitor is currently the only user of `UIStringEditPolicy`. Existing string fields pass no policy and therefore retain their previous behaviour.

`tests/lib/machine.py` also has one shared-harness metadata adjustment because this monitor revision can no longer be identified by the Help screen's removed Exit entry.

### Host tests

`target/pc/linux/machinemonitortest` now covers significantly more than command parsing, including:

- lexical input policy for each command
- Help dimensions
- inclusive ranges
- large range operations
- Assembly and DATA rendering
- Assembly baseline behaviour
- Back-layer behaviour
- editing across memory views
- atomic instruction commits
- popup ordering
- fixed source columns
- trace tokens
- `D` reservation
- exit and non-exit keys
- Transfer relocation

Passing:

```text
machinemonitortest
machinemonitorbookmarkstest
app_space_test
observability_test
route_policy_test
input_batching_test
```

`make host_tests` still fails in the ConfigIO test because it calls the obsolete three-argument `S_read_from_file`.

The relevant ConfigIO files are unchanged from `test-merge`, so this is a pre-existing base-branch failure rather than a regression introduced here.

### Firmware builds

All five firmware targets build in CI.

The Ultimate II+L was tested using the `.u2l` update package produced by this branch's CI.

This matters because the package includes an ECP5 bitstream that is not stored in git. A locally cached bitstream can otherwise be combined with firmware from a different commit.

The Ultimate 64 firmware was loaded over JTAG while retaining its existing FPGA image. This PR contains no U64 FPGA changes, so a full FPGA reflash was unnecessary.

### Application space

The original Ultimate II remains the tightest target:

| | Free |
| --- | ---: |
| Base branch | 3912 bytes |
| This PR | 3520 bytes |
| Difference | -392 bytes |

The firmware still fits.

The original U2 does not contain the monitor itself. Its 392-byte increase comes from the shared C64 and UI changes described above.

The base branch is already below the 1% free-space warning threshold, so this remains worth watching, but this PR does not exceed the image limit.

### Ultimate 64 monitor suite

```text
41 passed
4 skipped
0 failed
```

The skipped checks are target-specific rather than failures.

The run covers, among other things:

- repeated hex edits
- complete instruction commits
- command arguments
- CPU-bank scope
- operand relocation
- source-column placement
- reserved shortcuts
- monitor exits
- freeze toggling

### Combined E2E gate

One uninterrupted gate across all three configured devices completed with:

```text
targets:        c64u, u64, u2@c64u
suite verdicts: 144 OK
suite skips:    0
recoveries:     0
exit code:      0
```

The monitor suite was then rerun concurrently against both devices carrying this branch's firmware:

```text
[u64]     monitor_test: OK (60 checks, 461s)
[u2@c64u] monitor_test: OK (60 checks, 738s)
```

Within the full gate:

- U64 monitor: **41 passed, 4 skipped**
- U2+L monitor: **33 passed, 12 skipped**
- C64 Ultimate monitor: skipped because that device runs a firmware line without this monitor revision

Neither active monitor lane reported a monitor write that failed to land on its first attempt.

Additional reliability measurements include:

- **180/180** REST writes after the frozen-write fixes, versus **170/180** before
- **36/36** monitor edits across the same addresses
- **250/250** four-character Jump arguments, 1000 characters total, with nothing resent
- **210/210** cartridge command arguments in a separate stress run, with every character sent exactly once
- **64/64** events from a full input batch
- **640/640** keys in a batched soak
- **560/560** queued BASIC keyboard inputs on the U64

The monitor read path was also compared against `machine:readmem` over `$8000-$807F`; both agreed on all 128 bytes.

### C64 Ultimate monitor firmware

The C64 Ultimate currently runs a firmware line without this monitor revision.

Its monitor suite is therefore skipped with the machine/version reason.

This is intentional and this PR does not attempt to update that firmware line.

### C64 Ultimate keyboard path used by `u2@c64u`

On `u2@c64u`, injected keys pass through the C64 Ultimate keyboard matrix before the cartridge sees them.

Two effects remain:

- the first `machine:input` request from a fresh harness process can be lost
- under load, an occasional matrix-injected character can be missed

The tests do not hide these by blindly repeating input.

Command fields are read back and mismatches are reported explicitly. The dedicated reliability tests send each character only once.

The relevant keyboard implementation belongs to the C64 Ultimate firmware, which this branch does not build, so changing it is outside the scope of this PR.

### Shared frozen-DMA write behaviour

Longer measurements still show an occasional failure in the shared `C64::dma_transfer_frozen()` path rather than in the monitor's own write logic.

In one ten-round comparison on `u2@c64u`:

```text
monitor:          0/50 writes lost
machine:writemem: 1/50 writes lost
```

This indicates a remaining shared C64 access issue.

The monitor-specific failures found during this work have been fixed, but this PR does not claim to eliminate that broader intermittent completely.

Every first-attempt miss remains visible in the E2E artifacts.

A partially written assembled instruction is always an unconditional test failure because instructions are now sent as one block and must commit completely.

### Ultimate II+L performance

The cartridge remains substantially slower than either standalone machine:

| Target | REST round trip |
| --- | ---: |
| U64 | 13.2 ms |
| C64 Ultimate | 13.7 ms |
| `u2@c64u` | 133.9 ms |

Most of the additional time is spent in cartridge memory access rather than the test harness.

Stress sweeps therefore use fewer repetitions on the cartridge while exercising the same behaviour.


chrisgleissner:
Hi @GideonZ ,

this PR is now ready for review and merge.

The changes have been tested against an Ultimate 64 Elite I and UII+L.

More details in the PR description, as usual.

Thanks
Christian

GideonZ:
Looks pretty good, Christian!  The only thing I am not sure about is the C=+X for reset. This should actually be 'cut' in the same realm of Copy (C) and Paste (V), don't you think?  How about C+-R or something alike? Or C= + <--

GideonZ:
Thinking about this a bit more, I realize that C= + <-- sounds temping, but it doesn't exist under telnet, so we'd have to stick to the control keys that actually do exist. :-/

chrisgleissner:
@GideonZ Good call. I will change it to Ctrl-R by Monday evening.

The only reason I went for Ctrl-X is because Spiffy chose that shortcut. But we are not tied to that.

chrisgleissner:
Hi @GideonZ ,

This is ready for re-review now. 

I have changed the reset shortcut as discussed. I also ran E2E tests against these devices:
- C64U
- U64 Elite I
- U2+L (plugged into C64U)

Finally, I also updated https://github.com/GideonZ/1541u-documentation/pull/27 to be in line with the new code and to be merge-ready and focused by moving out the OpenAPI Yaml document to a new PR.

Thanks for merging
Christian

PR #788, opened 2026-08-15
Apply network service switches live

### LINKED ISSUE #783: Network service switches only take effect one way: enabling is live, disabling needs a reboot
The five switches under *Network Settings -> Services* only work in one direction. Turning a service **on** takes effect within about two seconds. Turning one **off** does nothing until the next reboot.

Noticed while testing the ident service for #782, but it is not specific to it — every one of the five behaves the same way.

## Mechanism

Each service task checks its setting once, in a gate at the top of the task, and never looks at it again:

| Service | Gate |
|---|---|
| Web Remote Control | `software/network/httpd.cc:29` |
| FTP File Service | `software/network/ftpd.cc:230` |
| Telnet Remote Menu | `software/network/socket_gui.cc:214` |
| Ultimate DMA Service | `software/network/socket_dma.cc:379` |
| Ultimate Ident Service | `software/network/socket_dma.cc:499` |

All five are the same shape:

```c
while (networkConfig.cfg->get_value(CFG_NETWORK_<X>_SERVICE) == 0) {
    vTaskDelay(2000 / portTICK_PERIOD_MS);
}
// ... from here on the setting is never read again
```

So the poll makes enabling live, and there is nothing that makes disabling live. The task has already left the loop and is sitting in `accept()` or `recvfrom()`.

## Reproduced

On an Ultimate 64 Elite, firmware 3.14d, over the wired interface, using the ident service because it is the easiest to probe:

```
GET  /v1/configs/Network Settings      ->  "Ultimate Ident Service" : "Enabled"

PUT  /v1/configs/Network Settings/Ultimate Ident Service?value=Disabled
GET  /v1/configs/Network Settings      ->  "Ultimate Ident Service" : "Disabled"

(4 seconds later, UDP probe to port 64)
                                       ->  stillthere,Ultimate-64-Elite-C89085,
                                           *** Ultimate 64 Elite (V1.49) 3.14d ***
```

The setting reads back as Disabled and the service keeps answering. Restoring it to Enabled leaves everything as it was. Nothing was written to flash for this test.

I have only exercised the ident service this way. The other four share the identical gate, so I expect the same, but I have not driven each one and did not want to imply I had.

## Why it is worth fixing

The switches arrived in 3.12 with @Gee-64's "Allow enabling and disabling network services individually", and the natural reading of a switch is that it works both ways. Someone turning off FTP or the web service — plausibly because they are about to put the device somewhere less trusted — currently gets no warning that the service is still running and still accepting connections until they happen to reboot.

## Possible directions

I do not have a preference and this is your architecture, so listing rather than proposing:

1. **Re-check inside the loop.** Each service tests its setting once per accepted connection or received datagram and exits the task when it is disabled. Smallest change, but it puts a config read in the hot path and leaves a task-restart problem to solve for re-enabling.
2. **Close the listening socket on change.** A change hook on the config item closes the socket, so the blocking call returns and the task can go back to its gate. Cleaner behaviour, more plumbing.
3. **Say so in the UI.** Leave the code alone and mark the items as taking effect after a restart, which is honest and costs nothing.

Happy to do the work on any of these if you say which you would prefer, or to leave it if the current behaviour is intended.

@GideonZ @chrisgleissner

### LINKED ISSUE #787: CFG loading again effectuates unrelated configuration groups
cc @GideonZ @chrisgleissner

## Summary

Loading a CFG file currently walks every registered configuration store instead of only stores that successfully loaded an item. Any unrelated store already marked stale can therefore be effectuated and cleared as a side effect.

This behavior was fixed by #765, but the loaded-store tracking was removed by commit 03284765de588788c75b32e8e7d3bbd37782e3d1 while adapting ConfigIO to the centralized loader introduced by commit 7a436e47.

## Reproduction

On a U64 Elite running firmware built from test-merge commit 2e5c9e05:

    ./run-tests -H 192.168.1.62 -s cfg-single-group

The failure is deterministic, including an isolated rerun:

    Expected [Audio Mixer] after CFG load, got [Audio Mixer, SID Sockets Configuration, UltiSID Configuration, SID Addressing, U64 Specific Settings, C64 and Cartridge Settings, Clock Settings, SoftIEC Drive Settings, Printer Settings, Network Settings, Ethernet Settings, WiFi settings, Tape Settings, LED Strip Settings, Drive A Settings, Drive B Settings, Data Streams, Modem Settings, User Interface Settings, Machine Monitor Bookmarks]

## Expected behavior

ConfigIO should retain the centralized S_load_and_effectuate path required by the newer CBM integration, while tracking each unique store for which S_read_store_element returns CFG_LINE_APPLIED. Only those stores should be considered for effectuation.

The existing cfg-single-group E2E test should remain unchanged.

barryw:
## Summary

- Make HTTP, FTP, Telnet, DMA, and Ident service settings take effect in both directions without rebooting.
- Keep socket ownership in each listener task: config callbacks publish atomic desired state, and the owning task closes and recreates its listener.
- Add an Ident enable/disable/re-enable hardware regression test to the shared E2E runner.

## Root cause

Each service checked its setting only while waiting to start. Once enabled, its task left that gate and blocked in accept or recvfrom forever, so changing the setting to Disabled could not affect the live listener.

Closing those sockets directly from the config callback is unsafe in this lwIP configuration because cross-thread netconn close support is disabled. Listener receive timeouts provide bounded wakeups for FTP, Telnet, DMA, and Ident. HTTP uses a loopback connection to wake its select loop, stops accepting, drains active clients, and then recreates the listener when re-enabled.

SO_REUSEADDR permits immediate rebind after shutdown. lwIP loopback is enabled for the internal HTTP wakeup.

## Hardware validation

Tested on an Ultimate 64 Elite over Wi-Fi with the generated U64 firmware flashed:

- Stock 3.14d: Ident enabled successfully; disabling still answered after 3 seconds.
- This branch: Ident enabled in 0.040s, disabled in 1.282s, and re-enabled in 0.816s.
- Registered ident-service-switch suite: 3/3 checks passed.
- Full E2E: 22/23 suites passed in 854 seconds; Wi-Fi remained stable throughout.

The only full-gate failure was cfg-single-group. It fails deterministically in isolation because commit 03284765 removed the loaded-store tracking from #765 while leaving its regression test active. That failure is unrelated to this network change and is tracked with a proposed compatible fix in #787.

## Build and host validation

- Cross-built U2, U2+, U2+L, U64, and U64-II application/lwIP targets on beast.
- Final make u64_no_esp build passed with 307.8 KiB application space free.
- make app_space_test: 36 tests passed.
- runner_policy_test.py: 39 checks passed.
- check_transport_usage.py: 50 files passed.
- Python compilation and git diff whitespace checks passed.

## Draft blocker

- [ ] Resolve #787 and rerun the full E2E gate green before marking this PR ready for review.

Fixes #783

### DIFF
--- a/run-tests
+++ b/run-tests
@@ -339,6 +339,10 @@ SUITES: Sequence[Suite] = (
     # under --mode telnet.
     Suite("e2e", "machine-code-monitor", "tests/e2e/monitor/monitor_test.py",
           "--host @HOST@ --password @PASS@ --timeout @TIMEOUT@ --mode @MODE@", profile=profiles.DEEP),
+    # Reproduces #783: service settings used to start listeners live but could
+    # not stop them again until reboot. Ident is the smallest protocol probe.
+    Suite("e2e", "ident-service-switch", "tests/e2e/network/ident_service_switch_test.py",
+          "-H @HOST@ -p @PASS@ -t @TIMEOUT@"),
     # The device's own FTP server: a name it reports in a listing has to be a
     # name it accepts back for SIZE and DELE.
     # In smoke because nothing else there touches a filesystem. The rest of
--- a/software/network/config/lwipopts.h
+++ b/software/network/config/lwipopts.h
@@ -766,13 +766,13 @@
  * LWIP_NETIF_LOOPBACK==1: Support sending packets with a destination IP
  * address equal to the netif IP address, looping them back up the stack.
  */
-#define LWIP_NETIF_LOOPBACK             0
+#define LWIP_NETIF_LOOPBACK             1
 
 /**
  * LWIP_LOOPBACK_MAX_PBUFS: Maximum number of pbufs on queue for loopback
  * sending for each netif (0 = disabled)
  */
-#define LWIP_LOOPBACK_MAX_PBUFS         0
+#define LWIP_LOOPBACK_MAX_PBUFS         8
 
 /**
  * LWIP_NETIF_LOOPBACK_MULTITHREADING: Indicates whether threading is enabled in
@@ -1011,9 +1011,10 @@
 #define RECV_BUFSIZE_DEFAULT            INT_MAX
 
 /**
- * SO_REUSE==1: Enable SO_REUSEADDR and SO_REUSEPORT options. DO NOT USE!
+ * SO_REUSE==1: Enable SO_REUSEADDR and SO_REUSEPORT options. Listener tasks
+ * use this to rebind immediately after a service is disabled and re-enabled.
  */
-#define SO_REUSE                        0
+#define SO_REUSE                        1
 
 /*
    ----------------------------------------
--- a/software/network/ftpd.cc
+++ b/software/network/ftpd.cc
@@ -212,9 +212,17 @@ static int EndsWith(const char *str, const char *suffix)
 }
 FTPDaemon *ftpd = NULL; 
 
-FTPDaemon::FTPDaemon()
+FTPDaemon::FTPDaemon() : enabled(false)
 {
+    cfg = networkConfig.cfg;
+    cfg->addObject(this);
+    enabled = cfg->get_value(CFG_NETWORK_FTP_SERVICE) != 0;
     xTaskCreate(ftp_listen_task, "FTP Listener", configMINIMAL_STACK_SIZE, this, PRIO_NETSERVICE, &listenTaskHandle);
+}
+
+void FTPDaemon::effectuate_settings(void)
+{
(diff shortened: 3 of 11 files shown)


chrisgleissner:
Hi @barryw , thanks for resolving the conflicts on this PR. Also, could you please move those PRs that are ready to be merged out of 'Draft' status and resolve any unticked checkmarks they may have?

it would be good to get them into a merge-ready state by tomorrow noon (Dutch time) since tomorrow evening is the big merge party and release train departure. :-)

Thanks

PR #789, opened 2026-08-15
Fix configuration store effectuation lifecycle

### LINKED ISSUE #787: CFG loading again effectuates unrelated configuration groups

barryw:
## Summary

- Clear each configuration store's pending-effect flag when its initial settings are applied.
- Preserve unconditional runtime reapplication for reconnect, reset, shared-UI, and UltiCopy paths.
- Mark hardware-backed and lazy/no-observer stores clean once their initial state is known.
- Keep `ConfigIO`'s all-store behavior and make the regression test observe actual effectuation.

## Root cause

Every `ConfigStore` starts with `staleEffect` set, and flash reads set it again. Most boot paths called `effectuate_settings()` directly instead of `ConfigStore::effectuate()`, so settings reached the hardware but the store was never marked clean.

The first later CFG load therefore effectuated every still-stale store. That produced unrelated side effects and made `cfg-single-group` fail only on its first run after boot; the load itself cleared the flags, so subsequent runs passed.

This fixes the lifecycle rather than filtering the CFG loader. A small `effectuate_registered_settings()` helper uses the existing store wrapper for the first application, then directly reapplies clean settings when runtime behavior requires it. Custom stores close the same lifecycle explicitly after reading live hardware or completing their own effectuation.

## Validation

- Reproduced on exact `test-merge` `5bdbb5eb`: the first post-flash one-group CFG load effectuated Audio Mixer plus 17 unrelated stores; the immediate second run passed.
- Exact `e4dd6d3e` firmware, first post-flash `cfg-single-group`: 2/2 checks passed with only Audio Mixer effectuated.
- Exact `e4dd6d3e` full hardware E2E over WiFi: all 22 overlay suites passed in 847 seconds; no recovery was needed.
- ConfigIO host test: 23/23 checks passed.
- Final application builds and space gates passed for U2, U2+, U2+L, U64, U64E2 50T, and U64E2 100T.
- `git diff --check`: clean.

Fixes #787
Unblocks #788

### DIFF
--- a/software/components/config.h
+++ b/software/components/config.h
@@ -334,6 +334,15 @@ public:
 
     virtual void on_edit() { }
     virtual void effectuate_settings() { }
+
+    void effectuate_registered_settings()
+    {
+        if (cfg && cfg->need_effectuate()) {
+            cfg->effectuate();
+        } else {
+            effectuate_settings();
+        }
+    }
 };
 
 extern const char *en_dis[];
--- a/software/drive/c1541.cc
+++ b/software/drive/c1541.cc
@@ -229,7 +229,7 @@ void C1541 :: init(void)
     disk_state = e_no_disk;
 
     mfm_controller->init();
-    effectuate_settings();
+    effectuate_registered_settings();
 
 	xTaskCreate( C1541 :: run, (const char *)(this->drive_name.c_str()), configMINIMAL_STACK_SIZE, this, PRIO_FLOPPY, &taskHandle );
 }
@@ -1210,7 +1210,7 @@ void C1541 :: set_rom_config(int idx, const char *fname)
     uint8_t config_ids[] = { CFG_C1541_ROMFILE0, CFG_C1541_ROMFILE1, CFG_C1541_ROMFILE2 };
     cfg->set_string(config_ids[idx], fname);
     cfg->write();
-    effectuate_settings();
+    effectuate_registered_settings();
 }
 
 void C1541 :: list_roms(ConfigItem *it, IndexedList<char *>& strings)
--- a/software/drive/c1541.h
+++ b/software/drive/c1541.h
@@ -164,6 +164,8 @@ class C1541 : public SubSystem, ConfigurableObject, ObjectWithMenu
     SubsysResultCode_e change_drive_type(t_drive_type drv,  UserInterface *ui);
     SubsysResultCode_e load_dos_from_file(const char *path, const char *filename);
 public:
+    using ConfigurableObject::effectuate_registered_settings;
+
     C1541(volatile uint8_t *regs, char letter);
     ~C1541();
 
--- a/software/drive/c1581.cc
+++ b/software/drive/c1581.cc
@@ -136,7 +136,7 @@ void C1581 :: init(void)
     disk_state = e_no_disk;
 
 	xTaskCreate( C1581 :: run, (const char *)(this->drive_name.c_str()), configMINIMAL_STACK_SIZE, this, PRIO_FLOPPY, &taskHandle );
-    effectuate_settings();
+    effectuate_registered_settings();
     ioWrite8(ITU_IRQ_HIGH_EN, ioRead8(ITU_IRQ_HIGH_EN) | 2);
 }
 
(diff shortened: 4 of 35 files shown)


barryw:
@GideonZ @chrisgleissner, could you sanity-check whether this is the lifecycle fix you had in mind?

I reworked the PR after Gideon's comment: `ConfigIO` still iterates all stores and no longer tracks loaded stores. Initial `ConfigurableObject` applications use a small helper, which routes dirty stores through `ConfigStore::effectuate()` so `staleEffect` is cleared; later reconnect/reset/shared-UI reapplications remain unconditional. Hardware-backed and lazy/no-observer stores are marked clean once their initial state is known. The parser already dirties a store only when a loaded value differs, so an unchanged CFG item does not effectuate a clean store.

Exact `e4dd6d3e` firmware passed the first-post-flash `cfg-single-group` regression with only Audio Mixer effectuated, full E2E 22/22, and all six application/space-gate variants. If this matches the intended direction, an approval would be appreciated; otherwise, please point me at the lifecycle boundary you prefer.

chrisgleissner:
Hi @barryw,

Thanks a lot for this change.

One question. Could we possibly reduce the churn around the rename of the effectuate function? If you look through this PR, there are many renames related to this.

Maybe the original effectuate function could be repurposed to avoid the rename? Not sure if I am missing something.

Basically, you added a decorator function on top of the original function, using a new name. I suggest explorig whether the inverse reduces code churn: turn the original effectuate into a decorator which then delegates to a newly introduced function that has the same behavior as the original effectuate function. 

It may well increase code churn, in which case, just ignore my suggestion. It boils down to how many callers need the old. vs the new behavior. 

Best wishes
Christian

barryw:
Hi @chrisgleissner,

Thanks for the suggestion — I tried the inversion locally. It does preserve the existing `effectuate_settings()` call sites, but because that method is virtual, turning it into the decorator requires introducing a separate implementation hook and renaming every override and its declaration.

The current PR changes 36 files with 87 additions and 62 deletions (149 changed lines). The inverted version changed 51 files with 120 additions and 106 deletions (226 changed lines). There was no behavioral difference; the extra changes were entirely the override/interface rename.

So your caveat applies here: the inversion increases the overall churn. I reverted the experiment and left the current `effectuate_registered_settings()` helper in place.

Thanks again for raising it — it was worth measuring rather than guessing.

chrisgleissner:
Thanks a lot for having tried it @barryw . In that case, your chosen approach is clearly better. 

Also thanks for resolving the merge conflict.

chrisgleissner:
Hi @GideonZ ,

thanks for reviewing this PR and answering Barry's question which he posed on 15 August: 'could you sanity-check whether this is the lifecycle fix you had in mind?' 

Barry mentioned that this PR can only be moved out of draft status once he receives a response from you.

Thanks
Christian

PR #796, opened 2026-08-18
Bound USB key repeat with a negotiated idle rate

### LINKED ISSUE #797: Single short USB key press sometimes repeats until another key is pressed
**Describe the bug**

On a wireless USB keyboard ([Keychron B1P-K1-UK](https://www.keychron.uk/products/keychron-b1-pro-ultra-slim-wireless-keyboard-iso-layout-collection?srsltid=AfmBOor3Q1YqnDo3a0DKxeD8iYDmBVGlGGe2iWgg4whYl_4MPVfySt_R)), a key pressed briefly and once sometimes repeats rapidly. The repetition continues until another key is pressed. It happens roughly once every ten minutes of use.

I have so far not found a way to reproduce it. It has existed on `test-merge` for many months, but I blamed it on my keyboard. Further research shows that it may not be the keyboard after all:
- USB keyboards are configured with a SET_IDLE duration of 0 in `UsbHidDriver::install()`, so they report only when their state changes. 
- `Keyboard_USB` keeps `num_keys` at 1 and keeps auto-repeating for as long as it believes the key is down. 
- The behaviour is consistent with the release report not reaching `Keyboard_USB::process_data()`: nothing bounds the repeat until some later report arrives, which is why the next key press ends it.

**To Reproduce**

1. Plug the keyboard's wireless USB receiver into one of the rear USB ports.
2. Open the menu.
3. Press single keys normally for several minutes.
4. Occasionally one press produces a burst of the same character.

I have not found a way to trigger it deliberately.

**Expected behavior**

A single short key press produces a single character.

 - Hardware type: U64E1 (Ultimate 64 Elite)
 - Firmware Version: 3.15
 - Git Hash: d58a0c96 (current `test-merge`)

**Additional context**

The keyboard is wireless and connects through its own wireless USB receiver, plugged directly into a rear USB port of the Ultimate 64 Elite I, with no hub in between. I have only used a wireless keyboard on this machine; I have not tried a wired one, so I cannot say whether it happens there too.

In the menu the burst is about 12 characters per second, which matches `repeat_speed` and the 20 ms user interface poll.

**Making it more likely (not a reliable trigger)**

None of this reproduces the fault on demand. It is intermittent, and a session set up this way can pass without it appearing at all. These steps only try to load the path the report has to travel, so an occurrence is more likely to fall inside a session I am watching.

1. Plug the keyboard's wireless receiver into a rear USB port, and put a USB stick in the other one.
2. Move the keyboard to the far end of its useful range, so its own link has to retry.
3. Add a mouse, and if you have one a hub, so more interrupt endpoints are polled every frame. The HID interrupt endpoint is polled at 50 Hz and the completions are serviced from one shared USB task, so more devices means more work in the same path.
4. Open the menu and browse a large directory on the stick, so file reads and the keyboard share that path while you type.
5. Tap a single key, once and briefly, then wait a moment. Repeat for several minutes. Cursor keys in a long listing are convenient, because the effect is easy to see.
6. Watch for a tap that produces a run of the same character. In the menu the run starts after about 320 ms and then repeats about 12 times per second, which is `first_delay` of 16 polls and `repeat_speed` of 4 in `Keyboard_USB`.
7. Press any other key to stop it. That is the part that identifies this fault rather than a stuck key: the next report ends the run.

I have not been able to turn this into a deterministic reproduction, and I would not claim the load is what causes it. It is what the code suggests, and the fault shows up during ordinary menu use roughly once every ten minutes without any of it.

chrisgleissner:
Fixes #797.

## Problem

A short press on a USB keyboard can sometimes produce the same character many times.

USB keyboards normally report only when their state changes. If a release report is delayed or lost, the firmware continues to believe that the key is held and auto-repeat runs until another report arrives. In the menu this produces about 12 characters per second.

The existing repeat logic therefore has no way to distinguish a genuinely held key from stale keyboard state.

## Fix

For a supported keyboard, the HID driver now requests a periodic report every 100 ms using `SET_IDLE` with a duration of 25 units of 4 ms.

`Keyboard_USB` uses those reports as a liveness signal:

- while reports keep arriving, a held key repeats normally;
- after more than 300 ms without a report, repeat pauses;
- when reports resume, repeat resumes.

The 300 ms timeout is shorter than the menu's 320 ms initial repeat delay, so a lost release normally stops repeat before another character is emitted.

Periodic reports do not restart the initial repeat delay. Otherwise a report every 100 ms would continuously reset the delay and a genuinely held key would never repeat.

Silence does not prove that a release was lost. A USB delay is indistinguishable from a missing release, so the implementation only treats the repeat state as temporarily untrustworthy and resumes it when reports return.

The stale state is latched once detected. This is required because the 16-bit timer wraps after 65.536 seconds and the same timer value can otherwise make old input appear fresh again.

## Safe HID idle-rate negotiation

The timeout is enabled only when the firmware can rely on one periodic keyboard report stream.

### Keyboard-only interfaces

The 100 ms idle rate is requested only for keyboard-only HID interfaces.

Mouse-only interfaces keep the default infinite idle period. Composite keyboard/mouse interfaces also keep the previous behaviour because `SET_IDLE` with report ID 0 applies to every input report on the interface and would otherwise cause an idle mouse to generate periodic traffic.

### Exactly one keyboard

`Keyboard_USB` receives a merged report stream and does not know which physical keyboard produced each report.

With two keyboards attached, periodic reports from one keyboard could therefore keep stale state from another keyboard looking alive. The timeout is consequently enabled only when exactly one keyboard interface is registered and has successfully negotiated the periodic rate.

All other configurations retain the previous repeat behaviour.

The keyboard registration remains active until `deinstall()` or `relinquish_boot_function()` removes that keyboard from the merged report source. Releasing it earlier from `disable()` would create a window where stale input from an unplugged keyboard remained merged while another keyboard appeared to be the sole active source.

### Verify `SET_IDLE`

A successful `SET_IDLE` request is not treated as proof that the keyboard accepted the duration.

The control transfer returns no data, and some devices may acknowledge the request without storing the requested idle rate. Assuming periodic reports in that situation could make normal auto-repeat stop working.

The driver therefore follows `SET_IDLE` with `GET_IDLE` for report ID 0 and enables the timeout only when the requested value is read back.

## Keyboard buffer handling

The same change also fixes two closely related cases where stale USB input could survive an input reset.

`Keyboard_C64::clear_buffer()` previously cleared only its own ring buffer even though `Keyboard_C64::getch()` also consumes input from `system_usb_keyboard`. Pending USB characters could therefore reappear after the buffer had supposedly been cleared.

A new `Keyboard_USB::clear_pending_input()` operation now:

- removes queued USB characters;
- resets an active repeat to the normal initial delay;
- leaves explicitly injected keys untouched.

Using the existing full `Keyboard_USB::clear_buffer()` here was deliberately avoided because it would also discard injected input that may legitimately be in flight when a popup clears its keyboard buffer.

Resetting the repeat delay means that a key which is still genuinely held must serve the normal 320 ms initial delay again before repeating.

## Boot-function cleanup

`relinquish_boot_function()` now removes every keyboard or mouse function provided by the interface once one of its boot functions is taken over.

Previously, taking over a boot-mouse function could disable the interface while leaving keyboard state from the same report descriptor merged into the global keyboard state. That could leave a key logically pressed with no active source able to release it.

The pure `usb_hid_relinquish_actions()` helper determines which functions must be released.

## Tests

The USB host tests cover the complete behaviour, including:

- lost keyboard release;
- normal repeat with periodic reports;
- keyboard without a negotiated rate;
- repeat resuming when reports return;
- 16-bit timer wrap with stale state already latched;
- keyboard-only, mouse-only and composite idle-rate selection;
- one versus multiple keyboard interfaces;
- keyboard unplug lifecycle;
- successful and unsuccessful `SET_IDLE`/`GET_IDLE` negotiation;
- relinquishing keyboard and mouse functions;
- clearing pending USB input;
- preserving injected input;
- resetting an active repeat after a buffer clear.

A dedicated `keyboardC64ClearTest` verifies the interaction between `Keyboard_C64` and USB input.

### Test results

- `software/io/usb/tests`: 55 + 10 + 21 + 2 + 21 pass
- `software/api/tests`: 11 + 27 pass
- `software/filemanager/tests`: 18 pass
- `target/pc/linux/machinemonitortest`: both binaries pass

The machine-monitor `Keyboard_USB` stub was extended with `clear_pending_input()`.

## Builds

- `u64`: verified artifact produced
- `u64ii`: verified artifact produced
- `u2pl`: application links and size report passes at 995.1 / 1408.0 KiB
- `u2`: application links but remains over its existing size limit - 17,088 bytes over on `test-merge` and 17,696 bytes over with this change

`u2pl` updater packaging fails in this working copy because its packaging rules cannot read Git metadata through the linked worktree. This is unrelated to the change.

The original hardware fault is now reproduced automatically against a real USB keyboard by a separate hardware end-to-end fixture. See "Hardware end-to-end evidence" below.

## Known limits

- USB report state and repeat state are read across the USB and user-interface tasks without locking, consistent with the existing `num_keys` and `delay_count` handling. A report arriving during the check can affect one repeat decision, corrected by the following report. Adding a critical section to the USB event path was considered a greater risk.
- Staleness is latched only when `getch()` observes that the report age exceeded the limit. If the user-interface task itself stalls for longer than the 65.536 second timer period, the wrapped timer can hide the age for up to three idle periods.
- The timeout prevents stale input from producing characters in the user interface. A genuinely lost release report can still leave the corresponding key asserted in the C64 keyboard matrix.

The two software timing limits are documented at `Keyboard_USB::repeatIsLive()`.

## Red/green evidence

Measured on branch head `40bd5729`.

Each behaviour below was individually reverted, its test target run, the file restored with `git checkout --`, and the target run again. The host harness reports failure counts rather than individual test names.

Green baseline for the five affected targets: **55, 10, 21, 2 and 21 tests pass**.

| Reverted behaviour | Target | Red | Green |
|---|---|---:|---:|
| `getch()` repeats without `repeatIsLive()` | keyboard-queue | 3 failed | 21 passed |
| every periodic report restarts the initial delay | keyboard-queue | 3 failed | 21 passed |
| stale state is not latched across timer wrap | keyboard-queue | 1 failed | 21 passed |
| `clear_pending_input()` does not reset the repeat delay | keyboard-queue | 1 failed | 21 passed |
| every selected HID interface receives the 100 ms rate | selection | 1 failed | 21 passed |
| rate accepted without `GET_IDLE` verification | selection | 1 failed | 21 passed |
| timeout enabled with multiple keyboards | selection | 1 failed | 21 passed |
| relinquish releases only the requested boot function | selection | 2 failed | 21 passed |
| `Keyboard_C64::clear_buffer()` does not clear pending USB input | keyboard-c64 | 1 failed | 2 passed |
| `Keyboard_C64::clear_buffer()` clears injected USB input as well | keyboard-c64 | 1 failed | 2 passed |

The working tree was clean before and after the red/green run, and the experiment introduced no commits.

## Hardware end-to-end evidence

The host tests above verify the fixed logic in isolation, against a stubbed USB stack. This section reports a separate, hardware-in-the-loop reproduction of the original field fault against a real USB keyboard, plus a red/green run of the same fixture across both firmware states.

### Fixture

`tests/soak/io/usb/usb_keyboard_repeat_soak_test.py` (doc: `tests/soak/doc/usb-keyboard-repeat-pico.md`) uses a Raspberry Pi Pico 2 W as a real USB boot keyboard, connected to a rear USB port of the Ultimate 64 under test and controlled over Wi-Fi. It injects two fault conditions at the USB report layer:

- `drop_release_once`: the release report is never transmitted, while periodic idle reports of the already-released state continue.
- `silence_after_press`: every report, including periodic idle reports, is suppressed for a fixed window.

Each fault is applied both to a short tap, where the release is lost before any repeat has started, and to a hold long enough that one or more genuine repeats have already fired before the release is lost, which is closer to the shape of the original field report than a fresh tap.

Detection reads the selected row's exact list index off the on-screen menu (not a fuzzy text or colour comparison) and compares the resulting movement count against a baseline measured on the live device, by holding the same key for the same duration with a normal release, rather than against a constant derived by hand from the firmware's timing. A single extra repeat is consequently distinguishable from a runaway.

### Red, unpatched (`d58a0c96`, the commit immediately before this PR's first commit)

Flashing firmware built from `d58a0c96` and running the fixture's own preflight check fails immediately, because `SET_IDLE`/`GET_IDLE` negotiation is absent and the Pico reports `idle_rate: 0`.

Bypassing that one preflight check to run the fault injection and detection directly against the same unpatched firmware forced the runaway repeat in 12 of 12 scenarios (6 short taps, 6 held-then-released), each turning one keystroke into several menu movements:

| Phase | Key | Fault | Hold | Movements produced |
|---|---|---|---:|---:|
| short tap | up | `silence_after_press` | 12ms | 3 |
| short tap | down | `drop_release_once` | 25ms | 2 |
| short tap | up | `drop_release_once` | 35ms | 6 |
| short tap | down | `silence_after_press` | 35ms | 9 |
| short tap | up | `drop_release_once` | 45ms | 6 |
| short tap | up | `silence_after_press` | 60ms | 4 |
| held-then-lost | up | `drop_release_once` | 310ms | 11 |
| held-then-lost | up | `drop_release_once` | 330ms | 10 |
| held-then-lost | up | `silence_after_press` | 640ms | 12 |
| held-then-lost | down | `silence_after_press` | 640ms | 8 |
| held-then-lost | up | `silence_after_press` | 560ms | 11 |
| held-then-lost | up | `drop_release_once` | 400ms | 9 |

Every case exceeded the allowed count of 1 movement; the expected count for a single keystroke, under either fault, on the unpatched firmware.

### Green, patched (this branch, `05d5ba1a`)

Flashing firmware built from this branch and running the fixture's `--profile soak` for 2 hours against the same fault/phase combinations - 2,506 fault iterations plus 501 genuine-hold positive controls (each proving the fix has not simply disabled auto-repeat) - produced 0 failures.

### Reproducing

```sh
# One-time: provisions the Pico only, never the Ultimate 64.
PICO_WIFI_SSID='...' PICO_WIFI_PASSWORD='...' \
  tests/soak/io/usb/usb_keyboard_repeat_soak_test.py --setup-pico

# ~2 minutes, against the Ultimate 64 under test:
./run-tests -H u64 --soak -s usb-keyboard-repeat

# 12 hours:
tests/soak/io/usb/usb_keyboard_repeat_soak_test.py -H u64 --profile soak
```

Full setup and operating instructions, including why the Pico must be the only USB keyboard attached, are in `tests/soak/doc/usb-keyboard-repeat-pico.md`.


chrisgleissner:
Hi @GideonZ ,

this fixes a long-standing issue we've had in the firmware which results in sporadic repetition of shortly pressed keys on a USB keyboard. They repeat very quickly until another key is pressed.

As you can see in the E2E test section of the PR description, the fixed firmware survives a 2h soak of a new E2E test. The same E2E test fails immediately on the unpatched firmware.

I have of course also verified manually that the patched firmware works correctly on my Ultimate 64 Elite I and auto-repeat of held keys in both the menu and elsewhere (e.g. Basic) still works as before.

Thanks for review and merge.

Best wishes
Christian

PR #804, opened 2026-08-25
Add Doom C64U bisection harness and manual E2E suite

chrisgleissner:
Adds a hardware harness for bisecting Doom C64U failures on the Ultimate 64, and a manual E2E suite that runs the published release and checks it works. Nothing in the firmware changes.

## Why

Doom C64U streams level geometry, textures and sprites out of a 16 MB REU every frame while the CPU runs at turbo speed. That makes it the most demanding public exercise of the U64 bus, and it broke between 3.14 and 3.15 (slesinger/doom#5) while every existing suite stayed green.

## What is in the change

- `tests/e2e/io/c64/doom_release_test.py`, registered in `run-tests` as the manual suite `doom-release`. It downloads the current release from GitHub, applies the settings the release needs through the config API, uploads the level image straight into REU memory, starts the game, and checks the engine's own integrity flag, the frame rate, and the picture. About 25 seconds. It is manual because it reaches github.com and changes machine settings, and it puts the settings back.
- `tests/bisect/doom/`, the bisection harness. `bisect_run.py` sweeps every FPGA bitstream in a commit range, deploying each over JTAG, measuring it across several launches, and recording each one to an mp4 with sound. It writes a run directory laid out the way `run-tests --record` writes one, with a Markdown index for people and a JSON index for programs.
- One shared-library change: `RunnersApi.upload` now accepts a `timeout`. The REU image is megabytes and the device only answers once all of it is written, so the ordinary per-request timeout can expire on a request that is progressing normally.

## Two things worth knowing before reading the harness

**The FPGA bitstream is the variable, not the firmware commit.** Only thirteen distinct `external/u64.sof` binaries exist between v3.14 and v3.15, and different branches carry different bitstream lineages, so an ordinary `git bisect` over commits spends most of its rounds on builds whose hardware is identical. The harness enumerates and searches the bitstreams instead.

**Deploying is volatile.** `deploy_commit.sh` programs the FPGA with `quartus_pgm` and downloads the Nios application with `nios2-download`. Nothing is written to flash, so a power cycle always restores the flashed firmware. HDMI stays dark for the session, because the firmware initialises the HDMI transmitter only at application start and it does not come back on an older core; the network video stream still works and is what the harness measures.

## How the picture is judged

The player never moves, so every frame of video should look identical. Some frames still differ on a machine that is working correctly, so "any difference" cannot be the test. What separates the two cases is whether the same pixels keep changing or different ones each time. For each differing frame the harness records which pixels differed, then counts how many *different* such sets it saw. A working machine changes the same pixels repeatedly and stays in single figures; a machine returning corrupt REU data changes a different set nearly every frame.

Calibrated by sweeping every bitstream between v3.14 and v3.15 on an Ultimate 64 Elite, three launches each, about 1000 frames per launch:

| bitstream | frames that differed | different sets of pixels |
|---|---|---|
| v3.14 | none at all, in all three launches | 0 |
| every bitstream up to `f2a14e51` | 0 to 31% | 1 to 3 |
| `c4be69a2` | 100% | 503 |

A build is called broken at 10 or more different sets accounting for at least 30% of the differing frames. Both halves are needed: a couple of different sets appear on healthy machines, and a high proportion means nothing when only three frames differed.

## What the harness found

Two failures, on adjacent bitstream bumps, both changing only the prebuilt binaries `external/u64.sof` and `external/u64_mk2_artix.bit`:

| | first bad commit | last good | effect |
|---|---|---|---|
| Picture corruption | `c4be69a2` | `1e387255`, which carries the `f2a14e51` bitstream | the game runs at full speed, pixels in the 3D view are wrong |
| Game does not start | `77f3b381` | `198353a8` | `mapOK` is 0, the frame counter never advances, the screen is garbled |

`Bus Operation Mode = Compatibility` restores a correct picture on affected firmware but drops the machine to roughly 1 MHz: 0.30 fps against 12.6. It is a diagnostic, not a workaround.

## How it was verified

The E2E suite was run against four deployed commits, red and green in both directions:

| deployed commit | suite result |
|---|---|
| v3.14 | pass |
| `1e387255` | pass |
| `c4be69a2` | fail, on the picture check |
| `77f3b381` | fail, on the engine's integrity flag |

The bisection scripts were run end to end and independently reproduce `c4be69a2`. The three device-free gate suites (`transport-usage`, `runner-policy`, `esp-depends`) pass.

## Measurement traps the scripts handle

Each of these produced a wrong answer before it was fixed, and each is commented where it is handled:

- Frames must be assembled by header offset. Concatenating packet payloads in arrival order turns a reordered packet into a scrambled picture, scrambled differently each time, which is indistinguishable from the defect being hunted.
- The video multicast group is shared between devices on a bench, so packets from another Ultimate get assembled into these frames unless the sender is checked.
- A joystick in port 2 moves the player untouched. The harness reads the player position and voids such a run instead of counting it as a defect.
- The defect can hide for a whole launch, so several launches are required before a build is called clean.

### DIFF
--- a/run-tests
+++ b/run-tests
@@ -264,6 +264,12 @@ SUITES: Sequence[Suite] = (
     # cases. Skips rather than fails when the service is unreachable.
     Suite("e2e", "assembly64", "tests/e2e/io/c64/assembly64_test.py",
           "-H @HOST@ -p @PASS@ -t @TIMEOUT@ --mode @MODE@"),
+    # Manual: downloads a release from github.com and changes machine settings
+    # while it runs. Third-party software, but the most demanding public
+    # exercise of the U64 bus there is, streaming a 16 MB REU every frame under
+    # turbo. Bitstreams that pass every other suite here have broken it.
+    Suite("e2e", "doom-release", "tests/e2e/io/c64/doom_release_test.py",
+          "-H @HOST@ -p @PASS@ -t @TIMEOUT@", manual=True),
     # Measures the keystroke injection every other suite depends on: a lost key
     # fails whichever check was typing at the time and names no rate, so this
     # names one. One batch of the largest size the input API accepts, which is
--- a/tests/README.md
+++ b/tests/README.md
@@ -10,6 +10,7 @@ unit tests live next to their owning code, such as `software/api/tests/`,
 | [`perf/`](perf/) | Timing and throughput benchmarks that measure a number rather than assert a pass/fail outcome. Not the gate. |
 | [`soak/`](soak/) | Time- and load-based checks for leaks, exhaustion, races and transport degradation. Not the gate. |
 | [`lib/`](lib/) | Support code shared by all three categories, plus the two gate checks that need no device. |
+| [`bisect/`](bisect/) | Hardware bisection harnesses: drive a device across many firmware revisions to find the commit a defect arrived in. Not the gate, and not run by `run-tests`. |
 
 `./run-tests <target>` runs the E2E gate. Add `--perf`, `--soak` or `--all`
 to run more, and `-m` to repeat the E2E suites in more than one UI profile:
--- a/tests/bisect/doom/README.md
+++ b/tests/bisect/doom/README.md
@@ -186,17 +186,38 @@ records each with sound, and writes a run directory laid out the way
 `run-tests --record` writes one:
 
 ```
-runs/<stamp>/
-  index.md          the whole run, for a person
-  index.json        the whole run, for a program
-  01-<commit>-<blob>/
-    video.mp4       the game, with sound
-    metadata.json   verdict, per-launch measurements, device identity
-    deploy.txt      what the deploy printed
-    launch-N.txt    what each launch measured
-    frame0.png      the first captured frame
-    mask0.png       which pixels changed, if any did
+runs/20260825-1024/                           when the run itself was made
+  index.md                                    the whole run, for a person
+  index.json                                  the whole run, for a program
+  20260502T142608Z-c4be69a2-bff2d1eb/         one candidate
+    video.mp4                                 the game, with sound
+    metadata.json                             verdict and measurements
+    deploy.txt                                what the deploy printed
+    launch-1.txt launch-2.txt launch-3.txt    what each launch measured
+    frame0.png                                the first captured frame
+    mask0.png                                 which pixels changed, if any did
 ```
+
+#### How a candidate folder is named
+
+`20260502T142608Z-c4be69a2-bff2d1eb` has three parts:
+
+| part | meaning |
+|---|---|
+| `20260502T142608Z` | when that commit was made, in UTC, to the second |
+| `c4be69a2` | the **commit** that was checked out and deployed |
+| `bff2d1eb` | the **FPGA bitstream** that commit carries, which is the thing actually under test |
+
+The two hexadecimal parts are different kinds of identifier and are easy to
+confuse. The first is a commit in this repository. The second is the content
+hash of `external/u64.sof` at that commit, from
+`git rev-parse --short <commit>:external/u64.sof`. Several commits in a row
+normally carry the same bitstream, so the second part is what tells you whether
+two folders really tested different hardware.
+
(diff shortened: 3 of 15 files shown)


chrisgleissner:
## Sweep evidence attached

A full sweep of every FPGA bitstream between v3.14 and v3.15, on an Ultimate 64 Elite I, three launches per candidate, each recorded to video with sound:

**[doom-bisect-sweep-20260825-103503Z.zip](https://github.com/chrisgleissner/1541ultimate/releases/download/doom-bisect-sweep-20260825/doom-bisect-sweep-20260825-103503Z.zip)** (6.3 MB, 107 files)

Open `index.md` inside it. Each candidate folder is named `<UTC commit time>-<commit>-<bitstream blob>` and contains `video.mp4`, `metadata.json`, `deploy.txt`, one `launch-N.txt` per launch, and the first captured frame.

| candidate | date | bitstream | verdict | most frames differing | most different pixel sets |
|---|---|---|---|---|---|
| `v3.14` | 2025-11-11 | `87bd7700` | **GOOD** | 84 | 1 |
| `914b1e8b` | 2025-11-15 | `9ebdab92` | **GOOD** | 172 | 1 |
| `b0b689d5` | 2025-12-20 | `1d2a72a0` | **GOOD** | 0 | 0 |
| `e24d911b` | 2026-01-28 | `98a198fa` | **GOOD** | 0 | 0 |
| `c3c0a13d` | 2026-01-28 | `5dbf965d` | **GOOD** | 4 | 1 |
| `38a65ff4` | 2026-01-30 | `28870ab5` | **GOOD** | 212 | 1 |
| `12a4a556` | 2026-01-31 | `b1f83cbf` | **GOOD** | 12 | 1 |
| `a488e218` | 2026-02-08 | `0f503467` | **GOOD** | 0 | 0 |
| `5f8f64e8` | 2026-02-11 | `4de08349` | **GOOD** | 68 | 1 |
| `bbf0bd32` | 2026-02-15 | `8cbf892a` | **GOOD** | 946 | 2 |
| `10d3b54f` | 2026-03-01 | `244a28fd` | **GOOD** | 4 | 1 |
| `f2a14e51` | 2026-04-11 | `d0d1934a` | **GOOD** | 0 | 0 |
| `c4be69a2` | 2026-05-02 | `bff2d1eb` | **BAD** | 502 | 502 |
| `77f3b381` | 2026-05-11 | `77cf7e40` | **BROKEN** | 0 | 0 |

The player never moves during a measurement, so every frame of video should look identical. A working machine changes the same pixels over and over, so the count of *different* sets of changed pixels stays in single figures however many frames differ. A machine returning corrupt REU data changes a different set nearly every frame.

`c4be69a2` measured 502 different sets over 502 differing frames, a ratio of 1.00: every differing frame was wrong in its own way. Everything before it measured 0 to 2 sets. `77f3b381` does not start the engine at all.

Two earlier boundaries I reported were wrong and this sweep supersedes them. A binary search over a range whose only candidates were `c4be69a2` and `77f3b381` could not have returned anything else, and a threshold of 2 different pixel sets flagged healthy bitstreams that read 3. Every bitstream is now measured rather than assuming the pass/fail is monotonic.

Per @GideonZ, the bitstreams cannot be mapped to `ultimate64` source commits by date, because the repositories are not kept in sync and a checked-in bitstream can be older than the commits around it. This evidence therefore identifies the bitstream binary only, by its content hash.

chrisgleissner:
Hi @GideonZ ,

please find the full bisect results including videos attached above. This was created using the bisect harness of this PR against my Ultimate 64 Elite I.

In terms of videos, best to look at the last two which show the defects nicely. In the table above, these are the last two rows.

Thanks
Christian

chrisgleissner:
Moved back into draft as unlikely to be merged in its current shape as the bisection has done its job.

But maybe the manual E2E is useful to prevent future regressions @GideonZ

chrisgleissner:
Superseded by #811, which carries the E2E suite on its own plus an automated one built on Gideon's `reu_test.tas`. The bisection harness here was tooling for finding the commit, not a regression test, so it is not being merged.

PR #805, opened 2026-08-25
Fix eight firmware defects found by covering the whole REST API

chrisgleissner:
Any `files:create_d64` call took the whole device off the network: no HTTP, no FTP,
no ICMP, power cycle to recover. It shipped because no test ever called that route,
so this PR also adds a suite that calls every REST operation the firmware serves.
That suite found seven more defects.

## The lockup

```
#0  vAssertCalled (fileName="heap_4.c", lineNo=316)
#1  vPortFree (pv=0x41b868)
#2  ArgsURI::ClearAll ()   at software/api/routes.h:224
#3  ArgsURI::~ArgsURI ()
#4  execute_api_v1 ()      at software/api/routes.cc:137
```

The image is written correctly and the response is even sent. The device dies
afterwards, tearing down the request.

`enforce_diskname()` hands a `strdup()`'d string to `args.temporary()`, and
`ClearAll()` released it with `delete`. Since #719 the linker-wrapped `malloc` puts a
size header in front of the pointer the caller gets: `free()` subtracts it, `delete`
is a bare `vPortFree()` and does not. heap_4 is handed an address that is not a block
start, its "is this block allocated" assert fails, and `vAssertCalled` spins with
interrupts off.

Only the four `create_*` routes call `enforce_diskname()`, which is why nothing else
was affected. Firmware 3.14e predates #719 and is fine.

## Fixes

| Defect | Symptom | Fix |
| --- | --- | --- |
| `ClearAll()` freed `strdup()` memory with `delete` | device off the network | free it with `free()` |
| `create_*` returned without closing the `File` when `format()` failed | leaked a `File` and its open handle | scope the file system stack so there is one `fclose()` for both outcomes |
| `C64::freeze()` asserted when already frozen | device off the network; two `menu_button` presses in quick succession reach it | decline the second freeze |
| `C64::unfreeze()` asserted when nothing was frozen | device off the network; a reset issued with the menu open reaches it | decline and clear the flag |
| `GET /v1/configs/<unknown>` | HTTP 200 with an empty body, so a typo reads as an empty category | HTTP 404 |
| `PUT /v1/configs/{cat}/{item}/{value}` | HTTP 400, although `route_configs.cc` implements exactly that form | `value` is `P_OPTIONAL`, so the handler decides instead of the validator |
| `configs:load_from_flash` | reset the caller's connection every time, so success and crash looked identical | re-applying unchanged DHCP settings no longer stops and restarts the client, which cleared the address and made lwIP abort every connection |
| `GET /v1/help` | malformed reply: the status line arrived as body text | `html_response()` frames the reply like `json_response()`, with the header and `Content-Length` in the header buffer |

The three device-killers share one mechanism: `configASSERT` spins with interrupts
off, so an invariant meant to catch a programming mistake takes the machine off the
network instead. Two of the three are reachable with one ordinary REST call issued
twice.

## The coverage gate

`rest-api-coverage` reads the operation list from `doc/api/rest_api_openapi_*.yaml`
when the tree has it (from #801) and from the `API_CALL` macros otherwise, then
**fails while an operation is neither exercised nor written into one of three
tables**: `EXCLUDED`, `HAPPY_ELSEWHERE` or `NEGATIVE_ONLY`. A new endpoint cannot
arrive untested.

Only `machine:poweroff` is excluded. Everything else is called, including `reset`,
`reboot`, `menu_button` and the flash-backed settings operations, each restoring what
it changed. `configs:reset_to_default` without a category is behind
`--allow-global-reset`.

Where a call has an outcome the API can see, the case reads it back: the drive
listing after mounting, the item after writing it, the byte after `writemem`, the
menu screen after `menu_button`. Where there is none, the case says so rather than
letting a 200 stand in for a result.

Each case runs three times, spread across three worker threads so different
operations overlap; `--order sequential` runs the repetitions back to back instead.
Three workers, not four, because the firmware serves `MAX_HTTP_CLIENT = 4` and
saturating the last slot turns queueing into refused connections.

Checking against the OpenAPI documents was worth doing for its own sake. The surface
is 58 operations rather than the 53 handlers the macros list, because the
per-category flash operations and the category and item reads behave differently, and
the documents settled two behaviours I had guessed wrong: `unlink` leaves the disk in
the drive, and an uploaded image needs a filename with an extension.

## Not fixed

`streams:start` fails on a device with two interfaces. `DataStreamer::startStream`
always takes `NetworkInterface::getInterface(0)` and looks the destination up in that
interface's ARP table, so an Ultimate 64 Elite with Ethernet and WiFi answers 404
"Network Host Resolve Error" for the host it is serving HTTP to at that moment.
Pre-existing, and choosing the right interface is a feature change rather than part
of a lockup fix. The case recognises that answer and skips with it as the reason, so
a single-interface device still runs the check.

## Verified

Both suites, both devices, run concurrently. Three consecutive rounds with no
failures. Before the fixes the coverage suite took the device down twice and the
disk-image suite failed on its first create.

| | Ultimate 64 Elite | Ultimate II+L in a C64 Ultimate |
| --- | --- | --- |
| `create-disk-image` | 23/23 in 1.1 s | 23/23 in 2.1 s |
| `rest-api-coverage` | 83/83 in 28 s | 83/83 in 45 s |

77 cases run three times is 462 calls per device per round, each followed by a read
that proves the firmware is still running. No check reaches the ten-second threshold
the E2E README sets.

Skips name their reason: the cartridge builds neither `machine:input`,
`machine:debugreg` nor the data streams, and neither device has a password set for
the authentication case.

Nothing is tied to a particular device. The target comes from `-H` in the usual
`host` or `cartridge@computer` form, the harness's own address is derived from the
route to the device, and the settings item to write and the category to reset are
discovered at runtime. The categories that are never reset are excluded by behaviour,
not by name: User Interface Settings because Interface Type defaults to Freeze, and
the network categories because re-applying them bounces the link the run is using.

`u64`, `u64ii` and `u2pl` all build. The `u2` (MicroBlaze) target does not build in a
linked git worktree for unrelated reasons (`git show HEAD:...` inside the container).

### DIFF
--- a/run-tests
+++ b/run-tests
@@ -212,6 +212,12 @@ SUITES: Sequence[Suite] = (
     # fail for a reason that is not its own.
     Suite("e2e", "create-disk-image", "tests/e2e/api/create_disk_image_test.py",
           "-H @HOST@ -p @PASS@ -t @TIMEOUT@"),
+    # Calls every endpoint the firmware registers, and fails when one is
+    # neither probed nor excluded with a reason. The lockup that prompted it
+    # shipped because no suite called that route at all, so the gate matters
+    # more than any single probe. Runs in about a second.
+    Suite("e2e", "rest-api-coverage", "tests/e2e/api/rest_api_coverage_test.py",
+          "-H @HOST@ -p @PASS@ -t @TIMEOUT@"),
     Suite("e2e", "prg-context-menu", "tests/e2e/filemanager/prg_context_menu_test.py",
           "-H @HOST@ -p @PASS@ -t @TIMEOUT@ --mode @MODE@"),
     Suite("e2e", "browser-long-filename", "tests/e2e/filemanager/browser_long_filename_test.py",
--- a/run-tests
+++ b/run-tests
@@ -206,6 +206,12 @@ SUITES: Sequence[Suite] = (
           "-H @HOST@ -p @PASS@ -t @TIMEOUT@ --test all"),
     Suite("e2e", "input", "tests/e2e/api/input_test.py",
           "-H @HOST@ -p @PASS@ -t @TIMEOUT@"),
+    # Creates every disk-image kind and reads the device back after each one.
+    # A create that answers 200 and then takes the device down is the defect
+    # this catches, so it runs early: a dead device makes every later suite
+    # fail for a reason that is not its own.
+    Suite("e2e", "create-disk-image", "tests/e2e/api/create_disk_image_test.py",
+          "-H @HOST@ -p @PASS@ -t @TIMEOUT@"),
     Suite("e2e", "prg-context-menu", "tests/e2e/filemanager/prg_context_menu_test.py",
           "-H @HOST@ -p @PASS@ -t @TIMEOUT@ --mode @MODE@"),
     Suite("e2e", "browser-long-filename", "tests/e2e/filemanager/browser_long_filename_test.py",
--- a/software/api/route_configs.cc
+++ b/software/api/route_configs.cc
@@ -174,18 +174,32 @@ API_CALL(GET, configs, none, NULL, ARRAY ( { } ))
             list->add(s->get_store_name());
         }
     } else {  // path specified, so the output would list the stores that match the path
+        bool matched = false;
         for(int i=0; i<stores->get_elements(); i++) {
             ConfigStore *s = (*stores)[i];
             if ((path_elements < 1) || pattern_match(args.get_path(0), s->get_store_name())) {
                 emit_store(s, resp->json, args);
+                matched = true;
             }
+        }
+        if (!matched) {
+            // A name no store answers to used to come back as 200 with an empty
+            // body, which reads as "the category exists and holds nothing"
+            // rather than as the typo it is.
+            resp->error("No configuration category matches '%s'.", args.get_path(0));
+            resp->json_response(HTTP_NOT_FOUND);
+            return;
         }
     }
 
     resp->json_response(HTTP_OK);
 }
 
-API_CALL(PUT, configs, none, NULL, ARRAY ( { {"value", P_REQUIRED }} ))
+// 'value' is optional because this route takes it either as the query argument
+// or as a third path element, and the handler below chooses between them and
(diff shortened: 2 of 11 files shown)


chrisgleissner:
Hi @GideonZ ,

here the fix for the bug that was discovered during the Doom bisect. It is completely unrelated to that bisect.
As usual, with red/green TDD proof.

I added E2E tests for all other REST API endpoints. As soon as they all pass, I will update this PR.

Thanks
Christian

Grrrolf:
> `configs:load_from_flash`	reset the caller's connection every time, so success and crash looked identical

When I mentioned this to Gideon when the API first came out, Gideon explained to me that this is as expected. However, it would be nice if it would return a [status code 202](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/202) before loading the config file. That way the client application knows that the command/resource was accepted. FYI: there is nothing wrong with resetting the connection upon loading a complete config file. In essence, a reboot is something one would expect anyway, no.

Lots of APIs just leave the client application hanging.

chrisgleissner:
Thanks. You're right that a load which really changes the address has to drop the connection. And it still does - I did not change that behavior.

What broke here and what I fixed was narrower. `load_from_flash` dropped it even when flash held the values already in force: `ConfigStore::read()` always marks the store stale, so `dhcp_stop()`/`dhcp_start()` ran regardless, and `dhcp_stop()` clears the address, which makes lwIP abort every connection on it. Now DHCP only restarts if it wasn't already running. Nothing reboots either: the Ultimate and the C64 keep running, and FTP and Telnet stay up.

I did try to make 202 work. `json_response()` only fills a buffer, and the socket write happens after the handler returns, so a 202 dies with the connection. Deferring the apply until after the response is written would need a hook in the shared HTTP server, and even then the client may not have received the reply before the connection is aborted. The other two ideas I had, effectuating only stores whose values changed and a config counter the client reads after reconnecting, reach into every config store or into the API surface. None of these changes is small, plus we already have other an open PR #789 that deals with config store effectuation, so I'd rather tackle it as part of a new PR than extend this one.

Issue #802, opened 2026-08-24
UCI SOCKET_READ: a maxlen above 512 silently returns no data, even when the pending datagram is smaller than the request


JC-000:
**Device:** Ultimate 64 Elite, firmware **3.14d**, FPGA 122, core 1.49
**Interface:** UCI (host command interface at `$DF1B-$DF1F`), UDP socket

## Summary

`SOCKET_READ` issued with `maxlen = 1024` never returns data — not even for a 600-byte pending datagram that fits comfortably inside the request. The identical datagram is delivered immediately when `maxlen = 512`. No error bit is set; the read simply never becomes ready.

## Controlled measurement

Same device, same session, same host-side responder, same C64 program. The **only** variable is the `maxlen` passed to `SOCKET_READ`. A host sends the C64 a datagram of the stated size; the C64 polls `SOCKET_READ` and reports what it recovered.

| requested `maxlen` | datagram sizes 600–1280 | polls before giving up | data recovered |
|---|---|---|---|
| **1024** | delivered nothing | 7–10 | `rx_len = 0` |
| **512** | delivered 512 bytes | 1 | `rx_len = 512` |

Every size tested (600, 768, 900, 1024, 1280) behaves this way, including repeats.

The `maxlen = 512` control was run **after** the `maxlen = 1024` run, on a device that had therefore accumulated *more* session time, so this is not an ordering, warm-up, or device-state artifact. (This device does develop an unrelated degradation after roughly five program cycles, which is why the control was deliberately run in that direction.)

Note the 600 and 768 cases in particular: those datagrams are **smaller** than the 1024 requested, so no truncation question arises. They should be delivered whole. They are not delivered at all.

## Two secondary observations

Both with `maxlen = 512` throughout:

1. **Oversized datagrams are truncated and the remainder discarded.** Requests of 600 / 768 / 1024 / 1280 all yield exactly 512 bytes, and no subsequent read ever returns the remainder — 10 of 10 trials. Received tails are byte offsets 496–511, i.e. precisely the first 512 bytes.

   Truncate-and-discard matches POSIX `recvfrom` semantics, so this may well be intended. The gap is that there is **no truncation indication**: POSIX exposes `MSG_TRUNC` for exactly this, whereas here a caller cannot distinguish a complete 512-byte datagram from a truncated 1280-byte one. A protocol with per-datagram integrity (AEAD, checksums) sees only an authentication failure with no way to attribute it.

[...]

PR #806, opened 2026-08-26
Return a socket read over several reply blocks, and report UDP truncation (#802)

chrisgleissner:
Fixes the two defects behind #802 in `NetworkTarget::read_socket()`, and returns a complete UDP datagram of up to 1472 bytes by splitting the reply across UCI response blocks. An end-to-end suite fails against the current firmware and passes with this change on both an Ultimate 64 Elite and an Ultimate II+L.

The original report says that `SOCKET_READ` with `maxlen = 1024` returns no data, while the same datagram arrives with `maxlen = 512`. Testing the command directly through `$DF1B-$DF1F` on real hardware showed that two separate bugs are involved.

### 1. The largest accepted read never ends

A reply consists of a two-byte header followed by the payload. A read of 894 bytes therefore produces a 896-byte reply, exactly filling the response queue.

At this size, an FPGA boundary bug causes the response pointer to stop on the final byte while `response_valid` remains asserted. As a result:

- `DATA_AV` never clears.
- Every additional read returns the same final byte.
- A client that drains the response while `DATA_AV` is set loops forever.

The mechanism is in `fpga/io/command_interface/vhdl_source/command_protocol.vhd`. `response_pointer` stops incrementing once it reaches `c_cmd_if_response_buffer_end`, the last byte of the buffer, while `response_valid` stays asserted for as long as `(response_pointer - buffer_addr) < response_length`. At a length of exactly 896 the pointer saturates at offset 895 and that comparison stays true, so the two never agree.

At 893 bytes, the reply is 895 bytes long and drains normally.

Requests of 895 bytes or more were already rejected with:

```text
82,PARAMETER(S) OUT OF RANGE
```

The pending datagram is not consumed. Because the response queue remains empty, a client that reads two header bytes without first checking `DATA_AV` sees two zeroes and interprets them as a zero-length response. That explains the reported "no data" result at 1024 bytes.

The boundary before this change was therefore:

| Requested length | Result |
|---:|---|
| 0-893 | Reply drains normally |
| 894 | Reply never ends |
| 895 or more | Request is rejected on the status channel |

The length limit is not currently documented, which should be corrected separately in the manual.

### 2. Truncated UDP datagrams are reported as successful

When a UDP datagram is larger than the requested length, `lwip_recvfrom()` copies only the requested bytes and discards the rest of the datagram.

For example, reading a 200-byte datagram with a length of 64 previously returned:

- a header containing 64;
- the first 64 payload bytes;
- `00,OK`.

A genuine 64-byte datagram produced exactly the same result. The remaining 136 bytes from the larger datagram are gone, and the next read reports no data.

TCP does not have this problem because it is a stream. The same 200 bytes requested in blocks of 64 are returned as 64, 64, 64 and 8 bytes without loss.

For protocols that require the complete datagram, such as the reporter's WireGuard implementation, silent truncation makes a valid but oversized datagram indistinguishable from a corrupted one.

## What this change does

The firmware change is limited to `NetworkTarget::read_socket()` and the three fields of continuation state it needs.

### Return a datagram larger than one response block

`SOCKET_READ` now accepts lengths up to 1472 bytes. That is the largest UDP payload that can reach the device: 1500 bytes of Ethernet MTU less 20 bytes of IPv4 header and 8 bytes of UDP header. `IP_REASSEMBLY` is 0 in `software/network/config/lwipopts.h`, so a larger datagram arrives as fragments that are dropped before any socket sees them.

A reply whose payload does not fit one response block is handed back over the command interface's existing `Data More` mechanism, rather than by enlarging the FPGA response queue. The first block carries the two-byte header and the first 893 payload bytes; the continuation block carries the remaining payload and nothing else. The two-byte header is the total number of payload bytes the command returns, not the number in the first block, so a client that concatenates the blocks holds exactly the reply a large enough queue would have delivered in one.

For a 1420-byte datagram read with `maxlen = 1472`, the blocks measured on hardware are 895 and 527 bytes.

The whole payload is received once, while `SOCKET_READ` runs, into a buffer that outlives the call. A continuation block never goes back to the socket, because a second receive on a datagram socket would take the next datagram rather than the rest of this one.

### Keep every response block short enough to drain

No block the firmware builds exceeds 895 bytes:

- the first block is `2 + min(payload, 893)` bytes, so at most 895;
- a continuation block is at most 895 payload bytes;
- the largest accepted read is 1472 bytes, which is two blocks of 895 and 579.

Requests above 1472 continue to return:

```text
82,PARAMETER(S) OUT OF RANGE
```

### Report UDP truncation

`lwip_recv()` is replaced with `lwip_recvmsg()` using a single `iovec`.

For UDP sockets, `lwip_recvmsg()` reports the original datagram length and sets `MSG_TRUNC` when the supplied buffer is too small. The command still copies no more than the requested number of bytes, but now reports:

```text
04,DATAGRAM TRUNCATED: <original length>
```

For TCP sockets, the new call resolves to the same underlying TCP receive path as before.

Truncation and continuation compose. A 1200-byte datagram read with `maxlen = 1000` returns exactly 1000 payload bytes, split over two blocks, with a header of 1000 and status `04,DATAGRAM TRUNCATED: 1200`. The header stays the number of bytes the reply carries; the datagram's original length is reported only in the status text.

## Continuation state and abort

`NetworkTarget` holds three fields: the total payload length, how much of it has already been placed in a block, and the status to report on the block that ends the reply. There is no per-socket state.

- `parse_command()` drops any pending reply before it runs a command, so no path into this target can hand a client bytes left over from an earlier read.
- `get_more_data()` serves the next block from the buffer and clears the state when it emits the last one. With nothing pending it returns the existing `03,MORE DATA NOT SUPPORTED`, which is what a client asking for a block that was never announced used to get in every case.
- `abort()` discards the state.

`abort()` also changes signature. `CommandTarget` declares `abort(int)`, and `NetworkTarget` declared `abort(void)`, which overloads rather than overrides. A client's abort therefore reached the base class no-op and never reached this target. It now takes the `int` and overrides.

The status channel carries one result per command rather than one per block: an intermediate block leaves the status queue empty, and the block that ends the reply carries the text. A client that appends the status of every block ends up with exactly the text a single-block reply would have given it.

The command interface only accepts a command while it is idle, so a reply that is still being handed out cannot be interrupted by a new one. `parse_command()` clears the state anyway, which covers the one path that ends an exchange without an abort: a command sent with the no-reply bit set, where `run_task()` writes `HANDSHAKE_RESET` instead of a reply.

## Backwards compatibility

Requests up to 893 bytes are unchanged. The reply is a single block announced as `Data Last`, byte for byte what it was before, and a client that does not understand `Data More` never sees it.

Requests above 893 did not previously constitute a working operation, so introducing `Data More` for 894-1472 does not break a single-block contract that anyone could have relied on:

- 894 produced a reply that never terminates, which no protocol-compliant client can consume;
- 895 and above were rejected on the status channel.

The reference client, `ultimateii-dos-lib`, requests 892 bytes (`DATA_QUEUE_SZ - 4`) and is unaffected.

There are two intentional differences in behaviour:

1. A request of 894 to 1472 bytes now returns data across two blocks, where 894 used to hang and 895 and above were rejected.
2. A truncated UDP datagram now returns status `04` instead of incorrectly returning `00,OK`. Status code `04` was previously unused, so a client that ignores the status text receives the same payload bytes as before.

The header format is deliberately unchanged. `uci/network_target.rst` defines it as the number of bytes actually read, and existing clients use that value to locate the end of the returned payload.

## TCP

`read_socket()` does not look at the socket type, so a stream read is split over blocks by the same code. A TCP read with a length above 893 now returns up to that many bytes in one command instead of being rejected. Measured on both devices: 1420 bytes sent over TCP and read with `maxlen = 1472` come back in one command as blocks of 895 and 527.

Nothing else about TCP changes. `MSG_TRUNC` is never set on a stream, so a stream read is never reported as truncated, and repeated reads of less than is pending still lose nothing.

The 1472-byte ceiling applies to TCP as well, although a stream has no datagram size to derive it from. That is deliberate: the limit is what one receive buffer holds, TCP previously stopped at 893, and raising the stream case further would need a larger buffer and more blocks without a bound to justify the number.

### Report the original datagram length in the reply header

This would avoid introducing a new status code, but it would change the meaning of the header. A header larger than the returned payload could also cause existing clients to read beyond the available data.

The header therefore remains the number of bytes actually returned.

### Clamp oversized requests

A request above the supported limit could be silently reduced to 1472 bytes. This was rejected because the interface already reports an explicit out-of-range error and leaves the pending datagram untouched.

Clamping would silently change the meaning of the caller's request and could cause the datagram to be truncated instead.

### Fix the response queue in the FPGA

The endless reply originates in the shared FPGA command interface. Fixing it there would allow a 896-byte block.

However, that requires a new bitstream for every affected product, and devices already in the field would retain the bug until reflashed. There is also no existing testbench for this block.

The software invariant that no block exceeds 895 bytes protects all devices immediately. The underlying FPGA issue should still be fixed separately.

### Support datagrams above 1472 bytes

Datagrams up to 1472 bytes reach the device, while 1473-byte datagrams do not, because IP reassembly is disabled in lwIP. Enabling it affects every network service on the device rather than only `SOCKET_READ`, so it is outside the scope of this change. Fragmented datagrams and 64 KiB datagrams remain unsupported.

### Per-socket continuation state

Continuation is transaction-level rather than per-socket. The command interface serves one exchange at a time and will not accept a command while a reply is outstanding, so a second socket cannot have a reply in flight concurrently. Per-socket state would add bookkeeping with nothing to track.

### Fix the similar HTTP target case here

Code inspection shows that `http_target.cc` can also produce response blocks of exactly 896 bytes and may therefore reach the same FPGA boundary.

That path has not yet been measured on hardware. It is being reported separately rather than expanding this focused fix based only on code inspection.

## How it was tested

The end-to-end suite `tests/e2e/io/command_interface/net_target_test.py` is registered in `run-tests` as `uci-net-target`.

Every scenario runs twice, once down each way of reaching the registers:

- **rest**: `machine:readmem` and `machine:writemem`, which the Ultimate performs as DMA cycles on the cartridge bus.
- **native**: a small resident 6502 agent, `uci_agent.asm`, assembled on the fly with `tools/64tass` and started over `runners:run_prg`. It writes commands to `$DF1D` and drains the queues through ordinary 6510 loads from `$DF1E` and `$DF1F`.

The native route is how a program actually uses the interface. Its register accesses are microseconds apart rather than tens of milliseconds, so it exercises a timing envelope the DMA route cannot reach. The agent already followed a `Data More` reply through its blocks; this change adds an option that abandons a reply after its first block, so both routes can make the same abort measurement.

The host acts as the device's network peer:

1. It opens a UDP socket and TCP listener.
2. The device opens sockets back to the host.
3. The device sends a packet so the test can discover its ephemeral source port.
4. The host returns payloads in which every byte identifies its own offset.
5. The suite verifies the returned length, payload, status and queue state.

Because every byte names its own offset, a continuation block written from the wrong offset, or one that repeats or drops bytes, shows up as wrong data rather than as a plausible reply.

### Scenarios

| Scenario | What it establishes |
|---|---|
| `read-length-limit` | The accepted maximum, found by bisection, is 1472; 1473 is refused on the status channel and the boundary is sharp |
| `reply-blocks-are-drainable` | The first block of a read of 893, 894 and 1472 bytes ends, and no block reaches 896 |
| `datagram-spans-reply-blocks` | Datagrams of 893, 894, 1420 and 1472 bytes arrive complete with `00,OK`; 893 arrives in one block, the rest in blocks that each end |
| `datagram-size-ceiling` | Datagrams up to 1472 bytes reach the device and larger ones do not |
| `udp-truncation-is-detectable` | A 200-byte datagram read at 64 is distinguishable from a genuine 64-byte one |
| `truncation-spans-reply-blocks` | A 1200-byte datagram read at 1000 returns exactly 1000 bytes over two blocks with `04,DATAGRAM TRUNCATED: 1200`, and the remainder is gone |
| `tcp-read-is-lossless` | Stream reads below and above one block lose nothing, cross a block boundary, and are never reported as truncated |
| `oversize-request-keeps-the-datagram` | A refused request of 1473 does not consume the pending datagram |
| `multi-block-state-does-not-leak` | A reply abandoned after its first block returns the interface to Idle, and the next command is answered normally; so is the command after a reply that completed |

### Red/green hardware results

Both halves compare builds of this branch: one containing the suite but not the firmware change, one containing both.

| Device | Routes | Without the change | With the change |
|---|---|---|---|
| Ultimate 64 Elite, firmware 3.15, FPGA 123, core 1.4E | rest and native | 6 of 9 scenarios fail on each route, 176s | 73 checks pass, 330s |
| Ultimate II+L, firmware 3.15, FPGA 123, in a C64 Ultimate | rest only | 5 of 9 scenarios fail, 157s | 39 checks pass, 313s |

The failing scenarios without the change are `read-length-limit`, `reply-blocks-are-drainable`, `datagram-spans-reply-blocks`, `truncation-spans-reply-blocks`, `multi-block-state-does-not-leak`, and on the Ultimate 64 also `tcp-read-is-lossless`.

The three that pass both before and after are `datagram-size-ceiling`, `udp-truncation-is-detectable` and `oversize-request-keeps-the-datagram`. They bound what the change is allowed to alter: the truncation reporting from the earlier commits on this branch still works, the network stack's own datagram ceiling is unchanged, and a refused request still leaves the pending datagram alone.

The Ultimate II+L red run was recorded before the `tcp-read-is-lossless` scenario gained its stream-above-one-block phase, which is why it shows five failures rather than six. That phase fails without the change for the same reason on both devices: a read of 1472 bytes is rejected.

### The native route is skipped on cartridge targets

On a `cartridge@computer` target the two routes reach different devices, which was measured rather than assumed.

Running both routes against a U2+L in a C64 Ultimate showed the REST route reporting the patched cartridge and the native route reporting an unpatched interface. The host's own command interface claims `$DF1B-$DF1F` from its 6510, so a program running there measures the host rather than the cartridge under test. Disabling the host's command interface does not hand the range over either: the 6510 then reads `$FF`.

DMA is the asymmetry. A `readmem` issued by the host does reach the cartridge, which is why the REST route works on these targets.

The suite therefore runs the native route only on a target that is a single machine, and records the reason when it skips it.

### Other checks

- `uci-targets`, the other suite that shares the `tests/e2e/lib/uci.py` transport, passes unchanged on the Ultimate 64: 37 checks in 33s. The REST response drain now reads `$DF1C-$DF1E` in one `readmem` instead of polling and reading separately, which halves its cost; `$DF1D` has no read side effect in `command_protocol.vhd`, so the drain still gates each byte on `DATA_AV`.
- `network_target.cc` compiles for all four toolchains: Nios II for u64 and u2plus, RISC-V for u64ii and u2pl. On the Ultimate 64 the change costs 504 bytes of text and 12 bytes of BSS.

## Known limitation

If a client sets the data-accepted bit and the abort bit close enough together that the firmware sees both in one interrupt, `CommandInterface::run_task()` runs the abort and then still calls `get_more_data()`, which publishes an empty `Data Last` reply carrying `03,MORE DATA NOT SUPPORTED` instead of leaving the interface idle. This is in the shared command interface, applies to every target, and produces exactly the same reply as before this change, because `get_more_data()` returned that unconditionally. It is not addressed here.

### DIFF
--- a/run-tests
+++ b/run-tests
@@ -247,6 +247,12 @@ SUITES: Sequence[Suite] = (
     # and the file system.
     Suite("e2e", "uci-targets", "tests/e2e/io/command_interface/uci_targets_test.py",
           "-H @HOST@ -p @PASS@ -t @TIMEOUT@ --test all"),
+    # The UCI network target's READ_SOCKET: what the requested length does, and
+    # what happens to a datagram larger than it (#802). This host is the
+    # device's network peer, so it binds two ports and needs the device to be
+    # able to reach it; the scenarios that need that skip where it cannot.
+    Suite("e2e", "uci-net-target", "tests/e2e/io/command_interface/net_target_test.py",
+          "-H @HOST@ -p @PASS@ -t @TIMEOUT@ --test all"),
     # Two checks that are genuinely about the Telnet transport itself (a
     # concurrent poll-mode connection, per-keystroke output volume) only run
     # under --mode telnet.
--- a/software/io/network/network_target.cc
+++ b/software/io/network/network_target.cc
@@ -237,18 +237,37 @@ void NetworkTarget :: read_socket(Message *command, Message **reply, Message **s
 	uint8_t socketnr = command->message[2];
 	uint32_t length = ((uint32_t)command->message[3]) | (((uint32_t)command->message[4]) << 8);
 
-    if (length > (CMD_MAX_REPLY_LEN-2)) {
+    // The reply is a two byte header plus the payload, and the transport can
+    // only deliver a block shorter than the response queue: the FPGA stops the
+    // response pointer at the last byte of the buffer while it still reports
+    // data available, so a block of exactly CMD_MAX_REPLY_LEN never ends and a
+    // client reading on that flag never stops.
+    if (length > (CMD_MAX_REPLY_LEN-3)) {
         *status = &c_status_param_out_of_range;
         *reply = &c_message_empty;
         return;
     }
 
     *reply = &data_message;
-	int ret = lwip_recv(socketnr, &data_message.message[2], length, 0);
+    // lwip_recvmsg rather than lwip_recv: on a datagram socket it returns the
+    // length of the datagram itself and sets MSG_TRUNC when it did not fit,
+    // where lwip_recv reports only what was copied and drops the rest without
+    // a trace. On a stream socket the two behave the same way.
+    struct iovec iov;
+    struct msghdr msg;
+    iov.iov_base = &data_message.message[2];
+    iov.iov_len = length;
+    memset(&msg, 0, sizeof(msg));
+    msg.msg_iov = &iov;
+    msg.msg_iovlen = 1;
+    int ret = lwip_recvmsg(socketnr, &msg, 0);
+    // The header is the number of bytes this reply carries, which is what the
+    // network target document specifies and what existing clients read.
+    int copied = (ret > (int)length) ? (int)length : ret;
     data_message.length = 2;
     data_message.last_part = true;
-    data_message.message[0] = (ret & 0xFF);
-    data_message.message[1] = (ret & 0xFF00) >> 8;
+    data_message.message[0] = (copied & 0xFF);
+    data_message.message[1] = (copied & 0xFF00) >> 8;
 
     // printf("Reading %d bytes from socket %d resulted in %d\n", length, socketnr, ret);
 	if (ret == 0) {
@@ -257,7 +276,18 @@ void NetworkTarget :: read_socket(Message *command, Message **reply, Message **s
 		return;
 	}
 	if (ret > 0) {
-		data_message.length = ret + 2;
+		data_message.length = copied + 2;
+		if (msg.msg_flags & MSG_TRUNC) {
+			// The datagram was longer than the caller asked for and the rest is
+			// gone. Reporting that on the status channel leaves the reply itself
+			// unchanged, so a client that does not look is no worse off than
+			// before, and one that does can tell a short datagram from a
+			// truncated one.
+			*status = &status_message;
+			sprintf((char *)this->status_message.message, "04,DATAGRAM TRUNCATED: %d", ret);
+			this->status_message.length = strlen((char *)this->status_message.message);
+			return;
+		}
 		*status = &c_status_ok;
 		return;
 	}
--- a/software/io/network/network_target.cc
+++ b/software/io/network/network_target.cc
@@ -29,6 +29,7 @@ NetworkTarget::NetworkTarget(int id)
     command_targets[id] = this;
     data_message.message = new uint8_t[CMD_MAX_REPLY_LEN];
     status_message.message = new uint8_t[CMD_MAX_STATUS_LEN];
+    discard_read_reply();
 }
 
 NetworkTarget::~NetworkTarget()
@@ -41,6 +42,12 @@ NetworkTarget::~NetworkTarget()
 void NetworkTarget :: parse_command(Message *command, Message **reply, Message **status)
 {
     NetworkInterface *interface;
+
+    // The command interface only accepts a command while it is idle, so a
+    // reply that is still being handed out cannot normally be interrupted by
+    // one. Dropping it here as well means no path into this target can hand a
+    // client bytes left over from an earlier read.
+    discard_read_reply();
 
     switch(command->message[1]) {
         case NET_CMD_IDENTIFY:
@@ -237,32 +244,37 @@ void NetworkTarget :: read_socket(Message *command, Message **reply, Message **s
 	uint8_t socketnr = command->message[2];
 	uint32_t length = ((uint32_t)command->message[3]) | (((uint32_t)command->message[4]) << 8);
 
-    // The reply is a two byte header plus the payload, and the transport can
-    // only deliver a block shorter than the response queue: the FPGA stops the
-    // response pointer at the last byte of the buffer while it still reports
-    // data available, so a block of exactly CMD_MAX_REPLY_LEN never ends and a
-    // client reading on that flag never stops.
-    if (length > (CMD_MAX_REPLY_LEN-3)) {
+    // A complete unfragmented IPv4 UDP datagram is the most this command can
+    // ever return, so that is what it accepts. Anything larger cannot arrive.
+    if (length > NET_MAX_SOCKET_READ) {
         *status = &c_status_param_out_of_range;
         *reply = &c_message_empty;
         return;
     }
 
     *reply = &data_message;
+    // The whole payload is received here, into a buffer that outlives this
+    // call, so that a reply spanning several blocks never has to go back to
+    // the socket. A second receive on a datagram socket would take the next
+    // datagram rather than the rest of this one.
+    //
     // lwip_recvmsg rather than lwip_recv: on a datagram socket it returns the
     // length of the datagram itself and sets MSG_TRUNC when it did not fit,
     // where lwip_recv reports only what was copied and drops the rest without
     // a trace. On a stream socket the two behave the same way.
     struct iovec iov;
     struct msghdr msg;
-    iov.iov_base = &data_message.message[2];
+    iov.iov_base = buffer;
     iov.iov_len = length;
     memset(&msg, 0, sizeof(msg));
     msg.msg_iov = &iov;
     msg.msg_iovlen = 1;
     int ret = lwip_recvmsg(socketnr, &msg, 0);
     // The header is the number of bytes this reply carries, which is what the
-    // network target document specifies and what existing clients read.
+    // network target document specifies and what existing clients read. On a
+    // reply that spans blocks it is still the total, not the part in the first
+    // block, so a client that concatenates the blocks gets exactly the reply a
+    // large enough response queue would have delivered in one.
     int copied = (ret > (int)length) ? (int)length : ret;
     data_message.length = 2;
     data_message.last_part = true;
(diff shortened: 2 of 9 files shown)


chrisgleissner:
Hi @GideonZ ,

this fixes an UCI issue raised by a user in a minimal invasive manner. I considered various alternative implementation options before doing so.

I red/green tested the fix intensely against both U64 and U2. Please see full details above, both on issues, fixes, alternative fix options considered, and test approach / results.

Thanks
Christian

chrisgleissner:
Hi @GideonZ ,

as per your comment on the issue which this PR fixes, I will change the implementation to support UDP payloads up to 1472 bytes, in line with a typical MTU size of 1500 bytes:

```
1500 Ethernet MTU
- 20 IPv4 header
-  8 UDP header
= 1472 UDP payload
```

Thanks
Christian

chrisgleissner:
Hi @Gideon,

this is now done as requested: I have updated for the `Data More` request.

`SOCKET_READ` now accepts up to 1472 bytes, the maximum unfragmented IPv4 UDP payload, and splits larger replies using the existing continuation mechanism. A 1472-byte reply is sent as blocks of 895 and 579 bytes, so the firmware never crosses the 896-byte FPGA boundary.

The socket is read only once, so requesting the continuation cannot consume the next datagram. Pending continuation data is discarded when another command arrives or the client aborts. I also fixed `abort(int)`, which previously had the wrong signature and was never called through the base class.

Truncation reporting remains unchanged. For example, reading 1000 bytes from a 1200-byte datagram returns 1000 bytes across two blocks with `04,DATAGRAM TRUNCATED: 1200`.

Reads up to 893 bytes remain byte-for-byte unchanged. Larger reads did not work previously: 894 never completed, while 895 and above were rejected.

The expanded suite covers nine scenarios, including maximum-size datagrams, truncation, abandoned replies and stream reads.

| Device | Without fix | With fix |
|---|---:|---:|
| Ultimate 64 Elite | 6 of 9 scenarios fail | 73 checks pass |
| Ultimate II+L in C64 Ultimate | 5 of 9 scenarios fail | 39 checks pass |

The shared `uci-targets` suite also passes unchanged.

Thanks for review and merge
Christian

PR #811, opened 2026-08-28
Add an automated E2E regression test for the REU start-of-transfer defect

chrisgleissner:
Adds two E2E suites for the REU defect fixed in 59594060 ("Updated REU - no initial delay on U64/U64E2, because of Turbo Mode"). No firmware change.

## The defect

Starting an REU transfer has to stop the 6510 in the same cycle as the write to $DF01. Before the fix the core went through a delay state before asserting DMA. At 1 MHz that is invisible, because the CPU is halted before the next instruction can run. Above 1 MHz it is not: the instructions after the command write execute while the transfer is already in progress. Real REU code puts the register setup for the next transfer there, so those stores land on the base and length registers of the transfer still running, and the data that comes back is not the data that went out.

## `reu-turbo`, automated, about 22 seconds

The stimulus is `software/6502/unsorted/reu_test.tas`, the program the fix was made against, not a copy of it. `tests/e2e/io/c64/reu_turbo_test.py` assembles that file with the repository's own 64tass and starts it over `runners:run_prg`. The sequence under test is the part that must not change: the store to $DF01 is followed straight away by the register writes for the next transfer, with nothing in between. A copy of the program in `tests/` would be free to drift away from that, which is the whole reason the suite reaches for the original.

The suite runs the program twice and makes three measurements, in this order:

1. at 1 MHz. This is the control. A failure here is an REU that is off or broken outright, not the defect above, and it also rules out interference from the suite's own polling, since the host reads the result block over `machine:readmem`, which is a DMA cycle on the same bus.
2. at the highest CPU speed the machine offers. This is the guard.
3. that the fast run really was fast. Without it, a machine that ignored the CPU speed setting would run measurement 2 at 1 MHz and pass it for the wrong reason.

The REU, badline and turbo settings are read before they are changed and written back however the run ends.

### What the E2E build adds, and why the simulation build is untouched

`reu_test.tas` gains a `.weak` symbol `E2E`, defaulting to 0, and two `.if E2E` blocks. Assembled with no define it produces the same 179 bytes it produced before this change, verified by comparing the two binaries, so the VICE and nvc simulation runs are unaffected. The suite assembles it with `-D E2E=1`, which adds exactly two things, both of which a simulation must not have:

- **A settle wait.** The machine holds the CPU and the REU at reduced speed for the first seconds after a reset, on hardware but not in simulation. `runners:run_prg` resets, loads and starts in one step, so without the wait the run labelled "full speed" spends most of itself at 1 MHz, which is the one speed at which this defect cannot appear. Measured on the fixed core: with the wait the 48 MHz run takes 0.27s against 5.51s at 1 MHz, a factor of 20.6; with the wait removed it takes 2.76s, a factor of 2.0. The guarded block waits 250 frames on raster line $80, which is 5.0s on PAL and 4.2s on NTSC, counted in frames so it costs the same wall clock however fast the CPU is running. Five seconds of frames is not worth simulating, which is why it is behind the define.
- **A result block in RAM at $C000.** `machine:readmem` reads RAM, not I/O: on this device `$D012` reads `ff` twice in a row and `$D020` reads `00`. The `$D7FF` the program writes for VICE is a SID register mirror on real hardware, so neither it nor the border colour can be read back over REST. The block carries a started marker, an end-of-wait marker so the host times the rounds rather than the wait, and on a mismatch the round, the byte index and the value that should have been there.

## `doom-release`, manual, about 27 seconds

`tests/e2e/io/c64/doom_release_test.py`, carried over from the closed #804. It downloads the current Doom C64U release from github.com, applies the settings the release documents through the config API, uploads the level image into REU memory, starts the game, and checks the engine's own REU integrity flag, the frame rate and the picture. Manual because it reaches github.com and changes machine settings, both of which it puts back.

It is the wider test rather than the necessary one: it exercises a real program streaming out of a 16 MB REU with the CPU in turbo, where `reu-turbo` exercises one transfer sequence. Both were measured red on the core before the fix and green on the core after it, so `reu-turbo` is the one the gate needs.

Two shared-library changes come with it, both additive and both no-ops for existing callers:

- `RunnersApi.upload` takes a `timeout`. The REU image is megabytes and the device answers only once all of it is written, so the ordinary per-request timeout can expire on a request that is progressing normally. `RestClient.request` already defaulted `timeout` to `None` and resolves it as `self.timeout if timeout is None else timeout`, so an existing `upload` call behaves exactly as before.
- `assemble()` takes assembler defines, which is how the suite selects the E2E build. With no defines it builds the same 64tass command line as before; `uci_agent.asm` still assembles through it.

## How it was verified

Both `external/u64.sof` bitstreams were deployed to an Ultimate 64 Elite over JTAG, with the same Nios application downloaded after each, so the bitstream was the only variable. The pre-fix bitstream is the one at 6b5ffc21, the commit before the fix, and reports `core_version` 1.4E; the fixed one at 59594060 reports 1.4F.

| suite | core 1.4E, before the fix | core 1.4F, with the fix |
|---|---|---|
| `reu-turbo` | FAIL: `byte $00 of the block came back wrong in round 0 of 256` at 48 MHz, with the same program clean at 1 MHz in 5.48s | PASS, 3 consecutive runs. 5.5s at 1 MHz against 0.27s at 48 MHz, 20.6x to 21.0x |
| `doom-release` | FAIL at "the engine verified its REU image": `mapOK=0` | PASS, 12.29 fps, no corrupted frames |

What was reverted to show each check can fail:

- Swapping the bitstream is what makes measurement 2 fail, and it is the whole point of the suite. The failure it reports names the round and the byte.
- Deleting the settle wait from the E2E build makes measurement 3 fail: `the run at 48 MHz was only 2.0x the run at 1 MHz`. It does **not** make measurement 2 stop failing on core 1.4E, which was measured rather than assumed: the suite still fails at round 0 without the wait. The wait is there so that the guarded run actually happens in turbo and so that measurement 3 means something, not because the defect is otherwise undetectable.

On a product that serves neither an REU nor a CPU speed setting, both suites report SKIP with the product name and the settings it lacks, and exit 0. `run-tests` derives a suite's verdict from its exit code alone, so the suite's own closing line is the only place that difference appears. Exercised without a device, by calling each suite against a stub that answers `/v1/info` as an Ultimate II+L and refuses every config category.

`doom-release` was also run from this branch rather than from the branch it came from, which is what exercises the `RunnersApi.upload` timeout: the 8 MB REU image took 14.0s to upload and the suite passed all eight checks.

The device is shared with another session on this bench, and one wedge and two anomalous runs during this work were traced to its `nios2-download` landing mid-run rather than to the firmware. Runs quoted above are from windows where the device was held exclusively.

The three device-free gate suites (`transport-usage`, `runner-policy`, `esp-depends`) and `input-batching` and `observability` pass.

## What is not here

GideonZ/1541ultimate#804 also carried a bitstream bisection harness under `tests/bisect/`. That was tooling for finding the commit, not a regression test, and it is not in this change.

Simulation was considered instead of a hardware suite. It is the right place for this particular defect and it already runs there under nvc. It cannot stand in for the hardware suite in general: a simulator cannot reach a fault that takes seconds of wall clock to arrive, and it does not model machine behaviour such as the post-reset slowdown, which is why the E2E build needs a wait the simulation build must not have. Sharing one program between the two is what keeps them from drifting apart.


chrisgleissner:
@GideonZ This is almost ready - I am just trying to remove the duplication around the assembly file which is based on yours with some adjustments.

PR #812, opened 2026-08-28
Stop the C64 once per monitor redraw, not once per row

chrisgleissner:
## Problem

The machine-code monitor is unusable over Telnet on an Ultimate II+L. Every
keystroke takes about 1.75s to produce anything at all: the device sends no
bytes for that long, then one screen's worth of redraw. On an Ultimate 64 the
same keystrokes answer in 29ms.

Measured with a raw socket, sending a key and timing the first byte back, from
a freshly power-cycled device, 30 samples over cursor, page and view keys:

| device | first byte | worst |
| --- | ---: | ---: |
| Ultimate 64 | 29ms | 50ms |
| Ultimate II+L | 1.755s | 1.904s |

It is not the transport and not the device generally: the same II+L answers a
cursor key in the **file browser** over the same Telnet session in 49ms. Only
the monitor is slow.

## Cause

`MachineMonitor` draws a memory view one row at a time. `draw_hex`,
`draw_ascii`, `draw_screen_codes`, `draw_binary` and `draw_disassembly` each
loop over the visible rows calling `read_row`, and `read_row` calls
`backend->read_block` once per row.

`U2MemoryBackend::read_block` brackets each call with
`C64::begin_stopped_session` / `end_stopped_session`. `C64::stop(false)` asks
for the machine to stop on a safe R/Wn sequence and polls 100 times at 1ms
before giving up and forcing the stop. On the II+L that poll always expires, so
each call costs about 100ms.

An 18-row view therefore spends about 1.8s per redraw and stops the C64
eighteen times to paint one screen.

The confirming measurement: the ASCII view shows four times as many bytes as
the hex view and takes the same time (1.906s against 1.934s). The cost is per
`read_block` call, not per byte.

Why only this machine: `U64MemoryBackend::begin_session` says it deliberately
does not freeze because "per-access DMA stops handle the brief pauses required
to read/write memory". That holds at 29ms. It does not at 100ms.

## Change

`MemoryBackend` gains `begin_redraw` and `end_redraw`, both defaulting to
nothing. `MachineMonitor::draw` brackets the whole redraw with them.
`U2MemoryBackend` holds one stopped session across that bracket, so the
per-row stops inside it find the machine already stopped and become no-ops.

Nothing else changes. Backends that do not override the two hooks are
unaffected, which is every backend except the II+L's. The U2 classic does not
compile the monitor at all.

## Result

Same measurement, matched firmware pair built from one tree differing only by
this change, 30 samples each, freshly power-cycled:

| | first byte | worst |
| --- | ---: | ---: |
| without | 1.768s | 1.904s |
| with | 0.162s | 0.267s |

About eleven times faster. It is also less disruptive to the machine being
watched: one stop per redraw instead of eighteen, so the C64 is held still for
about 0.15s where it used to be held for about 1.8s.

## Verification

**Host test, no device.** `machine_monitor_test` gains a check that holds the
shape the timing depends on: one balanced bracket per redraw, every block read
that redraw makes inside it, and several row reads per bracket rather than one
bracket each. Reverting `MachineMonitor::draw`'s bracket fails it with
"Drawing the monitor opened no redraw bracket."; with the bracket the suite
passes.

**On hardware.**

| device | lane | firmware | result |
| --- | --- | --- | --- |
| Ultimate II+L | Telnet | without | 5 checks pass, then fails |
| Ultimate II+L | Telnet | with | 29 checks pass, then fails at `[30]`, see below |
| Ultimate II+L | overlay | with | OK, 59 checks, 688s, no retries |
| Ultimate 64 | overlay | with | OK, no retries |

The overlay lane is unchanged at 688s against 709s before, which is what the
design predicts: there the machine is already frozen, so both hooks are no-ops.

The four RISC-V targets build. The two Nios targets were not built here; the
change is behind virtuals neither of them overrides.

## What this does not fix

The Telnet lane is still not green. With this change it reaches
`[30] FILL covers its whole range`, which fails for an unrelated reason.

That failure is pre-existing and not caused by this change. Running the same
check three times on each firmware of the matched pair:

- without this change: 14 of 256, 0 of 256, 0 of 256 bytes written
- with this change: 10 of 256, 0 of 256, 0 of 256 bytes written

Timing it shows a Fill climbing to 239 of 256 bytes over tens of seconds and
then never completing, so it is losing writes rather than being slow. That
matches what `u2_memory_backend.cc` already records for the frozen write path:
"5 of 45 single-byte Hex edits at addresses of $1000 and above did not reach
memory". It needs its own fix and its own evidence.

## Considered and not done

**Raising the harness's first-byte budget** (`TELNET_FIRST_BYTE_TIMEOUT_SECONDS`,
1.0s) above the device's 1.9s. That makes the suite tolerate the defect rather
than remove it, and it would slow every Telnet capture that legitimately draws
nothing on every device. With this change the existing 1.0s budget covers the
II+L's 0.267s worst case with room to spare.

**Shortening `C64::stop`'s 100ms R/Wn poll.** That poll exists so the machine
stops at a safe point, and it is used by every caller of `stop`, not just the
monitor. Making it give up sooner trades a bus-safety property for speed
across the whole firmware. Batching the reads gets the same result without
touching it.

**Batching the range commands as well.** `fill_range` writes one byte per
`backend->write`, and each of those pays the same 100ms on this machine.
`transfer_read_range` and `transfer_write_range` are already batched through
`TRANSFER_BLOCK`. Fill wants the same treatment, but it belongs with the
write-loss defect above rather than here.

### DIFF
--- a/software/monitor/machine_monitor.cc
+++ b/software/monitor/machine_monitor.cc
@@ -1636,20 +1636,42 @@ class MonitorBlockReader
 {
     MemoryBackend *backend;
     uint8_t buffer[MONITOR_READ_BLOCK];
+    uint32_t first;
+    uint32_t last;
     uint32_t base;
+    uint16_t held;
     bool filled;
 
 public:
-    explicit MonitorBlockReader(MemoryBackend *backend)
-        : backend(backend), base(0), filled(false) { }
+    // `first` and `last` bound what the caller will actually ask for, so a
+    // short range costs one short read rather than a whole block. Without that
+    // a sixteen-byte Compare would fetch 256 bytes per side, which is free on a
+    // backend whose reads are cheap but is still work nobody asked for.
+    MonitorBlockReader(MemoryBackend *backend, uint16_t first, uint16_t last)
+        : backend(backend), first(first), last(last), base(0), held(0),
+          filled(false) { }
 
     uint8_t read(uint16_t address)
     {
         uint32_t block = (uint32_t)address & ~(uint32_t)(MONITOR_READ_BLOCK - 1);
 
-        if (!filled || block != base) {
-            backend->read_block((uint16_t)block, buffer, MONITOR_READ_BLOCK);
-            base = block;
+        if (!filled || address < base || address >= base + held) {
+            uint32_t start = (block < first) ? first : block;
+            uint32_t end = block + MONITOR_READ_BLOCK - 1;
+
+            // Compare's second range can run past $FFFF and wrap, which leaves
+            // last below first. The bound is only meaningful when it does not,
+            // and a block never crosses the wrap itself, so the block's own end
+            // is the safe answer there.
+            if (last >= first && end > last) {
+                end = last;
+            }
+            if (end < start) {
+                end = start;
+            }
+            base = start;
+            held = (uint16_t)(end - start + 1);
+            backend->read_block((uint16_t)base, buffer, held);
             filled = true;
         }
         return buffer[address - base];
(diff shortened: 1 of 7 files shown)


chrisgleissner:
Combined into #813 together with #810, at our maintainer's request. All five commits are carried across unchanged and distinct, and the combined description corrects what this one said about issue 30: the FILL failure is the same per-access stop/resume defect as the redraw, not a separate write-loss defect, and it is fixed here rather than left open.

PR #814, opened 2026-08-28
Close the UCI network target's sockets on a C64 reset (#808)

### LINKED ISSUE #808: UCI network target leaks UDP sockets that are never CLOSEd — pool exhausts, only a power cycle recovers
## Summary
A UCI client that opens connected UDP sockets (`OPEN_UDP` / socket + connect) and does not issue `CLOSE_SOCKET` before going away permanently leaks the socket's `udp_pcb` / `netconn` / socket-table slot (plus one `netbuf`+pbuf per datagram left queued on its recv mailbox). Nothing in the network target ever reclaims them: there is no idle reaper, no max-concurrent cap, and a **C64 reset does not close them**. After ~5 abandoned sockets the smallest pool (8) is exhausted and further opens fail; recovery requires a **wall power cycle** (`machine:reboot` does not help). This is a robustness gap — a misbehaving or mid-session-reset client can render the network interface unusable until the box is physically power-cycled.

## Reproduction (measured, from a physical power cycle)
Freshly power-cycled U64E, fw 3.15. Each "session" = the C64 opens ONE connected UDP socket (`OPEN_UDP` + `connect`), completes a WireGuard handshake + data exchange, and is dropped **without `CLOSE_SOCKET`** (the client program is reloaded per session, so the next boot doesn't know the prior socket id to close it):

```
session #1..#5 : socket opens, handshake completes (9/9)     [5 connected UDP sockets leaked]
session #6     : OPEN_UDP no longer yields a usable socket    -> client reports NO_SOCKET
               : net_init / DHCP / bind still succeed — only socket creation fails
```

Deterministic at the **6th** open from a clean power-up: `MEMP_NUM_UDP_PCB = 8` minus the firmware's own baseline UDP users (~3) leaves ~5, so the 6th `OPEN_UDP` exhausts the pool. Only a wall power cycle restores it. (Earlier sittings showed 3–8; the variation is just how many sockets the firmware had already used before we started.)

## Environment
- Ultimate 64 Elite, firmware **3.15** (`update_v3.15-74-g6b5ffc21.u64`), FPGA 123, core 1.4E.
- Client: a 6502 UCI driver (c64-wireguard). **The proximate cause is on our side — we don't `CLOSE_SOCKET` on session teardown, and we reload the client program per session so the fresh boot doesn't know the prior socket's id to close it. We are fixing that.** Filing here because the *firmware* offers no way to recover short of power-cycling, which is the part worth hardening.

**Allocation.** `OPEN_UDP` (0x08) → `open_socket(..., SOCK_DGRAM)`: `network_target.cc:203` `socket(AF_INET,SOCK_DGRAM,0)` takes a table slot (`sockets[NUM_SOCKETS]`, `alloc_socket()` `sockets.c:304`) + a `netconn` (`MEMP_NETCONN`) + a `udp_pcb` (`MEMP_UDP_PCB`); `:209` `connect()` only binds the remote.

**Pool limits** (`software/network/config/lwipopts.h`):
- `MEMP_NUM_UDP_PCB = 8` (`:197`) ← smallest, and the binding limit here
- `MEMP_NUM_NETBUF = 8` (`:249`) — one per queued datagram
- `MEMP_NUM_NETCONN = 16` (`:255`, = `NUM_SOCKETS`)

Firmware baseline already uses some UDP (`syslog.cc`, `data_streamer.cc`, `debug_streamer_u2p.cc`), so the free `udp_pcb` headroom is a few — consistent with our measured exhaustion on the 6th open (5 leaked + ~3 baseline = 8).

**No reclamation.** `NetworkTarget` keeps no record of open sockets (members are only the message/status buffers, `network_target.h:27-42`) — it cannot enumerate or reap them. The only `lwip_close` sites are explicit `CLOSE_SOCKET` (`:297`), `read_socket` when `recv()==0` (`:255`), and connect/guard failures (`:214,:220`). A connected UDP `recv()` returns `>0` or `<0` (EWOULDBLOCK), effectively **never 0**, so the auto-close never fires for UDP. An abandoned connected UDP socket is never freed.

**Reset doesn't help.** `NetworkTarget net(3)` is a static global built once at ARM boot (`:15`); the socket table + memp live in ARM RAM. `run_reset_task()` (C64 reset) only calls `Sampler::reset()` and does not touch the net stack; `NetworkTarget::abort()` is **empty** (`:314-317`). `machine:reboot` restarts the 6502, not the ARM app that owns lwip. Only power-up re-inits the pools — matching all field observations.

## Honest nuance (please confirm)
The source has **no infinite-block path** on exhaustion: `alloc_socket` returns −1 when full (`sockets.c:576`) and `open_socket` emits **`85,ERROR OPENING SOCKET`** on the STATUS channel (`:205`). So the firmware returns an *error*, not silence. On our side the apparent "hang" was our own client spin-waiting on the DATA register while the `85` sat unread on the STATUS channel (a known client bug of ours) — **so please read our "hangs" as "the open stops yielding a usable socket and reports 85 on STATUS", not "the ARM wedges".** A genuine ARM-side wedge (e.g. a FreeRTOS heap/queue-alloc failure under `netconn_new`) would be below the visible source and we have no evidence for it.

## `CLOSE_SOCKET` fully recovers a socket (so the client fix works)
`lwip_close` → `netconn_delete` → for UDP `udp_remove` frees the pcb (`api_msg.c:1171`); `netconn_drain` `netbuf_delete`s every queued datagram (`api_msg.c:227`), then the netconn and slot are freed. All four resources reclaimed. Issuing `CLOSE_SOCKET` on teardown prevents the leak entirely — but only helps clients that get the chance to (not ones reset mid-session).

## Suggested firmware hardening (any one bounds it)
1. **Max-concurrent cap on UCI-opened sockets** — track them in a fixed array in `NetworkTarget`; on a new open beyond the cap, `lwip_close` the oldest. Bounds the leak regardless of client behaviour.
2. **Close-on-C64-reset** — have the reset path (or the currently-empty `NetworkTarget::abort()`, `:314`) close all UCI-opened sockets, so a C64 reset recovers the interface without a power cycle.
3. **Idle/linger reaper** — timestamp each UCI socket, close after N seconds idle (needs the table from #1).

## Files for the maintainer
`software/io/network/network_target.cc` — `open_socket` `:174`, `read_socket` `:235` (UDP never hits the `ret==0` close at `:254`), `close_socket` `:294`, empty `abort` `:314`; `network_target.h:27` (no socket table). `software/network/config/lwipopts.h:197/:249/:255`. `software/io/command_interface/command_intf.cc` — reset path ignores the net target. lwip `26a2215`: `alloc_socket` `sockets.c:540` (returns −1, does not block), `recv_udp` `api_msg.c:218`, `netconn_drain`/`do_delconn` `api_msg.c:227,:1171`.

JC-000:
Fixes #808. A UCI client that opens sockets and never closes them holds them until the device is power cycled, because nothing but the client's own `CLOSE_SOCKET` ever returns one. `NetworkTarget` now records the sockets it opens, reads, writes and closes only those, and closes all of them when the C64 resets. A C64 reset therefore recovers the network target, which is the recovery path #808 asks for.

## What is broken

Measured through `$DF1B-$DF1F` on an Ultimate 64 Elite running unpatched `test-merge`:

- A C64 reset releases nothing. With sockets left open, they are still open afterwards and can still be closed by handle, so the reset did not touch them. The program that opened them is gone, so nothing can ever close them.
- Only the client's own `CLOSE_SOCKET` returns a socket. `NetworkTarget::open_socket()` in `software/io/network/network_target.cc` calls `socket()` and `connect()` and hands the raw lwip descriptor to the client. Nothing records it.
- Repeated `OPEN_UDP` without `CLOSE_SOCKET` therefore exhausts the pool and further opens answer `85,ERROR OPENING SOCKET`. `MEMP_NUM_UDP_PCB` is 8 in `software/network/config/lwipopts.h` and the firmware's own UDP users hold the rest, so how many opens a client gets depends on what the firmware has open at the time. It was 5 on one device and 4 on another.
- lwip reuses a closed descriptor's number for the next socket anyone opens, including sockets the firmware opens for itself. A client still using a stale handle reaches whichever socket took that number.

## What this change does

All inside `NetworkTarget`, plus a reset hook in the command interface.

### A record of the sockets this target opened

`NetworkTarget` keeps `int sockets[NET_MAX_SOCKETS]` and `socket_count`. `OPEN_TCP` and `OPEN_UDP` add to it. `CLOSE_SOCKET` and the existing auto-close on `recv() == 0` remove from it.

`NET_MAX_SOCKETS` is 16, which is lwip's `MEMP_NUM_NETCONN` in `software/network/config/lwipopts.h` and therefore the number of sockets lwip can create at all. The table can hold every socket this target could ever be given, so it cannot overflow and no client meets a limit of its own. An open is refused only when lwip refuses it, with the same `85,ERROR OPENING SOCKET` as before this change.

`track_socket()` is bounds checked and returns false rather than writing past the table. A `static_assert(NET_MAX_SOCKETS >= MEMP_NUM_NETCONN)` fails the build if `lwipopts.h` ever outgrows the table. Without that check an untracked socket would be handed to a client, survive a reset, and reintroduce this issue with no test able to detect it.

### Only owned sockets are touched

`READ_SOCKET`, `WRITE_SOCKET` and `CLOSE_SOCKET` refuse a handle that is not in the table by setting `errno = EBADF`, which produces the status texts an unknown handle already produced: `02,NO DATA: 9`, `12,SEND ERROR: 9` and `12,ERROR ON CLOSE: 9`.

This is required by the reset hook below. Closing sockets on a reset returns their descriptor numbers to lwip, which reuses them, so without this gate a program running after a reset could read from or write to a socket the firmware opened for itself.

### Close everything on a C64 reset

The reset interrupt (`ResetInterruptHandlerCmdIf` in `software/io/command_interface/command_intf.cc`) already posts a `CMD_ABORT_DATA` event to the command interface's server task. It now also sets a `c64_reset_pending` flag. The server task reads and clears that flag as one step on every event it dequeues, and calls a new `CommandTarget::c64_reset()` on every target. `NetworkTarget` implements it as close-all plus discarding any pending read reply; the base class is a no-op.

Running this on the server task means a target is never told about the reset while it is in the middle of a command. Checking the flag on every event rather than only on the abort means a reset whose queue post did not fit is still acted on by the next event that does.

## Backwards compatibility

Intentional differences:

- After a C64 reset every UCI socket is closed. A program that survives a reset and expects its sockets to survive with it would break. The reset also aborts any UCI transaction in flight today.
- `READ_SOCKET` and `WRITE_SOCKET` on a handle not obtained from `OPEN_*`, or already closed, now fail with `... : 9` (EBADF) instead of acting on whatever socket lwip gave that number to. A handle that was never valid already failed that way.

Unchanged: the socket id format, the status returned by every command including `OPEN_*` at pool exhaustion, `lwipopts.h`, and the FPGA. There is no new status code and no change to the UCI protocol, so no documentation change is required.

Cost is 48 bytes of bss.

## Alternatives considered

- **A cap on how many sockets one client may hold.** Tried, and removed. On a device whose spare UDP control block count equals the cap, both builds refuse at the same open with the same status, so the cap changes nothing a client can observe and no test can distinguish it from pool exhaustion. The effective limit is `min(cap, free pool)` and the second term moves at runtime: one run refused at 3 rather than 4, failing a scenario on correct firmware. It does not reserve headroom for the firmware's own UDP users either, since a cap of 4 against a pool of 8 with a firmware baseline near 4 still lets a client take every remaining control block.
- **Giving the cap a status of its own** so that it becomes observable and testable. This is the only way a cap could be regression tested, but it adds UCI protocol surface that cannot be withdrawn once clients depend on it, and it needs a matching change in `GideonZ/1541u-documentation`. Days before 3.15 the smaller change is preferred. The table added here is what any later cap or idle reaper would be built on, so nothing is foreclosed.
- **Hooking the reset in `run_reset_task()`**, the task that wakes on the reset semaphore and resets the sampler. It runs on a different FreeRTOS task from `parse_command()`, so closing sockets there races a command in progress. Rejected for the flag on the server task.
- **Tagging the reset as a bit in the queued status byte.** Bit 7 of the hardware status byte is `DATA_AV` and `command_interface_irq()` passes bits 3 to 7 straight into the queue, so any spare bit collides. Rejected.
- **An idle reaper.** Needs a timer and a policy on what counts as idle. Close-on-reset already turns the failure from "power cycle" into "self healing". Not done.

## Not addressed

A client that leaks sockets and never resets is not bounded. It still exhausts the pool and still receives `85,ERROR OPENING SOCKET`, as it does today. What changes is that any C64 reset now recovers the interface, rather than only a mains power cycle.

## How it was tested

One scenario, `reset-closes-uci-sockets`, added to `tests/e2e/io/command_interface/net_target_test.py` (`run-tests` id `uci-net-target`), running on both the REST route (DMA on the registers) and the native route (6502 agent on the C64). It opens four sockets, resets the C64, and requires that `CLOSE_SOCKET` on each answers an error because the reset already closed them. It then opens four more, which must succeed, since sockets still held would leave lwip that many fewer to give. Three cycles. It always closes what it opened, so the device is left clean even on failing firmware.

Red and green on an Ultimate 64 Elite (core 1.4F), two builds differing only by this change, both carrying the same test file. Baseline is `test-merge` `fcc2502d`.

| Run | baseline REST | baseline native | with change REST | with change native |
|---|---|---|---|---|
| 1 | FAIL | FAIL | OK | OK |
| 2 | FAIL | FAIL | OK | OK |
| 3 | FAIL | FAIL | OK | OK |
| 4 | FAIL | FAIL | OK | OK |
| 5 | FAIL | FAIL | OK | OK |
| 6 | FAIL | FAIL | OK | OK |
| 7 | FAIL | FAIL | OK | OK |
| 8 | FAIL | FAIL | OK | OK |
| 9 | FAIL | FAIL | OK | OK |
| 10 | FAIL | FAIL | OK | OK |
| **total** | **10/10 FAIL** | **10/10 FAIL** | **10/10 OK** | **10/10 OK** |

The baseline failure is the same on every run:

```
reset-closes-uci-sockets [rest] stopped: handles [8, 9, 10, 11] were still open
after the reset; the program that opened them is gone, so nothing else can ever
close them
```

Whole `uci-net-target` suite, both routes, on the same two builds:

| Build | Result |
|---|---|
| baseline `fcc2502d`, run 1 | 19 scenarios OK, `reset-closes-uci-sockets` FAIL on both routes |
| baseline `fcc2502d`, run 2 | 19 scenarios OK, `reset-closes-uci-sockets` FAIL on both routes |
| with this change, run 1 | 21 of 21 scenarios OK, 133 checks, 403 s |
| with this change, run 2 | 21 of 21 scenarios OK, 133 checks, 403 s |

### A defect in the test harness, fixed in `16d42416`

`Peer.learn_udp_peer()` took whatever datagram arrived next as its probe reply. A datagram left queued at the host by an earlier scenario was read as the current socket's, so the learned source port belonged to a socket that had already been closed, and every later host to device datagram went to a port nothing was listening on. The device then answered a truthful `02,NO DATA: 11` while its own probe still reached the host.

This produced seven failures on whichever route ran second, on both patched and unpatched builds, and none when a route ran alone. It was reported as a device-side flake before being traced. `16d42416` drains the host socket before probing. The nine scenarios unrelated to this change now pass on both routes on both builds.

### DIFF
--- a/software/io/command_interface/command_intf.cc
+++ b/software/io/command_interface/command_intf.cc
@@ -125,9 +125,14 @@ void CommandInterface :: run_task(void)
 #endif
 		// Checked on every event rather than only on the abort the reset
 		// interrupt posts, so a reset whose post did not fit the queue is
-		// still acted on by the next event that does.
-		if (c64_reset_pending) {
-			c64_reset_pending = false;
+		// still acted on by the next event that does. Read and cleared as one
+		// step, because a reset landing between the two would otherwise be
+		// cleared without ever being told to the targets.
+		portENTER_CRITICAL();
+		bool reset_seen = c64_reset_pending;
+		c64_reset_pending = false;
+		portEXIT_CRITICAL();
+		if (reset_seen) {
 			for(int i=0;i<=CMD_IF_MAX_TARGET;i++)
 				command_targets[i]->c64_reset();
 		}
--- a/software/io/command_interface/command_intf.cc
+++ b/software/io/command_interface/command_intf.cc
@@ -18,6 +18,10 @@ CommandTarget *command_targets[CMD_IF_MAX_TARGET+1];
 
 // Semaphore set by interrupt
 static SemaphoreHandle_t resetSemaphore;
+// Set by the reset interrupt for the server task, which tells the targets on
+// its own thread rather than the reset task's, so a target never hears of the
+// reset while it is in the middle of a command.
+static volatile bool c64_reset_pending;
 
 #define CMD_IF_DEBUG 0
 
@@ -27,6 +31,7 @@ void ResetInterruptHandlerCmdIf()
 {
     BaseType_t woken;
     uint8_t new_flags = CMD_ABORT_DATA;
+    c64_reset_pending = true;
     xSemaphoreGiveFromISR(resetSemaphore, &woken);
     xQueueSendFromISR(cmd_if.queue, &new_flags, &woken);
 }
@@ -126,6 +131,11 @@ void CommandInterface :: run_task(void)
 			}
 			CMD_IF_HANDSHAKE_OUT = HANDSHAKE_RESET;
 			CMD_IF_IRQMASK_CLEAR = CMD_ABORT_DATA;
+			if (c64_reset_pending) {
+				c64_reset_pending = false;
+				for(int i=0;i<=CMD_IF_MAX_TARGET;i++)
+					command_targets[i]->c64_reset();
+			}
 		}
 		if(status_byte & CMD_DATA_ACCEPTED) {
 			if (target != CMD_TARGET_NONE) {
--- a/software/io/command_interface/command_intf.cc
+++ b/software/io/command_interface/command_intf.cc
@@ -18,9 +18,8 @@ CommandTarget *command_targets[CMD_IF_MAX_TARGET+1];
 
 // Semaphore set by interrupt
 static SemaphoreHandle_t resetSemaphore;
-// Set by the reset interrupt for the server task, which tells the targets on
-// its own thread rather than the reset task's, so a target never hears of the
-// reset while it is in the middle of a command.
+// Set by the reset ISR for the server task, so a target is told about the
+// reset between commands rather than during one.
 static volatile bool c64_reset_pending;
 
 #define CMD_IF_DEBUG 0
@@ -123,11 +122,9 @@ void CommandInterface :: run_task(void)
 #if CMD_IF_DEBUG
 		printf("{%b}", status_byte);
 #endif
-		// Checked on every event rather than only on the abort the reset
-		// interrupt posts, so a reset whose post did not fit the queue is
-		// still acted on by the next event that does. Read and cleared as one
-		// step, because a reset landing between the two would otherwise be
-		// cleared without ever being told to the targets.
+		// Checked on every event, so a reset whose queue post was dropped is
+		// still acted on. Read and cleared as one step: a reset landing
+		// between the two would be cleared without ever reaching the targets.
 		portENTER_CRITICAL();
 		bool reset_seen = c64_reset_pending;
 		c64_reset_pending = false;
--- a/software/io/command_interface/command_intf.cc
+++ b/software/io/command_interface/command_intf.cc
@@ -123,6 +123,14 @@ void CommandInterface :: run_task(void)
 #if CMD_IF_DEBUG
 		printf("{%b}", status_byte);
 #endif
+		// Checked on every event rather than only on the abort the reset
+		// interrupt posts, so a reset whose post did not fit the queue is
+		// still acted on by the next event that does.
+		if (c64_reset_pending) {
+			c64_reset_pending = false;
+			for(int i=0;i<=CMD_IF_MAX_TARGET;i++)
+				command_targets[i]->c64_reset();
+		}
 		if(status_byte & CMD_ABORT_DATA) {
 			//printf("Abort received.\n");
 			if (target != CMD_TARGET_NONE) {
@@ -131,11 +139,6 @@ void CommandInterface :: run_task(void)
 			}
 			CMD_IF_HANDSHAKE_OUT = HANDSHAKE_RESET;
 			CMD_IF_IRQMASK_CLEAR = CMD_ABORT_DATA;
-			if (c64_reset_pending) {
-				c64_reset_pending = false;
-				for(int i=0;i<=CMD_IF_MAX_TARGET;i++)
-					command_targets[i]->c64_reset();
-			}
 		}
 		if(status_byte & CMD_DATA_ACCEPTED) {
 			if (target != CMD_TARGET_NONE) {
--- a/software/io/command_interface/command_target.h
+++ b/software/io/command_interface/command_target.h
@@ -44,6 +44,10 @@ public:
     }
     
     virtual void abort(int a) { }
+
+    // The C64 was reset, so whatever program was talking to this target is
+    // gone and can never finish what it started.
+    virtual void c64_reset(void) { }
 };
 
 extern CommandTarget *command_targets[];
--- a/software/io/command_interface/command_target.h
+++ b/software/io/command_interface/command_target.h
@@ -45,8 +45,7 @@ public:
     
     virtual void abort(int a) { }
 
-    // The C64 was reset, so whatever program was talking to this target is
-    // gone and can never finish what it started.
+    // The C64 was reset, so the program talking to this target is gone.
     virtual void c64_reset(void) { }
 };
 
--- a/software/io/network/network_target.cc
+++ b/software/io/network/network_target.cc
@@ -209,10 +209,8 @@ void NetworkTarget :: open_socket(Message *command, Message **reply, Message **s
 		return;
 	}
 
-	// The client's own limit, refused rather than enforced by taking a socket
-	// back: a socket the client still believes it holds must not go away
-	// underneath it. 85 is already what a client sees when there is no socket
-	// to give, so this adds no status a client has to learn.
+	// The client's own limit. Refused rather than reclaimed, with the answer
+	// an exhausted pool already gives. See NET_MAX_SOCKETS.
 	if (socket_count >= NET_MAX_SOCKETS) {
 		*status = &c_status_no_socket;
 		return;
@@ -447,11 +445,9 @@ void NetworkTarget :: close_all_sockets(void)
 
 void NetworkTarget :: c64_reset(void)
 {
-    // The reply is dropped here as well as in abort(), because this hook runs
-    // for a reset whose CMD_ABORT_DATA post did not fit the queue, and that is
-    // exactly the case where abort() is never called. Without it the next
-    // program asking for a further block over Data More would be handed what
-    // is left of the previous program's read.
+    // abort() also drops the reply, but it does not run when the reset's
+    // queue post was dropped. Without this the next program could be handed
+    // what is left of the previous one's read over Data More.
     discard_read_reply();
     close_all_sockets();
 }
--- a/software/io/network/network_target.cc
+++ b/software/io/network/network_target.cc
@@ -29,11 +29,13 @@ NetworkTarget::NetworkTarget(int id)
     command_targets[id] = this;
     data_message.message = new uint8_t[CMD_MAX_REPLY_LEN];
     status_message.message = new uint8_t[CMD_MAX_STATUS_LEN];
+    socket_count = 0;
     discard_read_reply();
 }
 
 NetworkTarget::~NetworkTarget()
 {
+    close_all_sockets();
     delete[] data_message.message;
     delete[] status_message.message;
 }
@@ -207,6 +209,14 @@ void NetworkTarget :: open_socket(Message *command, Message **reply, Message **s
 		return;
 	}
 
+	// Freed before the new socket is taken, so the client never sees the
+	// pool run out before its own limit does.
+	if (socket_count == NET_MAX_SOCKETS) {
+		int oldest = sockets[0];
+		untrack_socket(oldest);
+		lwip_close(oldest);
+	}
+
 	int socket = socket(AF_INET, type, 0);
 	if (socket < 0) {
 		*status = &c_status_no_socket;
@@ -228,6 +238,7 @@ void NetworkTarget :: open_socket(Message *command, Message **reply, Message **s
         return;
 	}
 
+	track_socket(socket);
 	*reply = &data_message;
 	this->data_message.message[0] = (uint8_t)socket;
 	this->data_message.length = 1;
@@ -283,6 +294,7 @@ void NetworkTarget :: read_socket(Message *command, Message **reply, Message **s
 
     // printf("Reading %d bytes from socket %d resulted in %d\n", length, socketnr, ret);
 	if (ret == 0) {
+		untrack_socket(socketnr);
 		lwip_close(socketnr);
 		*status = &c_status_socket_closed;
 		return;
@@ -366,7 +378,13 @@ void NetworkTarget :: write_socket(Message *command, Message **reply, Message **
 void NetworkTarget :: close_socket(Message *command, Message **reply, Message **status)
 {
     uint8_t socketnr = command->message[2];
-    int result = lwip_close(socketnr);
+    int result = -1;
+    if (owns_socket(socketnr)) {
+        untrack_socket(socketnr);
+        result = lwip_close(socketnr);
+    } else {
+        errno = EBADF;
+    }
     *reply = &c_message_empty;
     if (result < 0) {
         *status = &status_message;
@@ -375,6 +393,47 @@ void NetworkTarget :: close_socket(Message *command, Message **reply, Message **
     } else {
         *status = &c_status_ok;
     }
+}
+
+void NetworkTarget :: track_socket(int socketnr)
+{
+    sockets[socket_count++] = socketnr;
+}
+
+void NetworkTarget :: untrack_socket(int socketnr)
+{
+    for (int i = 0; i < socket_count; i++) {
+        if (sockets[i] == socketnr) {
+            socket_count--;
+            for (; i < socket_count; i++) {
+                sockets[i] = sockets[i + 1];
+            }
+            return;
+        }
+    }
+}
+
+bool NetworkTarget :: owns_socket(int socketnr)
+{
+    for (int i = 0; i < socket_count; i++) {
+        if (sockets[i] == socketnr) {
+            return true;
+        }
+    }
+    return false;
+}
+
+void NetworkTarget :: close_all_sockets(void)
+{
+    for (int i = 0; i < socket_count; i++) {
+        lwip_close(sockets[i]);
+    }
+    socket_count = 0;
+}
+
+void NetworkTarget :: c64_reset(void)
+{
+    close_all_sockets();
 }
 
 void NetworkTarget :: get_more_data(Message **reply, Message **status)
--- a/software/io/network/network_target.cc
+++ b/software/io/network/network_target.cc
@@ -8,6 +8,7 @@
 #include "network_target.h"
 #include "network_interface.h"
 #include "socket.h"
+#include "lwip/opt.h"
 #include "netdb.h"
 #include <errno.h>
 #include <stdio.h>
@@ -209,13 +210,6 @@ void NetworkTarget :: open_socket(Message *command, Message **reply, Message **s
 		return;
 	}
 
-	// The client's own limit. Refused rather than reclaimed, with the answer
-	// an exhausted pool already gives. See NET_MAX_SOCKETS.
-	if (socket_count >= NET_MAX_SOCKETS) {
-		*status = &c_status_no_socket;
-		return;
-	}
-
 	int socket = socket(AF_INET, type, 0);
 	if (socket < 0) {
 		*status = &c_status_no_socket;
@@ -237,8 +231,14 @@ void NetworkTarget :: open_socket(Message *command, Message **reply, Message **s
         return;
 	}
 
-	// The cap above left room, so this is within the table.
-	track_socket(socket);
+	// The table has room for every socket lwip can create, so this cannot
+	// fail. Refusing rather than writing past the table keeps that a check
+	// instead of an assumption.
+	if (!track_socket(socket)) {
+		*status = &c_status_no_socket;
+		lwip_close(socket);
+		return;
+	}
 	*reply = &data_message;
 	this->data_message.message[0] = (uint8_t)socket;
 	this->data_message.length = 1;
@@ -407,9 +407,20 @@ void NetworkTarget :: close_socket(Message *command, Message **reply, Message **
     }
 }
 
-void NetworkTarget :: track_socket(int socketnr)
+// A socket handed to the client without a table entry would survive a C64
+// reset, which is the defect in #808, so the table must cover every descriptor
+// lwip can return. lwip's NUM_SOCKETS is MEMP_NUM_NETCONN, tested here rather
+// than including lwip's private socket header.
+static_assert(NET_MAX_SOCKETS >= MEMP_NUM_NETCONN,
+              "the socket table is smaller than the number of lwip sockets");
+
+bool NetworkTarget :: track_socket(int socketnr)
 {
+    if (socket_count >= NET_MAX_SOCKETS) {
+        return false;
+    }
     sockets[socket_count++] = socketnr;
+    return true;
 }
 
 void NetworkTarget :: untrack_socket(int socketnr)
--- a/software/io/network/network_target.cc
+++ b/software/io/network/network_target.cc
@@ -209,23 +209,22 @@ void NetworkTarget :: open_socket(Message *command, Message **reply, Message **s
 		return;
 	}
 
-	// Freed before the new socket is taken, so the client never sees the
-	// pool run out before its own limit does. An open that then fails has
-	// still cost the client its oldest socket: the slot only exists once
-	// something has been closed.
-	//
-	// A loop rather than a single test so that track_socket() below is
-	// provably within the table. sockets[NET_MAX_SOCKETS] would be
-	// socket_count itself, so a count that ever exceeded the table would
-	// overwrite the count rather than fail where it could be seen.
-	// untrack_socket() removes exactly the entry named, so this terminates.
-	while (socket_count >= NET_MAX_SOCKETS) {
-		int oldest = sockets[0];
-		untrack_socket(oldest);
-		lwip_close(oldest);
-	}
-
+	// A socket is given up only for one that is going to be handed out, so an
+	// open that fails costs the client nothing. This matters for TCP, where a
+	// refused or unanswered connect() is the ordinary error path rather than
+	// an edge case: a client retrying a connection to a host that is down
+	// would otherwise lose every socket it holds, one per attempt.
 	int socket = socket(AF_INET, type, 0);
+	if (socket < 0 && socket_count >= NET_MAX_SOCKETS) {
+		// The pool is out while this client holds its full share of it, so
+		// the client gives up its oldest and tries once more. Once only: a
+		// second refusal is the rest of the device using the pool rather than
+		// this client, and retrying would close everything the client has for
+		// nothing. Below the cap the answer stays 85, because a pool the
+		// firmware itself filled is not the client's to pay for.
+		close_oldest_socket();
+		socket = socket(AF_INET, type, 0);
+	}
 	if (socket < 0) {
 		*status = &c_status_no_socket;
 		return;
@@ -246,6 +245,18 @@ void NetworkTarget :: open_socket(Message *command, Message **reply, Message **s
         return;
 	}
 
+	// The cap, applied now that the socket exists and is about to be handed
+	// out. Between socket() above and here the client holds one more than the
+	// cap, which only happens when the pool had room to spare: had it not,
+	// socket() would have failed and taken the evict-and-retry branch.
+	//
+	// A loop rather than a single test so that track_socket() is provably
+	// within the table. sockets[NET_MAX_SOCKETS] would be socket_count
+	// itself, so a count that ever exceeded the table would overwrite the
+	// count rather than fail where it could be seen.
+	while (socket_count >= NET_MAX_SOCKETS) {
+		close_oldest_socket();
+	}
 	track_socket(socket);
 	*reply = &data_message;
 	this->data_message.message[0] = (uint8_t)socket;
@@ -441,6 +452,18 @@ bool NetworkTarget :: owns_socket(int socketnr)
         }
     }
     return false;
+}
+
+void NetworkTarget :: close_oldest_socket(void)
+{
+    // untrack_socket() removes exactly the entry named, so a caller that
+    // loops on socket_count terminates.
+    if (socket_count == 0) {
+        return;
+    }
+    int oldest = sockets[0];
+    untrack_socket(oldest);
+    lwip_close(oldest);
 }
 
 void NetworkTarget :: close_all_sockets(void)
--- a/software/io/network/network_target.cc
+++ b/software/io/network/network_target.cc
@@ -213,7 +213,13 @@ void NetworkTarget :: open_socket(Message *command, Message **reply, Message **s
 	// pool run out before its own limit does. An open that then fails has
 	// still cost the client its oldest socket: the slot only exists once
 	// something has been closed.
-	if (socket_count == NET_MAX_SOCKETS) {
+	//
+	// A loop rather than a single test so that track_socket() below is
+	// provably within the table. sockets[NET_MAX_SOCKETS] would be
+	// socket_count itself, so a count that ever exceeded the table would
+	// overwrite the count rather than fail where it could be seen.
+	// untrack_socket() removes exactly the entry named, so this terminates.
+	while (socket_count >= NET_MAX_SOCKETS) {
 		int oldest = sockets[0];
 		untrack_socket(oldest);
 		lwip_close(oldest);
@@ -447,6 +453,12 @@ void NetworkTarget :: close_all_sockets(void)
 
 void NetworkTarget :: c64_reset(void)
 {
+    // The reply is dropped here as well as in abort(), because this hook runs
+    // for a reset whose CMD_ABORT_DATA post did not fit the queue, and that is
+    // exactly the case where abort() is never called. Without it the next
+    // program asking for a further block over Data More would be handed what
+    // is left of the previous program's read.
+    discard_read_reply();
     close_all_sockets();
 }
 
--- a/software/io/network/network_target.cc
+++ b/software/io/network/network_target.cc
@@ -209,22 +209,16 @@ void NetworkTarget :: open_socket(Message *command, Message **reply, Message **s
 		return;
 	}
 
-	// A socket is given up only for one that is going to be handed out, so an
-	// open that fails costs the client nothing. This matters for TCP, where a
-	// refused or unanswered connect() is the ordinary error path rather than
-	// an edge case: a client retrying a connection to a host that is down
-	// would otherwise lose every socket it holds, one per attempt.
-	int socket = socket(AF_INET, type, 0);
-	if (socket < 0 && socket_count >= NET_MAX_SOCKETS) {
-		// The pool is out while this client holds its full share of it, so
-		// the client gives up its oldest and tries once more. Once only: a
-		// second refusal is the rest of the device using the pool rather than
-		// this client, and retrying would close everything the client has for
-		// nothing. Below the cap the answer stays 85, because a pool the
-		// firmware itself filled is not the client's to pay for.
-		close_oldest_socket();
-		socket = socket(AF_INET, type, 0);
+	// The client's own limit, refused rather than enforced by taking a socket
+	// back: a socket the client still believes it holds must not go away
+	// underneath it. 85 is already what a client sees when there is no socket
+	// to give, so this adds no status a client has to learn.
+	if (socket_count >= NET_MAX_SOCKETS) {
+		*status = &c_status_no_socket;
+		return;
 	}
+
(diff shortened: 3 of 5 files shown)


JC-000:
I had time and thought it would be interesting to set up the build chain, so I went ahead and worked it. I won't be offended at all if this is not the way you want to solve it.

JC-000:
Please hold off merging this for the moment.

Reviewing my own reset path I found that the read-and-clear of `c64_reset_pending` in `run_task()` is not atomic against the reset ISR: a reset landing between the test and the clear is cleared without ever reaching the targets, so sockets opened before it survive the reset this branch exists to close. It is a narrow window, but it is exactly the guarantee the PR claims, so I would rather fix it than ship it.

The fix takes the flag under `portENTER_CRITICAL()` and acts on what was read. It builds clean; I am flashing it to the rig now and will re-run both scenarios plus the full suite, then push the commit and post the updated evidence here.

Nothing else in the diff changes.

chrisgleissner:
Hi @JC-000 ,

thanks a lot for this PR and all the efforts you put into this. It is much appreciated. Please place it in "Draft" whilst it is not yet ready for review/merge. Once you are fully satisfied that it is as good as can be, please move it out of that mode.

Thanks

JC-000:
Fixed and re-tested. `18dee7b6` is up.

The read-and-clear of `c64_reset_pending` in `CommandInterface::run_task()` now happens under `portENTER_CRITICAL()`, and the loop acts on the value it read. Previously a reset landing between the test and the clear was cleared without ever reaching the targets, so sockets opened before it survived the reset — which is the one guarantee this branch adds.

Evidence, on a U64 Elite (fw 3.15, FPGA 123, core 1.4E) flashed with the fix build:

- `tests/e2e/io/command_interface/net_target_test.py`: **175 checks OK**, all 11 scenarios on both the REST and native routes, 726s.
- `reset-closes-uci-sockets` and `abandoned-sockets-are-bounded` green on both routes; free heap flat across reset cycles (5287268 -> 5287080 over 3 cycles).

One honest gap: I could not write a red test for this commit. Reproducing it needs the reset IRQ to land inside a two-instruction window, which the E2E level cannot steer. The existing `reset-closes-uci-sockets` scenario covers the guarantee it protects and stays green, but I am flagging that this commit is not red/green driven rather than implying it was.

The red evidence for the original defect is unchanged; this commit touches only the reset path, not the tests.

chrisgleissner:
Hi @JC-000 ,

as per your last comment, I understand you are now done with this work and it's ready for review. 

I am currently reviewing. 

Thanks
Christian

chrisgleissner:
Adversarial review of the diff. I have push access to this branch through maintainer edits, and a firmware cut is close, so I pushed the two fixes rather than send them back as a request. Revert anything you disagree with.

## Pushed

**1. `c64_reset()` left a pending read reply in place (9bf50704).** `abort()` is what discards a partly delivered multi-block reply. The reset hook was added for the case where the reset's `CMD_ABORT_DATA` post did not fit the queue, and that is exactly the case where `abort()` is never called. The FPGA handshake is not reset either on that path, so the next program can ask for a further block over Data More and be handed what is left of the previous program's read. `c64_reset()` now calls `discard_read_reply()` before `close_all_sockets()`.

**2. The cap test could not protect `track_socket()` (same commit).** `if (socket_count == NET_MAX_SOCKETS)` skips eviction for any count above the cap, and `track_socket()` has no bound of its own. Because `sockets[NET_MAX_SOCKETS]` and `socket_count` are adjacent members in that order, an out of range write lands on the count itself, which is the one variable that would hide the problem. Changed to `while (socket_count >= NET_MAX_SOCKETS)`. `untrack_socket()` removes exactly the entry named, so the loop terminates, and behaviour at the cap is unchanged.

**3. The cap scenario's oracle was one-sided (19463ffe).** It checked that the first of eight sockets was gone, so firmware with a cap of seven passed it, and firmware that evicted the newest instead of the oldest passed it too. The scenario now writes to all eight handles and collects the source ports that arrive. It requires that no port from the first four answers, and that every port from the last four does. Both bounds use source ports rather than handle numbers, for the reason your own comment gives. `Peer.drain_udp()` and a retry were added so a lost datagram is not read as a missing socket.

## Not changed, but I think it needs a decision

**4. Close-oldest fires on a failed `connect()`.** The code comment treats this as unavoidable ("the slot only exists once something has been closed"), but it is avoidable, and for TCP it is not an edge case. `connect()` to a host that refuses or does not answer is the normal error path, and each failed attempt now destroys one of the client's live sockets while creating nothing. A client retrying a connection to a down server loses all four.

It can be avoided without giving up either property you wanted:

```c
int socket = socket(AF_INET, type, 0);
if (socket < 0 && socket_count == NET_MAX_SOCKETS) {
    // The pool refused and this client is at its cap, so give up the oldest.
    evict_oldest();
    socket = socket(AF_INET, type, 0);
}
if (socket < 0) { *status = &c_status_no_socket; return; }
// connect, and on failure close `socket` only
...
if (socket_count >= NET_MAX_SOCKETS) evict_oldest();
track_socket(socket);
```

The client still sees its own limit rather than the pool running out, the cap is still enforced, and below the cap a pool exhausted by the firmware still answers `85` rather than eating a client socket. I did not push this because it changes the mechanics of a decision you flagged for confirmation, and it is more surface to re-verify on hardware than the two fixes above.

## Noted, not changed

**5. `close_socket()` untracks before `lwip_close()`.** If the close fails, the socket is open, no longer in the table, and therefore no longer reachable by the client or by `close_all_sockets()` on the next reset. `lwip_close()` on a valid descriptor does not realistically fail, and always releasing the slot stops a client pinning the table, so I think the current order is right. Worth one line of comment saying that is deliberate.

**6. A command queued before a reset is executed after `c64_reset()` has run.** The queue is FIFO and the flag is read at the top of the loop, so `[CMD_NEW_COMMAND, CMD_ABORT_DATA]` closes all sockets and then executes the older `OPEN_UDP`, leaving one socket owned by a program that no longer exists. The cap bounds it to one socket and four more opens clear it, so I do not think it is worth code.

## On what can be tested

The cap and the reset are both E2E testable and are now tested from both sides. The gate on unowned handles is not, for the case it exists for.

`NEVER_OPENED` covers a handle that was never valid, and that already answered `02,NO DATA: 9` before this change, because lwip rejects the descriptor itself. A handle that was valid and has since been closed also answered EBADF before, because lwip frees the socket on close. The case the gate is actually for is a stale handle whose number lwip has since given to a socket the firmware opened for itself, and the E2E harness cannot arrange that: it has no way to make syslog, ftpd, httpd or a streamer open a socket with a chosen descriptor number. So that behaviour ships on the argument in the header comment, not on a test, and the tests here should not be read as covering it.

I considered a host unit test under `software/test` instead. `network_target.cc` pulls in `network_interface.h`, which needs `ConfigurableObject`, `IndexedList`, `Fifo`, `UserInterface` and lwip's `pbuf_custom`, so it needs a replacement header rather than a link stub, plus a fake socket layer for seven lwip calls whose names collide with host libc, plus an extended `command_intf.h` stub. That is over 300 lines of scaffolding for about 40 lines of logic, and `make host_tests` is not wired into CI, so it would gate nothing. Not worth it.

I also did not add a TCP variant of either scenario. `open_socket()` takes the socket type as a parameter and everything else on both paths is type independent, so a TCP variant would exercise no firmware path the UDP one does not, at the cost of a `listen()` backlog change and more device time.

## Verification

Both commits build. `u64` (Nios II) and `u64ii` (RISC-V) built clean here: `update.u64` 5,389,352 to 5,389,376 bytes, `update.ue2` 8,559,580 to 8,559,636. I could not build `u2`, `u2plus` or `u2pl` locally, because their recipe runs `git show HEAD:global_makefiles/...` inside the container and a git worktree's `.git` pointer does not resolve there. That is a local tooling limitation, and CI covers those targets.

I did not run the E2E suite. My devices were in use, so the test change in 19463ffe is reviewed but not executed. Please run `uci-net-target` once before this lands.

chrisgleissner:
@JC-000 this answers your decision 2, and I have pushed it (`a447aa0b`) so you can see it rather than just read about it. Revert it if you disagree.

**The cap now refuses with `85,ERROR OPENING SOCKET` instead of closing the oldest socket.** Everything else in your PR stays: the socket table, the owned-handle gate, the cap of 4, and close-all on C64 reset.

## Why

**1. Consistency with the rest of the firmware.** Every other capped subsystem on the device refuses:

| | Cap | Behaviour at the cap |
|---|---|---|
| `socket_gui.cc:41` | `TELNET_MAX_SESSIONS 4` | "Too many connections, try again later." |
| `ftpd.cc:27` | `FTPD_MAX_SESSIONS 4` | refuses the session |
| `server.h:24` | `MAX_HTTP_CLIENT 4` | fixed slots |

Close-oldest would have been the only place in this firmware where a subsystem takes a resource back from its holder. That matters beyond taste: the next person to read `open_socket()` already knows how the other three behave, and a reviewer looking at all four sees one rule rather than four. Consistency here is maintainability, not style.

**2. It fails fast on the client's bug instead of hiding it.** A client that leaks handles has a defect, and the open it cannot complete is exactly where that defect is. Reporting it there is what lets the author find it. Close-oldest keeps the leaking client running and moves the symptom to a later `READ_SOCKET` or `WRITE_SOCKET` on a socket the client still believes it holds, which is both harder to diagnose and, from the C64 side, indistinguishable from a network fault.

**3. No new observable behaviour at all.** `85` is already what a client sees when lwip has no socket to give. A client at its cap now sees exactly what it saw before this PR when the pool ran out. Nothing new to document, and note that `doc/uci_network_target_v1.0.pdf` and `uci/network_target.rst` in the documentation repo list no failure status for `OPEN_UDP` at all, so anything new here would have been undocumented behaviour by definition.

**4. It does not weaken the fix for your actual bug.** Your repro in #808 is one socket leaked per program reload. Launching a PRG resets the C64: `C64_Subsys::dma_load()` calls `C64::start_cartridge()`, which does the same `C64_MODE = C64_MODE_RESET` write at `c64.cc:1167` that `C64::reset()` does at `:615`, and your own green run proves that write closes the sockets. So the reset hook reclaims your socket at the moment you reload, and the cap never engages for you. Your client holds one socket at a time (`net_abi.inc:35`, a single `uci_socket_open` flag), so it is nowhere near four.

## What it costs

One case genuinely gets worse: a client that leaks four handles **within a single run without ever resetting** now stops until a reset, where close-oldest would have carried it. Your pre-fix rekey path did this, which is what `timer.s:107` and `session.s:401` now fix on your side. I think a loud `85` at the moment the client misbehaves is the better outcome, and it is what the device already does everywhere else, but you have more standing than I do on how UCI clients behave in practice, so say if you disagree.

## The code got smaller

`open_socket()` loses the eviction, `close_oldest_socket()`, and the ordering that eviction needed, and gains one early return:

```c
if (socket_count >= NET_MAX_SOCKETS) {
    *status = &c_status_no_socket;
    return;
}
```

Built for u64 (Nios II) and u64ii (RISC-V): `update.u64` 5,389,348 bytes against 5,389,352 at `18dee7b6`, and `update.ue2` unchanged at 8,559,580. Slightly smaller than your version, while also carrying the `discard_read_reply()` fix from `9bf50704`.

## E2E changes

`abandoned-sockets-are-bounded` is rewritten, and it gets simpler, because handle numbers are now stable for the life of the scenario. The port-identity comparison existed only because eviction recycled handle numbers.

1. open `SOCKET_CAP` UDP sockets, all succeed
2. the next `OPEN_UDP` must answer `85,ERROR OPENING SOCKET`
3. all four originals must still accept a `WRITE_SOCKET`, which fails on any firmware that made room by closing one of them
4. `CLOSE_SOCKET` one, then one more open must succeed, which shows the cap counts what the client holds now rather than being a lockout

Step 2 is red on unpatched firmware, where the fifth open succeeds. Step 4 is the part that holds regardless of how many PCBs the firmware itself is using.

`reset-closes-uci-sockets` needed one change: it used to open 8 sockets after the reset, which the cap now refuses at the fifth. It now fills the cap again instead, which is the stronger oracle anyway. If the reset had merely forgotten the sockets rather than closing them, the client would still be at its cap and the first of those opens would answer `85`.

`failed-open-keeps-existing-sockets`, which I added in `614ab86d`, is deleted along with the eviction bug it was testing. `source_port()`, `Peer.drain_udp()` and `Peer.refused_tcp_port()` go with it.

**I could not run any of this.** Both my devices are in use, so the E2E changes are written and reviewed but not executed. Please run `uci-net-target` on both routes before this lands, and post the log.

## One more thing, unrelated to the cap

`xlar54/ultimateii-dos-lib` defines `NET_CMD_TCP_LISTENER_START/STOP` and `GET_LISTENER_STATE/SOCKET` (0x12 to 0x15) and ships `u-echoserver.c`. This firmware does not implement them, and neither does the documentation repo describe them, so they answer `21,UNKNOWN COMMAND` today. But if a listener ever lands, an accepted socket is created outside `open_socket()` and `owns_socket()` would refuse the client its own handle. I have added that invariant as a comment on the table in `network_target.h` so the next person to add a command sees it.

Thanks for the PR. The reset hook is the part that fixes #808, and it is a good fix.

chrisgleissner:
One more push, `3d2cc270`: every comment this PR adds is now at most three lines, in both the firmware and the E2E suite.

House style here keeps the reasoning in the PR discussion, where it is searchable and dated, and keeps the source comment to what a reader needs at that line. The `NET_MAX_SOCKETS` block went from fifteen lines to three, for example, and still carries the three things that are not obvious from the code: the answer is `85`, it is the same shape as `TELNET_MAX_SESSIONS`, and refusing is what puts the report on the open that caused the leak.

No behaviour changed. `update.u64` and `update.ue2` are byte-identical to `a447aa0b`.

chrisgleissner:
Hi @JC-000 ,

I have pushed final changes that came out of my review. Please have another look.

I would appreciate it if you could run all E2E tests again on your device as part of red/green TDD. All my devices are currently occupied by other PR runs as we are getting ready for the 3.15 release.

That is the last thing to be done before we can merge this.

Thanks
Christian

chrisgleissner (approved):
Looks good. 

I am currently in the process of performing final red / green E2E runs on this branch.

chrisgleissner:
Ran this on an Ultimate 64 Elite (core 1.4F) with JTAG deploys, baseline against patched. Baseline is the current `test-merge` tip `fcc2502d`; patched is your `3d2cc270` merged into it. The merge is clean and the u64 software build is clean, so nothing here is about mergeability.

One of the two new scenarios is a real regression test. The other cannot fail on this device.

## `reset-closes-uci-sockets` is red/green

FAIL on baseline, both routes: `handles [8, 9, 10, 11] were still open after the reset`. PASS on patched, both routes. This is the scenario that demonstrates the #808 fix, and it does its job.

## `abandoned-sockets-are-bounded` passes on unpatched firmware

It passed on the baseline build on both routes. The reason is that this device's spare UDP PCB count is exactly 4, the same as `NET_MAX_SOCKETS`, so the unpatched firmware refuses the 5th `OPEN_UDP` with the same `85,ERROR OPENING SOCKET` that the cap produces.

I measured the spare count directly, opening `OPEN_UDP` until refusal and then closing what was opened:

| Build | Sockets opened before refusal |
|---|---|
| baseline `fcc2502d` | 4, in 16 of 16 rounds |
| patched | 4, in 16 of 16 rounds |

The status byte is the same in both cases, so the scenario's oracle cannot attribute the refusal to the cap rather than to the pool.

Your rig had 5 spare, which is why the scenario went red for you: the cap refused one socket earlier than the pool would have. Whether this scenario can fail is therefore a property of how many UDP control blocks the rest of the firmware happens to hold, not a property of the change.

The same coupling makes it fail on correct firmware. In one patched run the REST route got only 3 opens before `85`, because something else in the firmware held a control block at that moment, and the hardcoded `SOCKET_CAP = 4` turned that into a failure:

```
[41] OPEN_UDP #4 without closing the previous ones ... FAIL
     abandoned-sockets-are-bounded [rest] stopped: OPEN_UDP 192.168.1.185:39779
     answered b'85,ERROR OPENING SOCKET' with reply b''
```

So the scenario cannot go red for the defect it targets, and it can go red when nothing is wrong.

Three ways out, in the order I would consider them:

1. Give the cap refusal a status of its own, distinct from `85,ERROR OPENING SOCKET`. That gives the test an oracle only the patched firmware can satisfy, and it also tells a client whether it hit its own limit or the device's.
2. Drop the scenario and say in the PR that `NET_MAX_SOCKETS` is a structural bound on the size of the tracking table rather than separately observable behaviour. The table is what makes close-on-reset possible, so the cap is still justified.
3. Lower `NET_MAX_SOCKETS` below the spare pool so the cap always refuses first. This needs an argument for a number that holds across devices, which is the part I do not think can be made to hold.

## A pre-existing flake in this suite, not from this PR

Seven scenarios on the native route failed together in 2 of my 3 runs, on both the baseline and the patched build: `reply-blocks-are-drainable`, `datagram-spans-reply-blocks`, `datagram-size-ceiling`, `udp-truncation-is-detectable`, `truncation-spans-reply-blocks`, `oversize-request-keeps-the-datagram`, `multi-block-state-does-not-leak`. In each, every datagram sent from the host reads back as `02,NO DATA: 11`, while the device's own probe datagram reaches the host in the same scenario. The REST route passed those same scenarios in every run.

This is not caused by your change, and it may be specific to my bench. It does mean `uci-net-target` is not a reliable gate here until it is understood, so I would not read a single green run of this suite as evidence either way.

## Runs

| Run | Build | Result |
|---|---|---|
| red | baseline `fcc2502d` | `reset-closes-uci-sockets` FAIL both routes; `abandoned-sockets-are-bounded` PASS both routes; 7 native scenarios FAIL |
| green 1 | patched | native route all OK; REST `abandoned-sockets-are-bounded` and `reset-closes-uci-sockets` FAIL on the 3-socket margin above |
| green 2 | patched | both new scenarios PASS both routes; the same 7 native scenarios FAIL |

JC-000:
Your native-route flake is not bench-specific and it is not a flake — it is a defect in the suite, and I have pushed a fix as `16d42416`. Your description identified it precisely: datagrams from the host read back `02,NO DATA: 11` while the device's own probe reaches the host in the same scenario.

## Mechanism

`Peer.learn_udp_peer()` writes a probe and takes *whatever datagram arrives next* as the reply, without draining first. A datagram left queued at the host by an earlier scenario is read as this socket's probe, so `udp_peer` is set to the source port of a socket that is already closed. Every `send_udp()` afterwards goes to that dead port. The device genuinely has nothing to read, which is why the status is a truthful `02,NO DATA: 11`, and why the device-to-host direction keeps working — that path never consults the learned address.

That also explains the ordering dependence: it appears when a route runs after another route has left datagrams queued, and a native-only run passes.

## Fix

Drain before probing, in `learn_udp_peer`:

```python
self.udp.setblocking(False)
try:
    while True:
        self.udp.recvfrom(2048)
except OSError:
    pass
finally:
    self.udp.setblocking(True)
```

This is the technique your `19463ffe` added as `Peer.drain_udp()` for `source_port()`; `a447aa0b` removed it along with the eviction test. `learn_udp_peer` never called it, so the hazard predates both.

**Result on a U64 Elite (core 1.4E, fw 3.15), whole suite, both routes: 147 checks, 0 failures.** The native route went from 7 failures to 11 of 11, reproducibly. Runs before the fix showed the 7 failures deterministically whenever REST ran first, and passed when native ran alone.

One caveat on my earlier runs: two of them had corrupted status reads on the REST route (`expected b'00,OK', got b'00'`). That was another session on my bench driving the same UCI interface, confirmed from both sides by comparing run windows. This suite takes no device lock, so it neither excludes nor is excluded by the harness other tooling here uses. The clean run above was taken while holding that lock. Worth knowing before treating any single run of this suite as a gate.

## On `abandoned-sockets-are-bounded`, you are right and I withdraw my red for it

My rig had at least 5 spare UDP PCBs, so the unpatched build accepted a 5th `OPEN_UDP` and the scenario went red — a property of my bench, not of the change, exactly as you say. With 4 spare it cannot discriminate, and your 3-open failure on correct firmware shows it can also go red when nothing is wrong.

I would take your option 1: give the cap refusal a distinct status. It is the only one of the three that makes the bound observable, and observable is what makes it testable.

Worth naming the cost, since it reverses something you argued for the refuse-at-cap design: "no new observable behaviour at all — `85` is already what a client sees" was a real virtue, and this trades it away. I think that is the right trade, because a bound that cannot be distinguished from pool exhaustion cannot be regression-tested, and a client that cannot tell its own limit from the device's cannot respond sensibly to either.

If you would rather not add a status before 3.15, option 2 is fine by me and I will make the change: drop the scenario, and say in the PR that `NET_MAX_SOCKETS` bounds the tracking table that makes close-on-reset possible rather than being separately observable. `reset-closes-uci-sockets` is red/green on both our benches and is the scenario that demonstrates #808.

Tell me which you want and I will push it.

chrisgleissner:
@JC-000 I am investigating the E2E test failures further. For the time being, I moved this back to Draft mode.

Thanks for the fix above btw - I will try it now.

To avoid merge conflicts etc. please hold off on further commits on this branch for now.

Also, if you are interested in this change hitting the imminent 3.15 release, I'd rather keep changes now frozen and not introduce new result types. Otherwise we may have to park it and it will then hit the next release in half a year.

Thank you.

JC-000:
Understood, holding off — no further commits from me.

One thing in case it crossed with your message: `16d42416` is already on the branch and fixes those seven native failures. `learn_udp_peer` was taking a stale queued datagram as its probe reply and learning a closed socket's port, which is why host-to-device reads returned `02,NO DATA: 11` while the device's own probe still arrived. Full suite, both routes, 147 checks, 0 failures with it. Details in my comment above — worth a look before you spend bench time on it.

chrisgleissner:
Thanks @JC-000 , appreciated. I saw your fix and it did help.

I am currently reworking the PR a bit to be truly minimally invasive whilst still hopefully meeting your core requirements. No protocol changes, no new status codes, etc. Main reason is to ensure your change can land in 3.15 with minimal risk.

As anything in life, and all the more in software, we can always improve on it and change it. To  be then released ca. half a year from now or whenever 3.16 lands.

chrisgleissner:
Proposing that this PR keep only the close-on-reset mechanism and drop the socket cap, so that what ships before 3.15 is the smallest change that addresses the issue.

`16d42416` is confirmed here. With the drain in place, a baseline build passes all nine other scenarios on both routes, and only `reset-closes-uci-sockets` fails. The seven native-route failures reported earlier came from datagrams left queued at the host, not from the device.

## Measurement of the cap

On this Ultimate 64 Elite the number of spare UDP control blocks is 4, the same value as `NET_MAX_SOCKETS`. Opening `OPEN_UDP` until refusal, then closing what was opened:

| Build | Sockets opened before refusal |
|---|---|
| baseline `fcc2502d` | 4, in 16 of 16 rounds |
| with the cap | 4, in 16 of 16 rounds |

The count and the status byte are the same on both builds. On a device with 5 spare control blocks the cap refuses one socket before the pool does, which is the case where the scenario can distinguish the two builds. On a device with 4 it cannot.

The effective limit is also not fixed. One run refused at 3, because the firmware held a control block at that moment. What a client meets is `min(NET_MAX_SOCKETS, free pool)`, and the second term varies at runtime.

Reserving pool headroom for the firmware's own UDP users in `syslog.cc` and the streamers was considered as a separate justification for the cap. With a pool of 8 and a firmware baseline near 4, a cap of 4 still allows a client to take every remaining control block, so it reserves nothing for a streamer that starts later.

## Why the cap is dropped rather than given its own status

`alloc_socket()` already returns -1 at exhaustion, and `open_socket` already reports that as `85`. A cap adds a second place that decides the same question, and the two can disagree. The run that refused at 3 is that disagreement.

`TELNET_MAX_SESSIONS` and `FTPD_MAX_SESSIONS` bound listeners that own an accept loop and arbitrate between several remote clients. The UCI network target has one client, the C64, and hands out sockets on request, so a cap there limits that single client rather than arbitrating between several.

The part of those listeners that does transfer is their reclamation path, which is independent of the cap in each case: telnet and ftpd reap through TCP keepalive, httpd through `HTTP_CONN_IDLE_TIMEOUT`. Each detects that a session has ended and releases what it held. For the UCI target the session ends when the C64 resets, and close-on-reset is that detector.

On status codes, each of those listeners states a reason when it refuses: ftpd sends `421` from RFC 959, telnet sends its own text. A cap that refuses with `85`, which already means the pool is exhausted, gives two conditions one code. So the cap needs either its own status or removal. Removal is the smaller change, and it is the one that needs no new protocol surface.

## What the change keeps

This is option 2 of the three hardenings listed in #808:

- `NetworkTarget` tracks the sockets it opens. The table is sized to `MEMP_NUM_NETCONN`, so it holds every socket lwip can create and no client meets a limit of its own.
- It reads, writes and closes only sockets in that table.
- It closes all of them, and discards a pending read reply, when the C64 resets.
- `static_assert(NET_MAX_SOCKETS >= MEMP_NUM_NETCONN)` fails the build if `lwipopts.h` ever outgrows the table. Without it an untracked socket would survive a reset, which is the condition #808 describes, and no test would detect it.

Client-visible behaviour on open is unchanged. An open is refused only when lwip refuses, with the same `85` as before this PR. There is no new status, no UCI extension and no documentation change. `abandoned-sockets-are-bounded` is removed, since it covered a cap that no longer exists. The cost is 48 bytes of bss.

## Scope before 3.15

A new status code is protocol surface that cannot be withdrawn once clients depend on it, and it needs a matching change in `GideonZ/1541u-documentation`, whose `uci/network_target.rst` currently documents neither `85` nor `11,ERROR ON CONNECT` nor `86,INTERNAL ERROR`. Days before a release, the smaller change is the one with less to get wrong.

A cap or an idle reaper can be added later on its own merits, and the tracking table added here is what either would be built on.

## Not addressed

A client that leaks sockets and never resets is not bounded. It exhausts the pool and receives `85`, as it does today. The difference is that any C64 reset now recovers the interface, which removes the power cycle described in #808.

## Verification

Ultimate 64 Elite, core 1.4F, JTAG deploys of two builds differing only by this change. Baseline is `test-merge` `fcc2502d`. Both builds carry the same test file, so the only variable is the firmware.

`reset-closes-uci-sockets`, both routes, ten runs per build:

| Run | baseline REST | baseline native | with change REST | with change native |
|---|---|---|---|---|
| 1 | FAIL | FAIL | OK | OK |
| 2 | FAIL | FAIL | OK | OK |
| 3 | FAIL | FAIL | OK | OK |
| 4 | FAIL | FAIL | OK | OK |
| 5 | FAIL | FAIL | OK | OK |
| 6 | FAIL | FAIL | OK | OK |
| 7 | FAIL | FAIL | OK | OK |
| 8 | FAIL | FAIL | OK | OK |
| 9 | FAIL | FAIL | OK | OK |
| 10 | FAIL | FAIL | OK | OK |
| **total** | **10/10 FAIL** | **10/10 FAIL** | **10/10 OK** | **10/10 OK** |

The baseline failure is the scenario's first assertion, on every run:

```
reset-closes-uci-sockets [rest] stopped: handles [8, 9, 10, 11] were still open
after the reset; the program that opened them is gone, so nothing else can ever
close them
```

Its second assertion fails on baseline as well. Refilling after the reset answers `85,ERROR OPENING SOCKET`, because the leaked sockets still hold the pool.

Whole `uci-net-target` suite, both routes, on the same two builds:

| Build | Result |
|---|---|
| baseline `fcc2502d`, run 1 | 19 scenarios OK, `reset-closes-uci-sockets` FAIL on both routes |
| baseline `fcc2502d`, run 2 | 19 scenarios OK, `reset-closes-uci-sockets` FAIL on both routes |
| with this change, run 1 | 21 of 21 scenarios OK, 133 checks, 403 s |
| with this change, run 2 | 21 of 21 scenarios OK, 133 checks, 403 s |

The nine scenarios unrelated to this change pass on both routes on both builds, which is the state `16d42416` restored.

Pushed as `acdb9686`.

chrisgleissner:
Thanks @JC-000 for your contribution. I appreciate your engagement on this issue. I hope the new behaviour is useful for you.

As mentioned, we are closing down for the 3.15 release, so any more work is unlikely to be included in this release, and it may also take a bit longer going forward to get PRs merged. 

However, I hope you will continue to contribute and I wish you much success with C64 Wireshark. Sounds very exciting. Please share once it's ready to be used.

PR #815, opened 2026-08-28
Wake On Wi-Fi: a magic packet switches the machine on while it is off (UE2/C64U)

chrisgleissner:
Wake on Wi-Fi, the follow-up to #795: while the machine is off but still has input power, a standard magic packet addressed to the control module's station MAC switches it on.

## Authorship

All of this work is by @enver-haase. Every commit carries him as author, and the description below is his, edited only where the base branch changed what it had to say. The original pull request is enver-haase/1541ultimate#1, on branch `c64u-wake-on-wifi` in his fork.

That pull request targeted `c64u-poweron-behaviour` inside the fork, so no `pull_request` event ever reached this repository and CI never built it. #795 has since been squash-merged, so the thirteen Wake on Wi-Fi commits were rebased onto `test-merge` and pushed here, which is what gets them built. The rebase was clean, with no conflicts and no content changes. He is busy with other work for a while, and asked that the branch not be held up for that; follow-ups can be taken from his fork branch and pushed here the same way.

**Disclosure**, as on #795 and in his own words: written with Claude Code, Opus 5 at high reasoning effort, directed and reviewed by him. The adversarial review was run with the brief given on #795, minimal churn and changes consistent with the existing code base, before the original pull request was opened. It produced two changes, both in `u64ctrl: wake the machine on a magic packet while it is off`: a guard so the watcher is not armed before `wifi_init()` has recorded the input function it chains to, and a comment that named the wrong task as writing `wake_armed`.

## Fixes from the review on this branch

The commits from `1c906216` down are @enver-haase's Wake on Wi-Fi work. The commits after it come from two adversarial reviews here and are not his; they are listed separately so the feature and the review are easy to tell apart.

- **The arming decision is now host testable.** Whether the frame path is watched (machine off, frame path known, setting enabled) lived in `wifi_modem.c`, which no host build compiles, so nothing covered it. It is the predicate `power_should_watch_for_wake()` in `power_state.c` now, and `wifi_modem.c` acts on it.
- **The cold start is covered.** The watcher is armed from the control module's own start as well as from a power transition, and nothing exercised that path: every run so far put the machine off over REST or by hand. The host checks now cover it, and the E2E suite has a scenario for it that cuts mains with the power on behaviour at `Off` and then wakes the machine with a packet. That scenario is skipped unless a socket is scripted.
- **No answer from the module is no longer reported as a refusal.** A starved TX buffer returns 0, and so does success, so one unanswered call at boot disabled the setting for the whole session and printed that the module lacks the feature. Both settings are asked for twice before the item is taken away, and the message says no usable answer arrived instead of naming a cause.
- **The write is verified.** `wifi_set_wake_on_wifi()` could not report a starved buffer either, so a push that never reached the module looked like success and left the menu and the module disagreeing until the next boot. The value is read back now.
- **A failed E2E scenario no longer leaves the machine off.** `check()` re-raises, so a failure between switching the machine off and the press that follows skipped the press. The machine stayed off, the setting restore failed with it, and so did the runner's teardown and every suite after it. `recover_if_off()` runs on the way out of both `u64ctrl` suites.
- **The MAC lookup works without net-tools.** `arp` is no longer installed by default on current distributions, and the missing binary raised `FileNotFoundError` out of the generic handler rather than saying to pass `--mac`. `ip neigh` is tried first, `arp` second.
- **Both suites read their actuators from the environment**, so options the runner does not forward can still be given: `U64_POWER_BUTTON_CMD`, `U64_POWER_OFF_CMD`, `U64_POWER_ON_CMD`, `U64_WOL_MAC`, `U64_WOL_BROADCAST`, `U64_WOL_PORT`.
- **The matcher's scan window is pinned on both sides.** Only the offset that must not match was checked; the last offset that must still match is checked now.

A second, independent adversarial review (OpenAI Codex, `gpt-5.6-terra`, high reasoning) then went over the branch, and three of its five findings led to further changes:

- **The menu could say Enabled while the module held Disabled.** The verified write reported the failure to the console, but the config item, the REST answer and the flash all kept the value the module had refused. The item now goes back to what the module reports, with a message on the machine's own screen; a module that does not answer at all disables the item, as it already does at boot. What is in flash is deliberately left alone, so a choice the module refused is pushed down again at the next boot.
- **The watcher could be armed in name only.** Attaching the lwIP netif resets `netif->input`, so a Wi-Fi restart after the machine went off would leave the watcher installed in a slot that no longer exists. The module remembers the last known machine state and re-arms on `WIFI_EVENT_STA_START`, but only when the machine is off: while it is on that slot belongs to the application's own hook.
- **The E2E read back proved less than it claimed.** It shows the application took the value, not that the control module stored it; only what the machine does with a packet shows that. The comment says so now, and the failure text and docstring name a control module older than 1.14 as the first suspect, since that is the one case the suite cannot detect for itself.

Two findings were considered and deliberately not acted on. Writing the last power state to NVS on every transition was called a flash-wear risk: the write is skipped when the value has not changed, ESP-IDF's NVS wear-levels, and a machine switched on and off ten times a day reaches a few tens of thousands of writes over a decade, so the behaviour stays as #795 shipped it. Exposing the control module version over REST, so the E2E suite could skip on an old module instead of failing, would change a shared API response and its generated specification for a test convenience, and it could not be validated here without a device; the suite says what to check instead.

Host checks after these changes: `wolmagic` 17/17, `powerstate` 45/45, `runner-policy` 71/71, pyflakes clean. The new arming checks were confirmed to fail against a deliberately broken predicate before passing against the real one. The E2E changes have not been run on hardware: no device was available. What was verified without one is that both suites parse their arguments, import cleanly, take the skip path against a host that does not answer, and that the MAC lookup returns the right address on synthetic `ip neigh` output, reports a clear message when no entry exists, and does not raise when neither tool is installed.

## The shape

- A separate **Wake On Wi-Fi** item, Enabled/Disabled, in the Power Settings group, defaulting to Disabled.
- Held in the ESP32's NVS, one key beside the power on behaviour, and synchronised over the existing UART RPC: `CMD_SET_WAKE_ON_WIFI` (0x18) and `CMD_GET_WAKE_ON_WIFI` (0x19), both ids free on this base.
- A receive matcher that turns a standard magic packet for the station MAC into the existing `BUTTON_ON` event.
- An explicitly disabled radio still means no wake, and needs no code to say so: `CMD_WIFI_DISABLE` calls `esp_wifi_stop()`, so no frame arrives to match.
- Module version to 1.14, or the updater never flashes it, which is the trap that cost an evening on 1.13.

Three details worth a reviewer's eye:

**Where the matcher sits.** In `netif->input`, the slot `hacked_recv` occupies while the machine is on, so a frame is seen before the module's own stack consumes it. Unlike `hacked_recv` it does not take the frame over: matched or not, it is handed on to the stack that was there. That is what keeps the association, the DHCP lease and SNTP alive while the machine is off, and what answers the ARP a unicast magic packet needs.

**What it matches.** The pattern, not a port: six 0xFF bytes followed by the MAC sixteen times, anywhere in the frame. That accepts UDP port 9, port 7 and raw ethertype 0x0842 alike, which is what "a standard magic packet" has to mean when there is no application to be particular for. It lives in `wol_magic.c`, free of ESP-IDF and lwIP, which is what makes the code that ships host-testable.

**When it is armed.** On every transition into the off state and once at startup, so a machine that cold starts into off, whether from mode `Off` or from `Last State` with the machine off when the power was lost, is watched exactly like one switched off later. The setting is read on each of those, so a change in the menu needs no second message to the module. It is disarmed after the event is posted, because wake tools send the packet several times over.

One finding from #795 decides what users may expect, and it is now in the menu's help text rather than only in a comment: **only Wi-Fi can do this.** The wired jack cannot, on this hardware, while the machine is off. The RMII PHY and its MDIO sit in the FPGA power domain, which is down, and the control module has no wired path of its own.

## Tests

Host side, no hardware, `gcc` only:

```
make host_tests
make -C target/pc/linux/wolmagic test     # 15 checks, new
make -C target/pc/linux/powerstate test   # 37 checks, was 22
```

The matcher's 15 checks cover the three encapsulations, a SecureOn-style trailer, a packet for another station, a frame one byte short, fifteen of sixteen repetitions, all-ones and all-zero frames, a broadcast destination whose own 0xFF bytes precede the pattern, a target MAC of broadcast or zero that no station owns, and a pattern that ends past the scan window. The power-state checks grow by the new setting: default, both writes, a refused out-of-range write, a stored value out of range, and that the two settings share a namespace without disturbing each other.

End to end, `tests/e2e/u64ctrl/wake_on_wifi_test.py`, registered as `wake-on-wifi`, manual:

```
./run-tests --suite wake-on-wifi --manual c64u
```

It needs no mains interruption. The machine is put off over REST, which leaves the module powered and associated, exactly the state the feature is for. The order of the scenarios lets the harness recover the machine itself where it can: the packet for another station has to be ignored, which leaves the machine off, and the packet for this station is what brings it back. Only the last scenario, with the setting disabled, ends with a machine no packet may revive, which is what makes the suite manual, or takes `--power-button-cmd`, the actuator option that came out of the review comment on the power cycle suite.

Two conditions the suite cannot check for itself are stated in its docstring and in the failure text of the wake check: the device has to be on Wi-Fi, and the harness has to share its broadcast domain. Reporting correct firmware as broken would be worse than saying so.

## What is verified

- **On hardware.** A C64 Ultimate running a custom build on top of the Commodore 1.1.0 branch, carrying these commits, on Wi-Fi. `wake-on-wifi` passes 4/4: a magic packet for a neighbouring MAC left the machine dark for the full 90 second window, the packet for its own MAC brought it up in 3.4 seconds from switched-off, and with the setting Disabled that same packet was ignored for 90 seconds. Then, off the harness entirely: the setting switched on in the menu, the machine switched off by hand, a packet sent from a laptop, and the machine came up. So the value travels from the menu through the RPC into the module's NVS and survives the power-off, which is the path the suite does not exercise.
- **Not on Elite II hardware.** The run above is the same source on the 1.1.0 line, because that is the line the C64 Ultimate ships on. `target/u64ii` on this branch packages `u64e2_50t`/`u64e2_100t` gateware instead, which the author has no machine for. An Ultimate 64 Elite II run on a build of this branch is the gap, and it is the same gap #795 had until @Retro1968 tested it.
- Both halves build in the image from `docker/Dockerfile`, from a clean tree, with no warnings from any file this branch touches.
- Host checks on this rebased branch: `wolmagic` 15/15, `powerstate` 37/37, both reachable from `make host_tests`.
- Python: pyflakes clean; `runner-policy` 71 checks, `esp-depends` and `transport-usage` pass. The Python sender and the C matcher were cross-checked byte for byte, so the suite is not testing its own idea of the format.

## The adversarial review, and what it changed

Five findings, one commit each, all in this branch:

- the 256-byte scan buffer sat on the receive path's task stack, held across the call into the stack below. It is now `static`, since `netif->input` is delivered from one task;
- `wake_armed` was cleared before `extern_button_event()`, whose post can fail and was unchecked, so a full queue lost the wake permanently while the retry copies found the watcher disarmed. It now posts first and disarms on `pdTRUE`;
- `wifi_get_wake_on_wifi()` cannot report a starved TX buffer distinguishably from success, because `BUFARGS` returns `pdFALSE`, which is 0, which is also success. The value it would have written is now the witness. The same hole existed for the power mode, and the fix for both is in this branch;
- the matcher's host checks were unreachable from `make host_tests`;
- one check asserted an interference direction it never exercised.

The two `u64_config.cc` commits were kept out of #795 deliberately: they touch firmware, and #795 had already been reviewed and built for testers, so its firmware stayed what was reviewed. They belong here instead.


chrisgleissner:
Adversarial review of this branch is done, and the findings are fixed in the five commits after `1c906216`. The feature itself is @enver-haase's work and is unchanged; these are review fixes only, kept as separate commits so the two are easy to tell apart.

**Firmware**

1. *No answer from the control module was reported as a refusal.* A starved TX buffer returns 0, and so does success, so one unanswered call at boot disabled the setting for the rest of the session and printed that the module does not support the feature. Both settings are now asked for twice before the item is taken away, and the message says that no usable answer arrived rather than naming a cause.
2. *The write had no witness.* `wifi_set_wake_on_wifi()` cannot report a starved buffer either, so a push that never reached the module looked like success and left the menu and the module disagreeing until the next boot. The value is read back now.
3. *The arming decision could not be tested.* Whether the frame path is watched - machine off, frame path known, setting enabled - sat in `wifi_modem.c`, which no host build compiles. It is the predicate `power_should_watch_for_wake()` in `power_state.c` now, and `wifi_modem.c` acts on it.

**Tests**

4. *The cold start was covered nowhere.* The watcher is armed from the control module's own start as well as from a power transition, and every run so far put the machine off over REST or by hand, which is a transition. That is the path a user meets after a real power loss with the power on behaviour at its default. Host checks cover it now, and the E2E suite has a scenario that cuts mains with the mode at `Off` and then wakes the machine with a packet; it is skipped unless a socket is scripted.
5. *A failed E2E scenario left the machine off.* `check()` re-raises, so a failure between switching the machine off and the press that follows skipped the press. The machine stayed off, the setting restore failed with it, and so did the runner's teardown and every suite after it - none of which resemble the failure that caused it. Both `u64ctrl` suites now press on the way out.
6. *The MAC lookup needed net-tools.* `arp` is not installed by default on current distributions, and the missing binary raised `FileNotFoundError` out of the generic handler instead of saying to pass `--mac`. `ip neigh` is tried first, `arp` second.
7. *Suite options the runner does not forward* can now come from the environment: `U64_POWER_BUTTON_CMD`, `U64_POWER_OFF_CMD`, `U64_POWER_ON_CMD`, `U64_WOL_MAC`, `U64_WOL_BROADCAST`, `U64_WOL_PORT`.
8. *The matcher's scan window was pinned on one side only.* The offset that must not match was checked; the last offset that must still match is checked now.

**What held up:** the matcher is bounds safe at both edges of its window and rejects broadcast and all-zero MACs; `CFG_WAKE_ON_WIFI = 0x64` is unique across every config id in the tree; `en_dis` maps 0/1 onto `WAKE_ON_WIFI_DISABLED`/`ENABLED`; the RPC structures follow the existing `rpc_get_connection_resp` shape; `esp_depends.py` globs `u64ctrl/main`, so the new file moves the CI cache key; and the dropped-link retry uses `esp_wifi_connect()` rather than a stop and start, so it cannot silently unhook the watcher.

**Verification.** Host checks: `wolmagic` 17/17, `powerstate` 45/45, `runner-policy` 71/71, pyflakes clean. The new arming checks were confirmed to fail against a deliberately broken predicate before passing against the real one. The E2E changes have not been run on hardware, because no device was available; what was checked without one is that both suites parse their arguments, import cleanly, take the skip path against a host that does not answer, and that the MAC lookup handles a real neighbour entry, a missing entry and a missing binary. CI is building the branch.

chrisgleissner:
Second adversarial review, this time by a different model (OpenAI Codex, `gpt-5.6-terra`, high reasoning effort) so the findings are not from the same head that wrote the first round. Five findings, three fixed, two rejected with reasons. The feature itself is still @enver-haase's work and is unchanged by these commits.

**Fixed**

1. *The menu could say Enabled while the control module held Disabled.* The verified write from the previous round reported a refusal only on the debug console. The config item, the REST answer and the saved settings all kept the value the module had refused, so a user could switch waking on, save it, and get nothing. The item now goes back to what the module actually reports, with a message on the machine's own screen; a module that does not answer at all disables the item, as it already does at boot. Flash is deliberately left holding what was asked for, so the next boot pushes it down again and a module that was merely busy takes it then. The item then says what the machine will do now, the flash says what was asked for, and the comment in the source says why they differ.
2. *The watcher could be armed in name only.* Attaching the lwIP netif is where `netif->input` goes back to what esp_netif put there, so a Wi-Fi restart after the machine went off left the watcher pointing at a slot nothing calls. The module now remembers the last known machine state and re-arms on `WIFI_EVENT_STA_START`, and only when the machine is off: while it is on that slot belongs to the application's own hook, and taking it over would cut the machine off the network.
3. *The E2E read back proved less than the comment claimed.* Reading the item back shows the application took the value, not that the control module stored it. Only what the machine does with a packet shows that. The comment says so now, and both the wake failure text and the suite docstring name a control module older than 1.14 as the first suspect, because that is the one case the suite cannot detect: the menu greys the item out, but `cfg->disable()` is a menu affordance and the REST store still serves it.

**Considered and not acted on**

4. *Flash wear from writing the last power state on every transition.* The write is skipped when the value has not changed, ESP-IDF's NVS wear-levels across its pages, and a machine switched on and off ten times a day reaches a few tens of thousands of writes over a decade. The behaviour stays as #795 shipped it.
5. *Exposing the control module version over REST* so the suite could skip on an old module rather than fail. That changes a shared API response and its generated specification for a test convenience, and it could not be validated here without a device. The suite says what to check instead.

A verification pass by the same model over the fixes found one further point, which is now written into the source: after a refused write the item and the flash disagree on purpose, and the reason is stated so the next reader does not "fix" it.

**Verification.** `wolmagic` 17/17, `powerstate` 45/45, `runner-policy` 71/71, pyflakes clean. The firmware changes have not been run on hardware and the E2E changes have not been run at all: no device was available. CI is building the branch.

chrisgleissner:
Hi @enver-haase ,

I have not yet had a chance to test this PR on real hardware. Given you have done the work and you have previously verified it to work with the E2E tests you kindly added, would you mind re-testing it? We now have an official firmware build. 

This final check, especially after the two reviews and related fixes that I pushed ysterday evening, is the last remaining step to getting this merged.

No worries if you are busy, but I saw a comment from you a few hours ago, so wanted to check.

Thanks
Christian

enver-haase:
Happy to re-test — but one thing first, because I think it decides which binary I should be
flashing.

This branch packages Elite II gateware, and the 1.1.0 line does not:

```
this branch     external/u64e2_50t.bit, external/u64e2_100t.bit
the 1.1.0 line  external/u64_mk2_artix.bit
```

My machine is a C64 Ultimate, which ships on the 1.1.0 line. And the updater will not stop me
from doing the wrong thing: `update_u64ii.cc` selects on the FPGA part alone —

```c
const char *fpga_types[] = { "5CEBA2", "5CEBA4", "XC7A50T", "XC7A100T" };
```

— so a C64U, being a 7A50T, takes the `u64e2_50t` branch and gets Elite II runtime gateware
written to it. `"FPGA Type not supported."` only fires for a part outside that list of four.
That is the gap the description already names under **Not on Elite II hardware**, seen from
the other side: I have no Elite II, and the package built here is for one.

So before I flash anything: is there a build of this branch carrying the C64U gateware, or
should we ask @GideonZ whether writing the Elite II package to a C64U is safe? I would rather
ask than find out.

What I can do meanwhile, without that answer: four of your six review commits touch firmware
(`u64_config.cc`, `power_state.c/.h`, `wifi_modem.c`); the other two change only `tests/e2e`.
I have those four on a custom build on top of the Commodore 1.1.0 branch, so with the gateware
this machine actually runs, built and host-verified — `wolmagic` 17/17, `powerstate` 45/45. A
`wake-on-wifi` run against that exercises exactly your fixes on the board they would run on
here.

It is not the artifact you asked for, and it would not close the Elite II gap. But it would
tell us whether the six commits broke anything, which is the part I can answer today. Say the
word and I will run it and report.

Retro1968:
If I could help with a test on my UE2 (with working WiFi) please tell me what to do.

chrisgleissner:
Hi Enver, 

thanks for reaching out. 

As of firmware 3.15, the UE2 firmware built by the CI run of this repo can also be flashed on the C64U as I confirmed with Gideon. As per what he told me, the main regression of this is that the LED lights will stop working until the next time you reflash an original C64U firmware. However, I have never done this myself and I would only do it if you had an FT232H-based JTAG adapter so you can revive your device if anything were to go wrong. Gideon provided a tool in this repo specificaly for that revival scenario. 

Up to you - I am not sure if you have such an adapter. I always meant to build mine, but have not yet had a chance to do so, thus I am in the same boat.

It would help already if you could just use the test approach which you used before, and document it step by step with the actual commands you used, so I can understand whether your test results are transferable to the UE2.

@Retro1968 You very kindly tested Enver's other PR on a real UE2. Would you be willing to give this one a go? You can find it at https://github.com/GideonZ/1541ultimate/actions/runs/33220255643 Please note however that there is a small but non-zero risk that this change may break things. I don't think it is any higher than with the other PR that you have already tested.

As we are very close to cutting the 3.15 release, your help would be much appreciated.

Thanks
Christian

chrisgleissner:
Hi @Retro1968,

thanks for reaching out! Much appreciated indeed.

You can find the firmware at https://github.com/GideonZ/1541ultimate/actions/runs/33220255643

Then, please just install it like last time.

### What this one adds

Wake On Wi-Fi, the follow-up to #795. With the machine switched off but still connected to mains, a magic packet sent to its Wi-Fi MAC switches it back on.

There is one new setting: **Power Settings > Wake On Wi-Fi**, Enabled or Disabled, right below "Power On Behavior". It defaults to Disabled.

Two things to know before you start:

- **Wi-Fi only.** The wired jack cannot do this on this hardware. Its PHY is powered down together with the machine.
- The install also updates the ESP32 module to V1.14. That happens by itself, just give the first boot a moment.

### How to test it

You need two things: the device's **Wi-Fi MAC** and any wake tool.

The MAC is easiest to get from your router's client list, or with `arp -a` (Windows, macOS) or `ip neigh` (Linux) while the machine is still on. For sending, `wakeonlan <mac>` on Linux, "WakeMeOnLan" on Windows, or any "Wake On LAN" app on the phone. Port and encapsulation do not matter, the matcher looks for the pattern.

1. Connect the Ultimate to Wi-Fi, not to the wired jack.
2. Set Power Settings > Wake On Wi-Fi to Enabled and save.
3. Switch the machine off with the power button, leaving mains connected.
4. Send the magic packet from a PC or phone on the same network. The machine should come up within a few seconds, it was 3.4s here.

Then the two negative checks, which matter just as much:

5. With the machine off, send a packet to some other MAC. Nothing should happen.
6. Set Wake On Wi-Fi back to Disabled, switch off, send the packet again. It should stay off.

And if you have a minute for the cold start case:

7. Set Power On Behavior to Off, pull the mains, plug it back in. The machine stays off. Now send the packet, it should still wake.

All of the above passed on a C64 Ultimate. **Elite II is untested**, which is exactly the gap you closed for #795, so that is the interesting part.

The PR description has the full details if you want them. The feature itself is @enver-haase's work.

Thanks a lot!

Retro1968:
Unfortunately, I don't have such an adapter. I'm happy to help, but of course I don't want to damage my UE2 during testing.

If there's a way to test it with the same minimal risk as before, I'd be happy to help.

chrisgleissner:
Hi @Retro1968,

Completely fair question, and thank you for asking it rather than just going ahead. Here are some notes after I carefully compared the prior PR by Enver (which you already tested) and this new one:

**First, the JTAG adapter warning was not about you.** That was a reply to Enver, who owns a C64 Ultimate, not an Elite II. He was asking whether he could flash the Elite II package onto his C64U. That is a different product, and that is where the caution about the LEDs and needing a recovery adapter came from. You have a UE2, and the build linked above is the UE2 package. Right firmware, right machine. None of that applies to your situation.

**Second, this change is smaller than the one you already tested.** PR #795, the one you ran on your UE2, changed the line in the control module that decides whether the machine powers up at boot at all. This PR does not touch that file. What it adds sits after the power decision, not before it.

**Third, the new feature is switched off until you switch it on.** Wake On Wi-Fi defaults to Disabled, and the module falls back to Disabled if it ever reads anything it does not recognise. While it is off, the new code amounts to one small read from the module's storage and a pointer comparison that never matches. The packet matcher is never even installed. So until you go into the menu and enable it, you are effectively running the firmware you already tested plus two commands nothing calls.

**On reverting.** This is the part I would want reassurance about too, so I checked it specifically. The updater decides whether to reflash the Wi-Fi module by comparing version numbers for inequality, not for "newer". Flashing an older firmware therefore sees 1.14 against 1.13 and writes the old module firmware back automatically, the same way going up wrote the new one. And the module is put into its flashing mode by the FPGA pulling its boot pins, not by asking the module nicely, so it can be reflashed whatever it happens to be holding. Going back is the same operation as going forward.

The new setting is stored with an ID that older firmware skips over, so rolling back leaves nothing behind that confuses it.

**What I will not claim.** There is one risk in any firmware update, this one included: losing mains power while the "Flashing ESP32" progress bar is on screen. That is the same risk you already took with #795 and with every update before it. This PR does not make it larger. Just leave it plugged in until the update finishes and reboots.

There is also the ordinary chance that the feature itself does not work properly on Elite II, since nobody has run it on one yet. That is exactly what the test is for, and a feature that does not work is a comment on this PR, not a damaged machine.

**Also worth saying:** Enver has already tested this on his own hardware, just as he did with #795. Four out of four scenarios passed on his C64 Ultimate: a packet for a neighbouring MAC left the machine dark for a full 90 seconds, the packet for its own MAC woke it in 3.4 seconds, and with the setting disabled the same packet was ignored for 90 seconds. He then repeated it by hand, off the test harness. So the feature is known to work on the same source; what is missing is a run on Elite II gateware, which is the same gap you closed for #795.

So: my honest assessment is that the risk of your UE2 not coming back is very low, and no higher than the update you have already done once. If you would rather sit this one out, that is a completely reasonable call and nobody will think less of you for it. But if you are willing, the steps in my earlier comment are all there is to it, and you can stop after step 4 if you like.

Thanks either way.

chrisgleissner:
@Retro1968 If you tell me more about your environment (Windows / MacOS / Linux) and you want me to, I can also write a command line script for you so you can more easily test it. Just met me know. The testing steps are in my earlier message.

Also, any questions, just let me know.

Many thanks.

Retro1968:
I'm going to flash the file "update_v3.15-98-g7ccf1498.ue2" now and will get back to you afterwards. Let's continue from here step by step.

chrisgleissner:
@Retro1968 Sorry for the last minute announcement, but would you mind using https://github.com/GideonZ/1541ultimate/actions/runs/33243805039 instead once built?

I only removed needless line ending changes in one file. No functional change, just cleanup. 

If you have already flashed the firmware from the earlier link by the time you read this, then don't worry.

Also, my next comment has a more detailed walk-through of the test steps, including a command line script for Windows.

chrisgleissner:
Hi @Retro1968,

thank you again. Here is everything you need, step by step. You are on Windows, so this is a small PowerShell script, nothing to install.

### 1. Find the Wi-Fi MAC address of your UE2

On the Ultimate: press **F2**, go to **-- Network --**, then **WiFi settings**. At the bottom of that page there is a line **Interface MAC**, shown as something like `aa:bb:cc:dd:ee:ff`. Write it down.

Please take the one from the WiFi page. The MAC on the Ethernet page is a different interface and cannot wake the machine.

### 2. Prepare the machine

1. Connect the Ultimate to your **Wi-Fi**, not to the network cable. The wired jack cannot do this: it is switched off together with the machine.
2. Press **F2**, go to **Power Settings**, set **Wake On Wi-Fi** to **Enabled**, and save the settings.
3. Switch the machine off with its power button. Leave the mains cable connected.

### 3. Save the script

Open Notepad, paste the text below, and save it on your Desktop as `wake-ultimate.ps1`. In the save dialog, set "Save as type" to **All files**, otherwise Notepad adds `.txt` to the name.

```powershell
# wake-ultimate.ps1 - switches on a C64 Ultimate / Ultimate 64 Elite II that is
# off but still connected to mains, by sending a Wake On LAN "magic packet"
# to its Wi-Fi MAC address.
#
# Usage:  powershell -ExecutionPolicy Bypass -File .\wake-ultimate.ps1 aa:bb:cc:dd:ee:ff

param(
    [Parameter(Mandatory = $true, Position = 0)]
    [string]$Mac,
    [int]$Port = 9
)

$parts = $Mac -split '[:\-]'
if ($parts.Count -ne 6) {
    Write-Host "The MAC address must have six parts, for example aa:bb:cc:dd:ee:ff" -ForegroundColor Red
    exit 1
}

$addr = New-Object byte[] 6
for ($i = 0; $i -lt 6; $i++) {
    try { $addr[$i] = [Convert]::ToByte($parts[$i], 16) }
    catch {
        Write-Host "'$($parts[$i])' is not a hex number. Example: aa:bb:cc:dd:ee:ff" -ForegroundColor Red
        exit 1
    }
}

# A magic packet is 6 bytes of 0xFF followed by the MAC address 16 times.
$packet = New-Object byte[] 102
for ($i = 0; $i -lt 6;  $i++) { $packet[$i] = 0xFF }
for ($i = 0; $i -lt 16; $i++) { [Array]::Copy($addr, 0, $packet, 6 + $i * 6, 6) }

# Send to the general broadcast address and to the broadcast address of every
# network this PC is on, so it does not matter which adapter is the right one.
$targets = New-Object System.Collections.ArrayList
[void]$targets.Add("255.255.255.255")

foreach ($nic in [System.Net.NetworkInformation.NetworkInterface]::GetAllNetworkInterfaces()) {
    if ($nic.OperationalStatus -ne 'Up') { continue }
    if ($nic.NetworkInterfaceType -eq 'Loopback') { continue }
    foreach ($ua in $nic.GetIPProperties().UnicastAddresses) {
        if ($ua.Address.AddressFamily -ne 'InterNetwork') { continue }
        if ($ua.IPv4Mask -eq $null) { continue }
        $ip   = $ua.Address.GetAddressBytes()
        $mask = $ua.IPv4Mask.GetAddressBytes()
        if ($ip[0] -eq 127 -or ($ip[0] -eq 169 -and $ip[1] -eq 254)) { continue }
        $bc = New-Object byte[] 4
        for ($i = 0; $i -lt 4; $i++) { $bc[$i] = $ip[$i] -bor (0xFF -bxor $mask[$i]) }
        $text = (New-Object System.Net.IPAddress(,$bc)).ToString()
        if (-not $targets.Contains($text)) { [void]$targets.Add($text) }
    }
}

$udp = New-Object System.Net.Sockets.UdpClient
$udp.EnableBroadcast = $true
$sent = 0
try {
    foreach ($t in $targets) {
        $endpoint = New-Object System.Net.IPEndPoint([System.Net.IPAddress]::Parse($t), $Port)
        for ($i = 0; $i -lt 3; $i++) {
            try { [void]$udp.Send($packet, $packet.Length, $endpoint); $sent++ }
            catch { }
        }
        Write-Host "Sent to $t port $Port"
    }
} finally {
    $udp.Close()
}

if ($sent -eq 0) {
    Write-Host "Nothing could be sent. Is this PC connected to the network?" -ForegroundColor Red
    exit 1
}

Write-Host ""
Write-Host "$sent magic packets sent for $Mac." -ForegroundColor Green
Write-Host "The machine should switch on within a few seconds."
```

### 4. Run it

Open the Start menu, type `PowerShell`, and start **Windows PowerShell**. Then type this line, with your own MAC address from step 1, and press Enter:

```
powershell -ExecutionPolicy Bypass -File "$env:USERPROFILE\Desktop\wake-ultimate.ps1" aa:bb:cc:dd:ee:ff
```

The script prints which addresses it sent to. Your PC has to be on the same network as the Ultimate, because a magic packet is a broadcast and does not travel across a router.

### 5. What to report

1. **Wake On Wi-Fi = Enabled**, machine off, mains connected, script run: does the machine come up, and roughly how long does it take?
2. Switch the machine on by hand, set **Wake On Wi-Fi = Disabled**, save, switch it off again, run the script once more: it should now stay off. Then switch it on with the power button as usual.

That pair is the whole test. Anything you see that looks wrong is worth reporting too.

### If nothing happens

- Check that the machine is really on Wi-Fi and not on the network cable, and that the MAC is the one from the WiFi page.
- Check that your PC is on the same network, for example both on the same router.
- If you would rather not use a script at all, the freeware tool [WakeMeOnLan](https://www.nirsoft.net/utils/wake_on_lan.html) does the same thing: enter the MAC, press Wake Up.

The script is not guesswork: I captured the packet it sends and ran it through the matching code from this pull request, which accepted it for that MAC and ignored it for a different one. So if the machine stays dark, the script is not the reason.

Thanks again for helping out.

Retro1968:
Thanks for your great instructions. Here's my report:

It worked on the first try! :)

Retro1968:
> @Retro1968 Sorry for the last minute announcement, but would you mind using https://github.com/GideonZ/1541ultimate/actions/runs/33243805039 instead once built?
> 
> I only removed needless line ending changes in one file. No functional change, just cleanup.
> 
> If you have already flashed the firmware from the earlier link by the time you read this, then don't worry.
> 
> Also, my next comment has a more detailed walk-through of the test steps, including a command line script for Windows.

Oh i read this after my first try. Wil do the test with the new firmware right now and will report.

Retro1968:
Ok as you wrote above:

_"@Retro1968 Sorry for the last minute announcement, but would you mind using https://github.com/GideonZ/1541ultimate/actions/runs/33243805039 instead once built?
I only removed needless line ending changes in one file. No functional change, just cleanup.
If you have already flashed the firmware from the earlier link by the time you read this, then don't worry."_


I will not do another test. Because the first test with you first link, worked. Do i understand you right?

chrisgleissner:
Thank you *so much* @Retro1968 ! That is great news!

Also well done @enver-haase and thanks for your hard work on this one, and continued engagement.

Have a great weekend!

Retro1968:
I tested firmware "update_v3.15-100-g5ee21b65.ue2" on my UE2 and found a menuerror. You can see on this screenshot that i attached a jiffydosfile for the 1581. After that a "mod" is set in front of the filename. I guess this is an old menuerror. You can find it in menu "Memory Configuration" and the drive settings. It only affects the 1581 menu field.

menuerror

Issue #803, opened 2026-08-25
[BUG] FTP uploads to USB storage corrupt data; the machine's own writes of the same size are clean (C64U, 1.1.0)


enver-haase:
**Environment:** C64 Ultimate, firmware from the 1.1.0 branch (a custom build on top of it), FPGA 122, core 1.49, PAL, USB stick as `/USB0`. In the tree used here, the FPGA bitstreams under `external/`, `software/network/ftpd.cc`, all of `software/filesystem/`, and `software/io/usb/usb_base.*` / `usb_scsi.*` are byte-identical to tag `1.1.0`; the only local USB differences are in `keyboard_usb.*` / `usb_hid.cc`, i.e. the HID path, not mass storage.

**Summary:** Files uploaded to `/USB0` by FTP come back with a handful of wrong bytes, silently and with the correct file size. Files of the same size written by the machine itself to the same stick are clean, so this is not "large USB writes are broken" — it is specific to the FTP receive path, which writes in 1024-byte chunks, i.e. **2-sector** `disk_write()` calls. Two distinct failure modes appear, one of them a race and one deterministic.

**Repro** — a self-describing payload makes the damage readable: each 32-bit little-endian word holds its own index, so a wrong word states exactly what it became.

```python
n = 4603968
pat = b''.join(i.to_bytes(4, 'little') for i in range(n // 4))
# STOR to /USB0/<dir>/big.bin, then RETR and compare word by word
```

Three uploads of that payload gave 19, 25 and 16 wrong words out of 1,150,992.

**Mode A — the tail halfword of a sector, overwritten with a per-transfer constant.** 34 words across two runs. Always at byte offset ≡ **508 mod 512**, the last 32-bit word of a sector. The lower halfword is always intact; the upper halfword is replaced by a single value that stays constant for the whole transfer — `0x2FA8` in one run, `0xC440` in the next, `0xD478`–`0xD47A` in a third:

```
word     3967: expected 0x00000F7F, got 0xD4780F7F
word    90367: expected 0x000160BF, got 0xD47960BF
```

Within a burst the hits are spaced exactly 1024 bytes. Counts and positions move between runs — 25 and 16 wrong words with 11 offsets in common — so this one is timing-dependent.

**Mode B — a word at a 256 KiB boundary takes the value of the neighbouring block.** 7 words across two runs, and this one looks deterministic:

```
run 2  0x007FFFC   value belongs to +0x40000   boundary 0x80000
[...]

PR #816, opened 2026-08-29
E2E: catch the FTP upload that corrupts USB storage

enver-haase:
Covers #803, which has no test today. A file uploaded by FTP onto USB storage comes back with
individual 32-bit words altered, at exactly the right size and with no error reported by
either side.

The reason the two of us could not agree on reproducing it is that the trigger is the size of
the writes the *client* makes, not the link. 1350 bytes at a time leaves
`recv(socket, buffer, 1024)` returning 1024 and then the 326-byte remainder, which puts the
file pointer out of sector alignment; `f_write` then hands `disk_write` a pointer part-way
into the FTP buffer, and it reaches the USB DMA unchecked. 700 bytes at a time never leaves a
whole sector over, so that pointer is never formed. Both shapes are pure client behaviour, so
this needs no traffic shaping and no impaired link.

**Three checks, in this order for a reason.** The control writes 700 B at a time and must come
back intact; a failure there is not this defect and says the suite itself is unsound. The
discriminator repeats the regression shape against `/Temp`, which shares the daemon and the
buffer but has no USB write path -- it runs *before* the regression on purpose, because
`check()` re-raises and an answer collected afterwards would never reach the report. The
regression then writes 1350 B at a time.

**Each check writes a newly created file.** Measured, not assumed: overwriting a file of the
same size reuses the clusters it already owns and the defect does not appear. An earlier draft
shared one name between the checks and reported a false pass.

The transfer gets its own timeout, separate from the one governing command replies: the
closing reply arrives only once the device has written everything.

`manual=True` -- it needs real USB media in the machine and writes there.

**Verified against hardware.** A C64 Ultimate running a custom build on top of the Commodore
1.1.0 branch, on Wi-Fi. The control and the `/Temp` discriminator pass; the regression fails
with the signature from #803:

```
[03] a client writing 1350 B at a time stores the file intact ... FAIL
     1 word(s) differ; 1 at offset 508 of a 1024-byte pair, 1 with the lower halfword intact
     0x003DFC expected A5000F7F got D1100F7F
```

Also run from a second machine on the device's own subnet, where it reproduces likewise, and
`runner-policy` stays at 71/71 with the new entry registered.

No fix here -- this is the red half. Where the fix belongs is still open on #803; my
preference is a bounce buffer in `UsbBase::bulk_out` mirroring the one already live in
`bulk_in`, because that protects every caller rather than the one we happened to find.

**Disclosure:** written with Claude Code, Opus 5 at high reasoning effort, directed and
reviewed by me.

### DIFF
--- a/run-tests
+++ b/run-tests
@@ -286,6 +286,12 @@ SUITES: Sequence[Suite] = (
     # name it accepts back for SIZE and DELE.
     Suite("e2e", "ftp-server", "tests/e2e/network/ftp_server_test.py",
           "-H @HOST@ -p @PASS@ -t @TIMEOUT@"),
+    # Writes a 64 KiB file to real USB media, so it needs a stick in the machine
+    # and leaves nothing behind. The regression check fails on firmware carrying
+    # #803; the control and the RAM-disk comparison beside it say whether a
+    # failure is that defect or something else.
+    Suite("e2e", "ftp-usb-integrity", "tests/e2e/network/ftp_usb_integrity_test.py",
+          "-H @HOST@ -p @PASS@ -t @TIMEOUT@", manual=True),
     # Needs passwordless "sudo -n ip" plus an --iface/--victim-ip for this LAN,
     # and waits ~75s for the device to reap the sessions.
     Suite("e2e", "telnet-stale-session", "tests/e2e/network/telnet_stale_session_test.py",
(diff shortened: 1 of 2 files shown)


chrisgleissner:
Fantastic work @enver-haase !  Looking forward to your fix.

PR #821, opened 2026-08-31
Bounce a bulk_out transfer whose last word is only partly wanted (#803)

### LINKED ISSUE #803: [BUG] FTP uploads to USB storage corrupt data; the machine's own writes of the same size are clean (C64U, 1.1.0)

enver-haase:
`bulk_out()` hands the USB engine the caller's pointer unchanged. `do_mem_read` in `software/io/usb/nano_minimal.nan` supports an unaligned source by keeping the start address' byte offset and rounding the read up to whole words:

```
    LOADI Descriptor,Offset_Command_MemLo
    OUTP  MEM_ADDR_LO
    AND   #3
    STORE Current_Word_Offset
    ...
    STORE CurrentLength
    ADD   Current_Word_Offset
    ADD   #3
    OUTP  MEM_TRANSFER_READ
```

So a transfer whose final word is only partly wanted makes the engine fetch a word the caller never asked for, and that word reaches the data: an FTP upload to a USB stick comes back with one 32-bit word per occurrence keeping its lower halfword and losing its upper one. The file has the right size, so nothing notices.

### Where it comes from

`FTPDataConnection::receivefile()` does `recv(sock, buffer, 1024, 0)` into `char buffer[1024]` and hands what it got to `f_write`. FatFs writes whole sectors straight out of the caller's buffer when the file pointer sits on a sector boundary, so the DMA source is `buffer + k` for whatever `k` has been consumed so far. A client sending 1350 bytes at a time makes `recv` return 1024 and then the remainder, and `k` walks all over the buffer.

Exactly one value of `k` breaks: 510, in a 512-byte transfer, whose word-rounded read ends exactly at `buffer + 1024`. `k = 506` is equally misaligned and also ends on a partial word, but its read stops four bytes short, and it is clean.

### Evidence

Instrumented firmware — a custom build on top of the Commodore 1.1.0 branch — recorded every `recv` size and every source address the block layer was handed. Over nine natural traces: 516 direct writes out of the receive buffer, 273 of them misaligned, three from `k = 510`. Exactly those three corrupted a word; the other 270 misaligned transfers were fine.

The offending offset can be predicted from the client's send pattern alone, so it can also be forced: send 1024 bytes, then 2, then 1024s. The file pointer lands on 2 mod 512 and stays there, which turns roughly one opportunity per upload into 63.

| firmware | client | runs | `k = 510` per run, from the device trace | corrupted words per run |
|---|---|---|---|---|
| unpatched | same LAN segment | 5 | 63 | 63 |
| unpatched | across a VPN | 5 | 63 | 63 |
| **this patch** | same LAN segment | 5 | 63 | **0** |
| **this patch** | across a VPN | 5 | 63 | **0** |

630 corrupted words in ten unpatched runs, none in ten patched runs, with the trigger present and counted in all twenty. Hardware: C64 Ultimate, board revision 0x17, FAT32 stick with 32 KiB clusters. A 4 MiB upload and read-back on the patched build matches sha256.

### Why the copy is word-aligned, and why it is over-allocated

A variant that bounces to an address just as misaligned as the original, with the same spare word behind it, changing only where the data sits, is **worse than no patch at all**: 126 corrupted words per run instead of 63, in ten runs out of ten, adding the first word of every transfer to the last. So the alignment is what does the work here, and the cure is not a timing artefact of the copy — that variant copies exactly as much.

The `+ 4` is not redundant after aligning. Without it the rounded read would end exactly at the end of the allocation, which is the condition that goes wrong in the first place.

The test is a property of the transfer rather than of the pointer, so it also covers `usb_ax88772.cc` and `usb_em1010.cc`, which pass `pkt_len + 4` and `pkt_len + 2` from an aligned start.

### What this does not do

- It does not fix the root cause, which is in the nano's read path. This is a driver-side workaround; if the handling of a partial final word is corrected there, this can go.
- `bulk_in()` bounces only when the start is misaligned, and allocates exactly `len` with no slack. If the write path rounds up the same way, an aligned start with a length that is not a multiple of four could write past the end of the caller's buffer. Not tested — only noticed while reading.
- The cost is one allocation and one `memcpy` per affected transfer. I have not measured throughput against an unpatched build on the same workload.
- An Ultimate II+L on 3.14d, with the same `char buffer[1024]` and the same `recv`, did not reproduce in 20 runs. There was no instrumentation on that machine, so that is an absence rather than a negative result.

Fixes #803.

### DIFF
--- a/software/io/usb/usb_base.cc
+++ b/software/io/usb/usb_base.cc
@@ -758,15 +758,30 @@ int  UsbBase :: bulk_out(struct t_pipe *pipe, void *buf, int len, int timeout)
 	uint8_t sub = PROFILER_SUB; PROFILER_SUB = 13;
     volatile t_usb_descriptor *descr = USB2_DESCRIPTOR(0);
 
-/*
-    if (((uint32_t)buf) & 3) {
-    	printf("Bulk_out: Unaligned buffer %p (%d - len = %d)\n", buf, ((uint32_t)buf) & 3, len);
-    	dump_hex((void *)(((uint32_t)buf) & -4), 16);
-    	//while(1);
-    }
-*/
+	/* Issue #803. do_mem_read in the nano rounds a read up to whole words: it
+	   keeps the start address' byte offset and adds that, plus 3, to the
+	   length. A transfer whose last word is only partly wanted therefore makes
+	   the engine fetch a word the caller never asked for, and that word has
+	   been seen to reach the data written to a USB stick.
 
-    uint32_t addr = ((uint32_t)buf);
+	   Copy such a transfer to a buffer of our own. Two properties matter, and
+	   both were established on hardware:
+	     - the copy starts on a word boundary. Bouncing to an address that is
+	       just as misaligned as the original, with the same slack behind it,
+	       makes the corruption twice as bad instead of curing it.
+	     - the allocation has a spare word behind the rounded length, so the
+	       rounded read does not end exactly at the end of the block we own.
+	   get_mem() is malloc(), so the start is aligned for any fundamental type
+	   and no manual alignment is needed here. */
+	uint8_t *origbuf = (uint8_t *)buf;
+	uint8_t *newbuf = origbuf;
+
+	if (((((uint32_t)origbuf) & 3) + len) & 3) {
+		newbuf = new uint8_t[((len + 3) & ~3) + 4]; // get_mem panics, never returns 0
+		memcpy(newbuf, origbuf, len);
+	}
+
+    uint32_t addr = ((uint32_t)newbuf);
 	int total_trans = 0;
 
 	uint16_t result;
@@ -815,6 +830,10 @@ int  UsbBase :: bulk_out(struct t_pipe *pipe, void *buf, int len, int timeout)
 
     // printf("Bulk out done: %d\n", total_trans);
     xSemaphoreGive(mutex);
+
+	if (newbuf != origbuf) {
+		delete[] newbuf;
+	}
 
     PROFILER_SUB = sub;
 	return total_trans;


enver-haase:
**In one paragraph:** this PR is a **workaround**, not a fix. Two other places look like the
same defect and are **neither tested nor fixed** — `UsbBase::bulk_in()` in this file, and the
`last` state of `fpga/altera/align_read_to_bram.vhd`. The **real fix belongs in that same
file**, in the offset ≠ 0 assembly. We have a **GHDL testbench** for it, offered below; it
needs one correction before its output means anything.

Throughout, `k` is the byte offset into the 1024-byte FTP receive buffer that the DMA is
handed as its source — `buffer + k`. FatFs writes whole sectors straight out of that buffer,
so `k` is whatever the client's send pattern has left consumed.

### Why the workaround sits in the driver

`do_mem_read` in the nano already asks for the minimum: for `k = 510` and 512 bytes it sends
`512 + 2 + 3 = 517`, decoded as 129 words, exactly `ceil((2 + 512) / 4)`. The nano is a
sequencer, not a datapath — the bytes never pass through it, and `CMD_DATA_BUFFER_CTRL`
carries a buffer index and a byte count with no offset field, so it cannot read aligned and
then start the packet two bytes in. `usb_base.cc` is byte-identical at 1.1.0, v3.14d, v3.15
and `test-merge`, so the same C runs everywhere and its only lever is which address it hands
over. Not handing over a misaligned one is the only remedy above the HDL.

### Why the defect is in the read path

Two cases where the software is bit-for-bit identical and only the address differs, yet the
outcome differs. Nothing in `ff.c`, `ftpd.cc` or `usb_base.cc` branches on that, so whatever
decides it is downstream of the request.

- `k = 506` and `k = 510`: equally misaligned, same length, same code path. 506 clean 7 times
  out of 7; 510 corrupted 3 out of 3, and 63 out of 63 once forced.
- A build that bounces to an address *just as misaligned*, with owned slack behind it, is
  **worse than no patch**: 126 damaged words per run instead of 63, in 10 runs out of 10.

That is proof of location, not of mechanism.

### What I think is happening — and why I do not trust it far

I think the steady state of the aligner is fine and the trouble is at the **flush**, the
emission of the final output word. The damaged word is always the transfer's last one. The
keep-the-skew build added damage at the *first* word as well, which is the other end of the
same fill/flush pipeline. And `align_read_to_bram` has a dedicated `last` state for the
partial-final-word case which, by its own guard, **never runs** when the length is a multiple
of four — which every FatFs write is. The case the code has a state for is being handled by
the streaming path instead.

Against that: `k = 506` and `k = 510` present *identical* inputs to that state machine. Same
offset (both `& 3 == 2`), same length, same 517, same 129 beats, same 128 emissions. Their
start addresses differ by one word. One never fails, the other always does. No control logic
in the aligner can tell them apart.

So either the memory system behaves differently for two addresses four bytes apart, or the
rule we fitted — "misaligned start **and** the word-rounded read ends at the end of the
buffer" — is three data points dressed up as a law. I lean to the second, because that rule
refers to a boundary the hardware cannot see: `buffer + 1024` was at `0x812DB4` in August and
`0x816C04` on 31 August, neither a page, bank, burst nor cache boundary, and the read stops
at `BUF + 1023` and never reaches it anyway.

One property is solid whatever the mechanism: **it is deterministic.** Forced, it is 63 out
of 63, twice, from two different clients and at two different buffer addresses. That is not a
race that occasionally loses, which rules out refresh collisions and arbitration luck.

### What we will do next, and what each outcome would mean

Now that the offset can be forced, it can also be *swept*: the nudge size chooses where the
file pointer lands, so most values of `k` can be driven on purpose and tried ten times each.
That would replace three data points with the complete failure set. It needs no HDL and no
gateware, just the machine here for an hour.

- only `k = 510` fails → the fitted rule survives a real test and is worth defending;
- a family fails, with a period in `k` → that period names the mechanism, most likely a beat
  count or a FIFO depth;
- failures track the absolute address rather than `k` → it is the memory controller, and the
  aligner is off the hook.

**Is that worth having?** If yes I will run it tomorrow and post the table here. I would
rather not guess at which of those three worlds we are in for another round.

### What would settle it on your side

**First, an RTL experiment that needs nothing but the published sources.** Simulate
`align_read_to_bram` with the beats *not* back-to-back: delay `rdata_valid` before the final
beat and see whether the last emitted word takes a stale `remain`/`rdata` pair. If it does,
the flush is fragile to gaps in the beat stream, and that is fixable in the VHDL whatever the
controller does. This tests the idea above without depending on our testbench being right,
because it varies the cadence rather than the data.

**Second, if that comes back clean, an ILA on one `k = 510` transfer** capturing
`rdata_valid`, `mem_resp.dack_tag`, `wdata`, `wmask` and the state register, plus a timing
report on the USB memory path. That shows directly whether the final beat arrives late and
what the aligner emitted when it did. It is also the only way to observe the DDR controller
at all from outside: neither the Artix top level nor its controller is in this repository,
only `external/u64_mk2_artix.bit`.

### The testbench, yours if you want it

`tb_usb_mem.vhd` drives `usb_memory_ctrl` exactly as the nano does — buffer index, address
low, address high, then `len + offset + 3` — with memory modelled so that every halfword
names its own offset into the receive buffer and everything past the buffer reads back as the
value the device actually substituted. It runs under GHDL against a v3.14d or v3.15 checkout,
`ghdl -r --std=08 tb_usb_mem -gg_k=510`, and takes seconds.

Its one known flaw, so nobody wastes an afternoon on it: the memory model returns `3210`
regardless of offset. That is the wrong convention. `usb_memory_ctrl` passes the byte offset
down (`mem_req.address <= mem_addr_i & mem_addr_r(1 downto 0)`), the rotation only assembles
correctly for a controller that rotates, and on hardware 270 misaligned transfers came back
clean, which the non-rotating convention could not produce. The model has to rotate first.
Say the word and I will send it over, corrected or as it stands.

### Evidence that the workaround works

In the PR description. Short form: with `k = 510` forced to 63 occurrences per upload, ten
unpatched runs lost 63 words each and ten patched runs lost none, the trigger counted in the
device's own trace in all twenty.

### Neither tested nor fixed

- `bulk_in()` bounces only when the **start** is misaligned and allocates exactly `len` with
  no slack, so an aligned start with a length that is not a multiple of four could write past
  the caller's buffer if the write path rounds up the same way.
- In `align_read_to_bram`, the `last` state's byte enables are commented out in favour of an
  unconditional `wmask <= X"F"`. That cannot be this bug, since the state never runs for a
  multiple-of-four length — but it looks like a real defect for lengths that are not, which is
  what `usb_ax88772.cc` and `usb_em1010.cc` pass.
- No core has changed here: `align_read_to_bram.vhd` is byte-identical at v3.14a, v3.14d,
  v3.14e, v3.15 and `test-merge`, and the only `usb_memory_ctrl.vhd` change anywhere is a tag
  re-encoding on `test-merge` that leaves the alignment wiring untouched. So a newer core is
  not a way out of this one.

chrisgleissner:
Reviewed the alignment math and the surrounding call sites. The fix itself is sound; two things adjacent to it are worth tracking.

**Alignment math checks out for chunked transfers.** `bulk_out` splits any `len > 49152` into multiple descriptor commands, and I initially wondered whether the misalignment check (done once, on the original `buf`/`len`) still holds for the later chunks. It does: 49152 is a multiple of 4, and after the bounce the buffer starts at offset 0, so every full-size chunk keeps `addr` word-aligned going into the next chunk. Only the last chunk (`len mod 4`, or the whole thing if `len <= 49152`) can end up misaligned, and that's exactly what `(((origbuf & 3) + len) & 3)` tests against a word-aligned base. So the single up-front check is equivalent to checking each chunk.

**`bulk_in()` has the same defect, and it's a write-direction overflow rather than corrupted output data.** The PR already calls this out under "What this does not do," but I want to underline it: `bulk_in()` only bounces when `(uint32_t)buf & 3`, and allocates exactly `len` bytes with no spare word. If the nano's rounding applies symmetrically to the DMA-write path, any aligned buffer with a `len` that isn't a multiple of 4 will make the hardware write one extra word past the buffer. I checked the current callers (`stat_resp[32]`/`sense_data[32]` in `usb_scsi.cc`, `irq_data[16]` in `usb_ms_cbi.cc`, all read for 13/18/8 bytes respectively) and they all happen to have enough slack in the declared array that a 1-3 byte overrun stays inside the buffer today. So I don't have a live overflow to point at, but that's incidental to those specific declarations, not a property `bulk_in()` enforces. A future or existing caller with an exactly-sized aligned buffer and a non-multiple-of-4 length would get silent heap/stack corruption instead of the word-swap this PR fixes. Worth its own tracking issue given it's a memory-safety bug rather than a data-corruption one.

**Allocation failure here is a hard hang, not a recoverable error.** `new` in this codebase routes through `get_mem()`, which on failure prints `** PANIC **` and spins in `while(1);` — confirmed in `software/system/memory_rtos.cc` / `memory.cc`. The author's own trace shows roughly half of `bulk_out` transfers are misaligned (273/516), so this bounce path is not a rare corner case — it's a `new[]`/`delete[]` cycle on a large fraction of calls, including every SCSI CBW (31 bytes, essentially always misaligned) and any multi-sector mass-storage write. On a small embedded heap, sustained variable-sized alloc/free traffic is exactly the pattern that fragments over time. The PR already flags throughput as unmeasured; I'd add that heap fragmentation leading to that panic path is a reliability question worth at least some soak testing, since the failure mode is a full device hang rather than a dropped transfer.

**Minor:** the bounce condition is `((origbuf & 3) + len) & 3`, which doesn't gate on `len > 0`. A zero-length call with a misaligned `buf` will still allocate and copy nothing meaningful. Harmless in practice, just an avoidable allocation.

None of this blocks the fix for #803 itself — the core change is correct and the hardware evidence is solid. Flagging the `bulk_in()` sibling and the panic-on-exhaustion path as follow-ups.

chrisgleissner:
Thanks for the deep dig into the nano/aligner behavior, @enver-haase. That's a genuinely hard bug to pin down, and the hardware evidence you've gathered so far is great work. 

We're close to cutting 3.15, so I'd like to get this one properly locked in rather than merged in a hurry. 

Here's what I think still needs to happen before we merge:

1. **A red/green test for #803.** Something that fails on current `test-merge` and passes with this patch, so the fix is provable on its own rather than resting on manual trace captures.
2. **A full USB e2e rerun** against this branch, to confirm nothing else in the mass-storage/network USB paths regresses.
3. **The `k` sweep you proposed** — running the full set of offsets rather than just the three data points we have now. That would tell us whether we're fixing the one case we've seen or the whole family, and it directly informs whether the workaround's model is complete.
4. A decision (doesn't have to be in this PR) on the two adjacent items flagged in review: `bulk_in()` has the same rounding exposure on the write side, and allocation failure in the bounce path currently hangs the device rather than failing gracefully.

None of this is a knock on the fix. It looks right, and the instrumentation behind it is solid. 

Happy to help with the e2e run or the test scaffolding if needed.

chrisgleissner:
FYI @enver-haase @GideonZ I did a deep-dive sweep of the repo to see if similar issues exist elsewhere. You can see the findings in https://github.com/GideonZ/1541ultimate/issues/822

Questions / comments: 
- @GideonZ Several of the issues found in issue `822` are related to this ticket. Thus, to avoid merge conflicts which could arise for tests or test fixtures (possibly) used both here and in `822`, I suggest we tackle `822` only after this PR here will have been merged.
- @enver-haase If you are interested in fixing `822` , please let us know. You mentioned last week that you will be very busy due to work projects going forward, so I am also happy to look at `822`. But as you discovered this set of errors, I don't want to take away the joy that comes with fixing them. :-) Just let us know what you prefer.

GideonZ:
I'd rather have this fixed in the VHDL than by a band-aid in the software, although I can totally live with a temporary patch.

If I recall correctly, the memory rotation may or may not be part of the DDR2 controller; there are different versions of "align_read_to_bram", and I'll first have to figure out which one is active in which build before we continue. 

The best is to run some simulations; and apparently I you have been able to set up a simulation harness for the case k = 510. Is this VHDL harness available somewhere? If not, I can have a look to see what happens when I build a small harness manually to see what really happens. Maybe the rotation setting gets reset too early? Oh well, let's not speculate. Simulation with a proper wave viewer is the way to go.

enver-haase:
@GideonZ Agreed on the band-aid — if the aligner turns out to be it, the bounce should come back out. The harness is public now, everything in one place:

**https://github.com/enver-haase/ultimate-803-harness**

## Before you open `sim/`: the testbench's memory model is wrong

Saying this first so nobody loses an afternoon to it. `tb_usb_mem.vhd` runs, but **its output currently proves nothing**, and the reason is the one open question in the whole investigation.

`align_read_to_bram.vhd` documents two incompatible conventions for what the DRAM returns:

```
-- we always get 3210, regardless of the offset.
-- In case the offset was sent to the DRAM, we get: ... (x321 / xx32 / xxx3)
```

My testbench implements the **first**. Three things say this machine uses the **second**:

- The byte offset *is* handed to the controller: `usb_memory_ctrl` drives `mem_req.address <= mem_addr_i & mem_addr_r(1 downto 0)`, so the low two bits travel with the request.
- Both conventions exist in the tree, on different paths. `fpga/altera/ddr2_ctrl.vhd` does not rotate — its `rdata <= phy_rdata(34 downto 27) & ...` is a fixed de-interleave with no address dependence — while `fpga/ip/busses/vhdl_source/mem32_to_avalon_bridge.vhd` rotates by `memreq_address(1 downto 0)` and documents it in exactly the words the second comment block uses ("When address is 1 ... so word is rotated: 0321").
- On hardware, 270 misaligned transfers came back clean. Under the non-rotating convention `align_read_to_bram` would assemble *every* misaligned transfer wrongly, not three of them.

So the rotation the module performs (`rdata(31 downto 16) & remain(15 downto 0)` for offset 2) only assembles correctly for a rotating controller, and my model is modelling a different machine. Rotating the model's return data by the offset and re-running `-gg_k=510` is the first thing to change; I did not get to it. `sim/README.md` has the long form.

And this is not just a testbench bug — it is the question itself. Which convention the mk2 actually uses cannot be read off the sources: `fpga/fpga_top/` has only `cyclone4_test`, `ecp5_test`, `ultimate_fpga` and `video_fpga`, and for the Artix there is only the built `external/u64_mk2_artix.bit`. If your simulation settles which controller feeds `align_read_to_bram` on that board, that alone may decide #803.

## What the testbench does, once the model is right

It drives `usb_memory_ctrl` exactly as the nano does for a MEMREAD — buffer index, address low, address high, then `len + offset + 3` — with every halfword naming its own offset into the 1024-byte FTP receive buffer and everything past the buffer reading back as the value the device actually substituted. A sentinel appearing in the BRAM means data from beyond the buffer reached the transfer. It runs under GHDL against a v3.14d or v3.15 checkout and takes seconds.

Two things from the HDL that hold regardless of the simulation, in case they save you a detour:

- `len + offset + 3` is decoded as a word count in bits 9..2 (`mem_remain_counter` loads `cmd_wdata(9 downto 2)`) with `last_bytes` in bits 1..0. The round-up reading is confirmed from the hardware side.
- For any length that is a multiple of 4 — which every FatFs write is — `last_bytes = (offset+3) mod 4`, so the guard `offset > last_bytes` in the `stream` state is always true and the `last` state never runs. The commented-out byte-enable logic there therefore cannot be #803. It still looks like a real defect for lengths that are not a multiple of 4, which is what `usb_ax88772.cc` and `usb_em1010.cc` pass.

## Driving a real machine

The Python half needs nothing but Python 3, the FTP File Service enabled and a writable stick — no firmware sources, no instrumentation. The forcing is the useful part: `k = 510` arises when the file pointer sits at 2 mod 512 and a `recv` returns at least 1022 bytes, and after such a write the pointer lands back on 2 mod 512, so the condition sustains itself. Sending 1024, then 2, then 1024s turns one opportunity per upload into 63. On the unfixed image that is 63 corrupted words per run, ten runs out of ten, from two different clients; on the patched image, zero — with `k = 510` counted in the device's own trace in all twenty. So any offset you want tried can be driven on demand rather than waited for.

`instrumentation/` has the recording firmware: about sixty lines in `diskio.cc/h` and `ftpd.cc` that write `/Flash/ftptrace.txt` — every `recv` size in order, and the source pointer, LBA and sector count of every `disk_write`. Given as source rather than a patch, because `ftpd.cc` on the 1.1.0 line differs too much from 3.x for one to apply; the hooks are three calls. `results/` has the nine traces from 29 August and the forced runs from 31 August, and `results/analyse.py` reduces a trace to the buffer offsets it used.

## On which build carries which aligner

From the sources, nothing has moved. `align_read_to_bram.vhd` is byte-identical — blob `a9a79184` — at v3.14a, v3.14c, v3.14d, v3.14e, v3.15, `master` and `test-merge`. `usb_memory_ctrl.vhd` is one blob (`7ce236ea`) across all of those and differs only on `test-merge` (`16ab9769`), and that diff is the memory-bus tag re-encoding: the alignment wiring (`offset => mem_addr_r(1 downto 0)`, `last_bytes => size_r`) is untouched. On the software side `usb_base.cc` as a file has changed between releases, but `UsbBase::bulk_out` itself is character-for-character identical from 1.1.0 through every one of those refs. So the difference between builds, if there is one, is on the controller side the repository does not publish.

Happy to run any sweep on the machine here if that is faster than simulating it.

enver-haase:
@chrisgleissner Thanks — and yes to the sweep, with one condition on the order.

Gideon would rather fix this in the VHDL than keep the bounce, and I agree. So the first call is his: if he gets the gateware fix done in the time left before 3.15, the sweep is measuring a workaround that is on its way out. If he does not make it in that window, we run the full `k` sweep here and post the table, and the bounce ships as the temporary patch he said he can live with.

The forcing rig for it is public now, together with the traces and the GHDL testbench: https://github.com/enver-haase/ultimate-803-harness — `forced_k510.py` drives any offset on demand, so the sweep is an hour on the machine rather than a wait for luck.

On #822, correcting what I first wrote here: I had put it back to you, and that was the wrong way round, because you do not have the machine these findings need. Most of them cannot be settled on the code alone — the bounce path, the `bulk_in()` write side, anything that reaches the USB DMA — so let me take the hardware side rather than hand you a sweep you cannot verify. The C64U, both clients and the forcing rig are set up here and the turnaround is minutes: write whatever you want on the code side and I will run it, or hand me the list and I will work through it.

GideonZ:
I will have to try it and maybe build with a chipscope instance in the fpga. I did my own sim this afternoon cycling over 512 byte transfers with k ranging from 508 to 511. All clean.

GideonZ:
The Xilinx part uses the same aligner as the Altera part, the lower address bits are passed to the DRAM to rotate the burst. But there is a 32-to-64 bit memory block in between, since the memory controller has a 64 bit interface.

chrisgleissner:
@enver-haase I have been unable to reproduce this issue on my Ultimate 64 Elite I with the latest test-merge despite best efforts. Is there any e2e test you can share that forces the issue in your machine? Would be nice to add it to this PR, following the e2e conventions in the code base.

chrisgleissner:
FYI @enver-haase I was able to reproduce the issue on my unpatched C64U in a new E2E that fails red reliably in less than 2s.

It is called `usb-bulk-out-integrity`. For more details, see https://github.com/GideonZ/1541ultimate/pull/824

The same test against both U2+L and U64 Elite I, both on the latest `test-merge` firmware, passes reliably.

We can use this test to verify that the issue has been fixed and won't reoccur.

GideonZ:
Fixed in the FPGA.

The problem was that on a C64U / U64E2, the memory interface is 64 bits instead of 32. The design had been carefully crafted to support BYTE, WORD and DWORD reads and writes, from a 32-bit wide memory interface. However, it did not properly support rotated (unaligned) 32-bit wide reads, of which all of the bytes were supposed to be correct. AFAIK, the USB was the only client that used this strange not-to-be-compatible-with-anything-ever-built method. So I standardized the USB DMA controller such that is always uses aligned 32-bit reads from memory (two lowest address bits at '00'). It still supports unaligned addresses as source, but it now does the proper rotation in the DMA controller instead of relying on the memory sub-system.

PR #825, opened 2026-09-01
Improve E2E stability

chrisgleissner:
# Summary

This PR improves the E2E stability. 

## What this fixes

`./run-tests c64u -s ftp-usb-integrity` was supposed to go red on the FTP-to-USB corruption of #803 and green on firmware without it. It did neither. Three separate problems were in the way.

### The UI-state gate abandoned the run

```
UI state dirty: the menu will not open; the UI task is blocked in a modal; repairing
UI state check failed: the menu will not open; the UI task is blocked and RUN/STOP did not release it
ftp-usb-integrity: FAIL (precondition: the device is unhealthy, 0.000s)
```

The C64 Ultimate was answering REST, running the C64 (the jiffy clock at `$00A2` was advancing) and serving FTP, while `PUT /v1/machine:menu_button` returned 200 and `GET /v1/machine:menu_screen` stayed at 404 through repeated presses and RUN/STOP. The UI task was blocked somewhere no keystroke reaches. A single `PUT /v1/machine:reset` released it, and the menu opened on the next press.

`repair()` already keeps that reset as its last resort, but each round began with `open_menu()`, which raises on a menu that will not open. Round 0 therefore raised before the reset could run, and the gate reported the device as unhealthy.

The same hole existed for the other wedge shape. A run on an Ultimate II+L reported "the menu will not close; the UI task is not reading injected keys" and abandoned the rest of that run; `close_menu()` raises from inside a repair round, so the reset at the end of the round was equally unreachable.

Both are now fixed the same way. The open is split into `try_open_menu()`, which reports rather than raises. Both wedge messages are raised as `UiWedged`, a subclass of `Unrecoverable`, and a repair round catches it and resets the machine. A device that has gone off the network still raises plain `Unrecoverable` and is reported at once, rather than after four futile resets.

### The suite skipped itself

`--usb-dir` defaulted to `/USB0`. The device names a volume after the USB port its medium is in, so the same stick is `/USB0` on the Ultimate II+L and `/USB2` on both the C64 Ultimate and the Ultimate 64. On those two the suite found no `/USB0`, skipped, and the run reported OK. The volume is now discovered by the new `ftp.usb_volumes()` helper, which returns the served `USB<n>` volumes ordered by port number; `--usb-dir` still overrides it.

### The write shape did not reproduce the defect

The `ftp-usb-integrity` regression check wrote 1350 bytes at a time. On the affected firmware that round-tripped byte-exact four times in a row. Measured against the same device and the same volume:

| write shape | TCP_NODELAY | wrong words in 64 KiB |
|---|---|---|
| 1350 B throughout | no | 0 |
| 1350 B throughout | yes | 0 |
| 1024 B, 2 B, then 1024 B | no | 2 |
| 1024 B, 2 B, then 1024 B | yes | 59 |

The two-byte write is what leaves every write after it at 2 modulo 512, which is the file offset the defect needs. `TCP_NODELAY` matters because without it Nagle can join two client writes into one segment, the device's `recv` sees them joined, and the offset is never formed.

### One suite for #803 instead of two

Two manual suites covered this defect: `usb-bulk-out-integrity` (`tests/e2e/io/usb/bulk_out_integrity_test.py`, added by #824) carried the reproducing write shape and one check; `ftp-usb-integrity` (`tests/e2e/network/ftp_usb_integrity_test.py`, added by #816) carried a shape that does not reproduce it, plus two checks that make a failure interpretable.

They are merged into one suite under the original `usb-bulk-out-integrity` name and path. It runs three checks: the 700-byte control shape on USB, which must stay green because it exercises the same file, medium and link; the reproducing shape on `/Temp`, which is the layer split; and the reproducing shape on USB, which is the regression. `describe()` reported the wrong geometry for this shape — it counted words at offset 508 of a 1024-byte pair, and the corruption is at 1020 — so it now counts what was measured: the last word of each 1024-byte pair, with only the top byte wrong.

Sample output on the C64 Ultimate:

```
[01] a client writing 700 B at a time stores the file intact ... OK (2.109s)
[02] the same client shape stores the file intact on /Temp ... OK (1.408s)
[03] a client writing 1024 B, 2 B, then 1024 B at a time stores the file intact ... FAIL (1.522s)
     61 word(s) differ; 61 at offset 1020 of a 1024-byte pair, 61 with only the top byte wrong
     0x000FFC expected A50003FF got 020003FF
```

## Test plan

- `./run-tests c64u -s ftp-usb-integrity` (before the rename) — FAIL on the regression check, control and RAM-disk checks green. C64 Ultimate, firmware 1.2.0, fpga 122, core 1.4D.
- `python3 tests/e2e/io/usb/bulk_out_integrity_test.py -H c64u -t 30` (after the rename) — same result, 61 wrong words.
- `./run-tests u64 -s usb-bulk-out-integrity` — suite OK. Ultimate 64 Elite, firmware 3.15, fpga 123, core 1.4F, volume `/USB2`.
- `python3 tests/e2e/network/ftp_usb_integrity_test.py -H u2 -t 30` (pre-rename path, same content) — OK. Ultimate II+L, firmware 3.15, fpga 123, volume `/USB0`. Run directly rather than through the runner because a bare `u2` target does not satisfy the UI-state gate; that is unrelated to this change and behaves the same on `test-merge`.
- Checked all three devices afterwards: no `usb_bulk_out_*` or `ftp_usb_integrity_*` files left on any USB volume or in `/Temp`.
- Device-free suites: `runner-policy`, `transport-usage`, `esp-depends`, `input-batching`, `openapi-validator`, `observability` all pass.

`runner_policy_test.py` gains three checks for the repair ladder, driving `repair()` against a stub device: one whose UI ignores the menu button until it is reset, one whose menu opens but then ignores both RUN/STOP and the button, and one the reset never frees, which must raise after exactly `REPAIR_ROUNDS` resets. Restoring `open_menu(device)` at the top of the round loop makes the first fail; re-raising `UiWedged` instead of resetting makes the second fail.

## Not addressed

The firmware defect itself. The alignment guard that would catch the unaligned source pointer is live in `UsbBase::bulk_in` and commented out in `bulk_out`; #821 carries a workaround for a bulk-out bounce condition described as `(buffer alignment + length) mod 4 != 0`, which is not the same geometry as the 512-byte-aligned two-sector `disk_write` reproduced here. Whether #821 covers this case can now be answered by running this suite.

A third gate message, "the menu reported open but returned no screen", is still a plain `Unrecoverable`. It was seen once on the C64 Ultimate while another session was driving the same device, so it may be a race between two clients rather than a wedge. It is left alone rather than given a reset it may not need.

https://claude.ai/code/session_01JejHYaEnab55t9HgP9oMkB


chrisgleissner:
Superseded by #826, which carries `c8fc5a0e` and `b64695a6` from this branch unchanged and adds the rest of the C64 Ultimate work on top. Closing so there is one branch and one PR to review.

https://claude.ai/code/session_01L4whbWW6NFfYAg1cTo7JKh

PR #827, opened 2026-09-02
Report the interface MAC addresses from GET /v1/info

enver-haase:
### What

`GET /v1/info` gains three optional fields:

```json
{
  "product" : "Ultimate 64 Elite (V1.49) 3.14d",
  "firmware_version" : "3.14d",
  "git_commit_hash" : "5ee21b65",
  ...
  "ethernet_mac" : "02:15:41:C8:90:85",
  "wifi_mac" : "24:0A:C4:1A:2B:3C",
  "errors" : []
}
```

### Why the MAC addresses

Wake-on-Wi-Fi matches a magic packet, and a magic packet carries a MAC address, not an IP one. `wol_is_magic_packet()` compares against the station netif's `hwaddr`, which is the Wi-Fi STA address the firmware also gets back from `CMD_WIFI_GETMAC` — so `wifi_mac` is exactly the address a remote client has to send to.

Until now that address was on the configuration screen and nowhere else, so an app that wants a "switch it on" button had to have the user type it in or dig it out of an ARP table, both of which are wrong the first time the device is on a network the app has never seen it on. `ethernet_mac` is there for the wired side of the same question.

### Why the commit hash

A version number does not say which build is running. Between two firmwares that both call themselves 3.15 there is nothing in the API to tell them apart, so a bug report against "3.15" cannot be tied to a tree.

The firmware already knows: the `gitinfo` rule in `target/common/rules.mk` writes `APP_VERSION_HASH` from `git rev-parse --short HEAD`, and the System Information screen has been showing it all along. `/v1/info` reports the same string, so a caller on the network can read it without someone walking up to the machine.

### How

The addresses come from the interface registry, not from a product test:

* `NetworkInterface::is_wireless()` is new, `false` in the base class and `true` in `NetworkLWIP_WiFi`. It is the only thing that decides which of the two keys an interface fills.
* The first interface of each kind wins; the rest are skipped, so a USB Ethernet adapter does not displace the on-board one.
* An interface whose address is still all zeroes, because it never started, is left out entirely rather than reported as `00:00:00:00:00:00`. A build without a Wi-Fi module therefore has no `wifi_mac` at all, and no `#ifdef` is needed to arrange that.

Formatting matches what the configuration screen shows: upper case, colon separated.

### Documentation

`API_DOC` and `InfoResponse` in `tools/openapi/schemas.py` describe all three fields and when each is absent; `make openapi` regenerated both documents, and `make openapi_check` and `make openapi_test` pass. The size row in `doc/api/openapi-generation.md` is updated with the new byte counts. A PR against `1541ultimate-documentation` can follow once this one has a shape everybody is happy with.

### Tests

`tests/e2e/api/rest_api_coverage_test.py` gets two cases on `GET /v1/info`:

* the shape of whichever MAC addresses are reported, and that at least one is: the request it just answered arrived over some interface, so that interface knows its own address.
* `git_commit_hash` is present and looks like an abbreviated hash. An empty one means the firmware was built outside a git checkout, and a field that cannot answer what is running is worse than no field at all.

### Build

Built this time, unlike the first round: `make u64ii_no_esp` on this branch is green in `ghcr.io/gideonz/riscv`, `routes.cc` compiles and links without a new warning, and `app_space_test` and `openapi_check` pass as part of it. The generated `gitinfo.h` of that build reads `APP_VERSION_HASH "270e48587"`, which is what the new field returns.

The E2E cases have not run against hardware yet — that needs the branch flashed, and the updater asks its questions at the local keyboard.

### DIFF
--- a/doc/api/openapi-generation.md
+++ b/doc/api/openapi-generation.md
@@ -18,8 +18,8 @@ time, so one document cannot describe both families:
 
 | document | products | paths | operations | size |
 | --- | --- | --- | --- | --- |
-| `rest_api_openapi_u2.yaml` | Ultimate II, II+, II+L | 44 | 55 | 130,826 B |
-| `rest_api_openapi_u64.yaml` | Ultimate 64, 64 Elite, 64 Elite II | 47 | 59 | 141,412 B |
+| `rest_api_openapi_u2.yaml` | Ultimate II, II+, II+L | 44 | 55 | 131,333 B |
+| `rest_api_openapi_u64.yaml` | Ultimate 64, 64 Elite, 64 Elite II | 47 | 59 | 141,919 B |
 
 The Ultimate 64 document is larger because `route_streams.cc` is compiled only by
 its makefiles, and because `machine:debugreg` sits behind `#if U64`.
--- a/doc/api/openapi-generation.md
+++ b/doc/api/openapi-generation.md
@@ -18,8 +18,8 @@ time, so one document cannot describe both families:
 
 | document | products | paths | operations | size |
 | --- | --- | --- | --- | --- |
-| `rest_api_openapi_u2.yaml` | Ultimate II, II+, II+L | 44 | 55 | 129,837 B |
-| `rest_api_openapi_u64.yaml` | Ultimate 64, 64 Elite, 64 Elite II | 47 | 59 | 140,423 B |
+| `rest_api_openapi_u2.yaml` | Ultimate II, II+, II+L | 44 | 55 | 130,826 B |
+| `rest_api_openapi_u64.yaml` | Ultimate 64, 64 Elite, 64 Elite II | 47 | 59 | 141,412 B |
 
 The Ultimate 64 document is larger because `route_streams.cc` is compiled only by
 its makefiles, and because `machine:debugreg` sits behind `#if U64`.
--- a/doc/api/rest_api_openapi_u2.yaml
+++ b/doc/api/rest_api_openapi_u2.yaml
@@ -1812,6 +1812,8 @@ paths:
       description: |2-
         Identifies the device: product name, firmware version, FPGA build number, host name, unique id and hardware addresses. `core_version` is the version of the C64 core and is present on Ultimate 64 hardware only. `unique_id` is left out when the device has none configured.
 
+        `git_commit_hash` is the abbreviated hash of the commit the firmware was built from, the same one the System Information screen shows. It tells two builds that carry the same version number apart.
+
         `ethernet_mac` and `wifi_mac` are the addresses of the wired and the wireless interface. Each is left out when the device has no such interface, or when that interface has not started and its address is therefore not known yet. They are what a caller sends a Wake-on-LAN magic packet to, which carries a MAC and not an IP address.
 
         The same information is also available without HTTP, from the Ultimate Ident Service on UDP port 64, which answers a broadcast and so can be used to find a device whose address is not known.
@@ -1828,6 +1830,7 @@ paths:
                   value:
                     product: Ultimate 64 Elite (V1.49) 3.14d
                     firmware_version: '3.14d'
+                    git_commit_hash: '5ee21b65'
                     fpga_version: '122'
                     core_version: '1.49'
                     hostname: Ultimate-64-Elite-C89085
@@ -1839,6 +1842,7 @@ paths:
                   value:
                     product: '1541 Ultimate II+ 3.14d'
                     firmware_version: '3.14d'
+                    git_commit_hash: '5ee21b65'
                     fpga_version: '11f'
                     hostname: Ultimate-II-Plus
                     ethernet_mac: '02:15:41:AA:BB:CC'
@@ -3441,6 +3445,9 @@ components:
               description: Full product name, including the core version on Ultimate 64 hardware.
             firmware_version:
               type: string
+            git_commit_hash:
+              type: string
+              description: Abbreviated hash of the commit the firmware was built from. It tells two builds that carry the same version number apart.
             fpga_version:
               type: string
               description: FPGA build number in hexadecimal.
--- a/doc/api/rest_api_openapi_u2.yaml
+++ b/doc/api/rest_api_openapi_u2.yaml
@@ -1810,7 +1810,9 @@ paths:
         - About
       summary: Product and version information
       description: |2-
-        Identifies the device: product name, firmware version, FPGA build number, host name and unique id. `core_version` is the version of the C64 core and is present on Ultimate 64 hardware only. `unique_id` is left out when the device has none configured.
+        Identifies the device: product name, firmware version, FPGA build number, host name, unique id and hardware addresses. `core_version` is the version of the C64 core and is present on Ultimate 64 hardware only. `unique_id` is left out when the device has none configured.
+
+        `ethernet_mac` and `wifi_mac` are the addresses of the wired and the wireless interface. Each is left out when the device has no such interface, or when that interface has not started and its address is therefore not known yet. They are what a caller sends a Wake-on-LAN magic packet to, which carries a MAC and not an IP address.
 
         The same information is also available without HTTP, from the Ultimate Ident Service on UDP port 64, which answers a broadcast and so can be used to find a device whose address is not known.
       operationId: getInfo
@@ -1830,6 +1832,8 @@ paths:
                     core_version: '1.49'
                     hostname: Ultimate-64-Elite-C89085
                     unique_id: D09B96
+                    ethernet_mac: '02:15:41:C8:90:85'
+                    wifi_mac: '24:0A:C4:1A:2B:3C'
                     errors: []
                 Ultimate II+:
                   value:
@@ -1837,6 +1841,7 @@ paths:
                     firmware_version: '3.14d'
                     fpga_version: '11f'
                     hostname: Ultimate-II-Plus
+                    ethernet_mac: '02:15:41:AA:BB:CC'
                     errors: []
         '403':
           $ref: '#/components/responses/Forbidden'
@@ -3447,6 +3452,12 @@ components:
             unique_id:
               type: string
               description: Omitted when the device has no unique id configured.
+            ethernet_mac:
+              type: string
+              description: MAC address of the wired interface, upper case, colon separated. Omitted while the interface has not started and its address is not known.
+            wifi_mac:
+              type: string
+              description: MAC address of the Wi-Fi interface, upper case, colon separated. Omitted on hardware without a Wi-Fi module, and while the module has not started and its address is not known.
     InputBatch:
       type: object
       description: A batch of input events, applied in order. The whole batch is validated before any of it is applied, so a rejected batch changes nothing.
--- a/doc/api/rest_api_openapi_u64.yaml
+++ b/doc/api/rest_api_openapi_u64.yaml
@@ -1813,7 +1813,9 @@ paths:
         - About
       summary: Product and version information
       description: |2-
-        Identifies the device: product name, firmware version, FPGA build number, host name and unique id. `core_version` is the version of the C64 core and is present on Ultimate 64 hardware only. `unique_id` is left out when the device has none configured.
+        Identifies the device: product name, firmware version, FPGA build number, host name, unique id and hardware addresses. `core_version` is the version of the C64 core and is present on Ultimate 64 hardware only. `unique_id` is left out when the device has none configured.
+
+        `ethernet_mac` and `wifi_mac` are the addresses of the wired and the wireless interface. Each is left out when the device has no such interface, or when that interface has not started and its address is therefore not known yet. They are what a caller sends a Wake-on-LAN magic packet to, which carries a MAC and not an IP address.
 
         The same information is also available without HTTP, from the Ultimate Ident Service on UDP port 64, which answers a broadcast and so can be used to find a device whose address is not known.
       operationId: getInfo
@@ -1833,6 +1835,8 @@ paths:
                     core_version: '1.49'
                     hostname: Ultimate-64-Elite-C89085
                     unique_id: D09B96
+                    ethernet_mac: '02:15:41:C8:90:85'
+                    wifi_mac: '24:0A:C4:1A:2B:3C'
                     errors: []
                 Ultimate II+:
                   value:
@@ -1840,6 +1844,7 @@ paths:
                     firmware_version: '3.14d'
                     fpga_version: '11f'
                     hostname: Ultimate-II-Plus
+                    ethernet_mac: '02:15:41:AA:BB:CC'
                     errors: []
         '403':
           $ref: '#/components/responses/Forbidden'
@@ -3722,6 +3727,12 @@ components:
             unique_id:
               type: string
               description: Omitted when the device has no unique id configured.
+            ethernet_mac:
+              type: string
+              description: MAC address of the wired interface, upper case, colon separated. Omitted while the interface has not started and its address is not known.
+            wifi_mac:
+              type: string
+              description: MAC address of the Wi-Fi interface, upper case, colon separated. Omitted on hardware without a Wi-Fi module, and while the module has not started and its address is not known.
     InputBatch:
       type: object
       description: A batch of input events, applied in order. The whole batch is validated before any of it is applied, so a rejected batch changes nothing.
--- a/doc/api/rest_api_openapi_u64.yaml
+++ b/doc/api/rest_api_openapi_u64.yaml
@@ -1815,6 +1815,8 @@ paths:
       description: |2-
         Identifies the device: product name, firmware version, FPGA build number, host name, unique id and hardware addresses. `core_version` is the version of the C64 core and is present on Ultimate 64 hardware only. `unique_id` is left out when the device has none configured.
 
+        `git_commit_hash` is the abbreviated hash of the commit the firmware was built from, the same one the System Information screen shows. It tells two builds that carry the same version number apart.
+
         `ethernet_mac` and `wifi_mac` are the addresses of the wired and the wireless interface. Each is left out when the device has no such interface, or when that interface has not started and its address is therefore not known yet. They are what a caller sends a Wake-on-LAN magic packet to, which carries a MAC and not an IP address.
 
         The same information is also available without HTTP, from the Ultimate Ident Service on UDP port 64, which answers a broadcast and so can be used to find a device whose address is not known.
@@ -1831,6 +1833,7 @@ paths:
                   value:
                     product: Ultimate 64 Elite (V1.49) 3.14d
                     firmware_version: '3.14d'
+                    git_commit_hash: '5ee21b65'
                     fpga_version: '122'
                     core_version: '1.49'
                     hostname: Ultimate-64-Elite-C89085
@@ -1842,6 +1845,7 @@ paths:
                   value:
                     product: '1541 Ultimate II+ 3.14d'
                     firmware_version: '3.14d'
+                    git_commit_hash: '5ee21b65'
                     fpga_version: '11f'
                     hostname: Ultimate-II-Plus
                     ethernet_mac: '02:15:41:AA:BB:CC'
@@ -3716,6 +3720,9 @@ components:
               description: Full product name, including the core version on Ultimate 64 hardware.
             firmware_version:
               type: string
+            git_commit_hash:
+              type: string
+              description: Abbreviated hash of the commit the firmware was built from. It tells two builds that carry the same version number apart.
             fpga_version:
               type: string
               description: FPGA build number in hexadecimal.
--- a/software/api/routes.cc
+++ b/software/api/routes.cc
@@ -4,6 +4,7 @@
 #include "stream_uart.h"
 #include "dump_hex.h"
 #include "network_config.h"
+#include "network_interface.h"
 #include "product.h"
 #include "versions.h"
 #include "u64.h"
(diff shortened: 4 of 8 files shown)


enver-haase:
Withdrawing this until it has been built and run on real hardware. Reopening once the firmware build and the E2E case have actually passed.

chrisgleissner:
@enver-haase I like the idea of exposing additional information via this endpoint *a lot*. 

Your suggestion seems great to me, and maybe we can also add information about the Git ID form which the firmware was built? This would make it much easier to determine what exactly is running.

I would suggest one of the names `git_id`, `git_commit_id`, or `git_commit_hash`. I think this should already be exposed somewhere. What do you think?

Of course the E2E tests need to pass. You also need a PR against the master branch of `1541ultimate-documentation`, but don't worry about that for now; it can come later.

Btw, the train for the final merge has not yet passed, but we want to avoid all large changes at this point unless they are for truly critical bugs or simple quality of life extensions. But the train will depart tomorrow (Thursday) noon based on the latest train schedule. :-)

enver-haase:
@chrisgleissner Reopened, rebased onto current `test-merge`, and the git id is in.

You were right that it already exists: the `gitinfo` rule in `target/common/rules.mk` writes `APP_VERSION_HASH` from `git rev-parse --short HEAD`, and `system_info.cc` has been printing it on the System Information screen all along. So the endpoint only had to report the string that was already there. I took `git_commit_hash` from your three names because that is literally what it is; say the word and it becomes `git_id` or `git_commit_id` instead, it is a one-line change while nothing depends on it yet.

The reason I withdrew this is now dealt with: `make u64ii_no_esp` is green in `ghcr.io/gideonz/riscv`, `routes.cc` compiles and links with no new warning, and `app_space_test` and `openapi_check` run as part of it. The generated `gitinfo.h` of that build reads `APP_VERSION_HASH "270e48587"`, which is exactly what the new field returns.

What is still open is the E2E run against real hardware. Both cases are written, but running them means flashing this branch, and the updater asks its questions at the local keyboard — so that happens when I am at the machine, not from here. I will report the result rather than claim it.

Understood on the train and on the documentation PR. Three optional response fields and no behaviour change elsewhere is about as small as an addition gets, so if the train has left, this rides the next one.

chrisgleissner:
@enver-haase @GideonZ Looks good to me. It is well-tested and I find it useful to expose this information, especially for testing and figuring out what exactly runs on a device.

enver-haase:
@GideonZ I have just flashed my C64U (IIRC was seen as a mass-produced V2.2 Ultimate Elite II by the updater) and it is BRICKED now.

@chrisgleissner Handing this over I am afraid — I am out of test hardware for the time being, so the last open item on this PR is not one I can close myself.

  Where it stands: rebased onto current test-merge, two commits. The first reports ethernet_mac and wifi_mac, the second git_commit_hash, the git id you asked for. The CI build check is green.

  What is verified. The firmware builds from a scratch tree — I wiped every output/ and result/ directory and the ESP32 build first, because an incremental build across the branch boundary silently reuses objects compiled
  with the other branch's flags, and this tree's two lines differ in more than one of them. make u64ii is green in ghcr.io/gideonz/riscv including the ESP32 stage, routes.cc compiles and links without a new warning, and
  app_space_test and openapi_check run as part of it. On the host, make openapi_check is clean and the 191 openapi tests pass. The built package identifies itself as 3.15 / 47fcbad8d, which is what the new field returns.

  What is not verified. Neither of the two E2E cases has ever run against a device. They are written and they are in the PR, but the only claim I can make about them is that they exist. Please run them, or tell me what they
  should say, rather than taking them on trust.

  Two things worth your judgement while you have it open. The field name is git_commit_hash, one of your three; git_id or git_commit_id is a one-line change while nothing depends on it. And the E2E case requires the hash to
  be present and non-empty, which means a firmware built outside a git checkout fails that case. I think that is right — a field that cannot answer what is running is worse than no field — but it is a gate on your build
  environment, not only on mine, so it is your call.

  The documentation PR against 1541ultimate-documentation can follow whenever the shape here is settled.

chrisgleissner:
Hi @enver-haase , 

very sorry to learn about the brick. Gideon told me that it was fine to flash the UE2 built by this repo, so not sure what exactly happened. Which firmware exactly did you flash, i.e. which Git ID, and did you reset your config as part of the flash?

First, please try and reboot the device whilst pressing the `RESTORE` key. This will prevent your custom config from taking effect.

If that still does not revive it, you may want to get an Adafruit FT232H. You will also need  a soldering iron to solder some Dupont cables to it. You can then connect these Dupont cables directly to the JTAG header of the C64U and use the script that Gideon added to this repo to revive it by temporarily installing a known good firmware, then flashing the original C64U firmware. If you run into issues, please reach out. 

Thanks a lot for your amazing contribution.

Best
Christian

chrisgleissner:
Validated on an Ultimate 64 Elite via JTAG at `61278870` (current `test-merge` is merged).

Red/green evidence using the unchanged `rest_api_coverage_test.py -H u64 -r 1 --order sequential -w 1`:
- Base `43a30ceb`: both new `GET /v1/info` checks failed: no interface MAC reported and no `git_commit_hash`.
- `61278870`: the 79-case run succeeded (normal skips only); `/v1/info` returned `git_commit_hash: 61278870`, `ethernet_mac: 02:15:41:F8:3C:87`, and `wifi_mac: E8:31:CD:1B:1C:2C`. Device health checks also passed.

Local validation passed: `make openapi_check openapi_test app_space_test`, full U64 build, U64-II application build, U2 application build, and U2+L application build. The U2 and U2+L software-only updater packaging stopped only at the expected absent FPGA inputs (`rv700dd.chk` and `u2p_ecp5_impl1.bit`) after their applications linked. No source changes were needed beyond the base merge.

enver-haase:
Thanks a lot.

I have already got an Adafruit FT232H here, but cannot start the efforts today due to time constraints.

Would be great if the REST extension could make it into 3.15, then the next release of the Ultimate Assembly64 app on Apple and Google stores could get a power switch.

I flashed an own build of the 3.15 branch thinking I could test the
/info point and the FPGA fix of the 32/64bit memory issue in one go, bad luck... The git hash is on the USB stick, will post it later, I am on mobile now.

Hash is `47fcbad8d`, my two commits on top of `test-merge` at `d33b7802e` so the FPGA fix was in, and yes, the config was reset as part of the flash.

Thanks for all the effort and looking fwd to an even greater new software release!

chrisgleissner:
Thanks @enver-haase , and good luck with the unbricking. 

Still trying to figure out what bricked it. To diagnose further could you please share this as well:
- How exactly does the C64U behave now? Which lights are still illuminated, if any? Does the Ethernet port on the back still blink - assuming you have a cable plugged in?
- For avoidance of doubt - you flashed the UE2 that is built by the test-merge build in this repo for Git ID `47fcbad8d`, right? Or was this something you built locally?

Best wishes
Christian

enver-haase:
Hi Christian,

The board is alive again — and Gideon has already fixed this in the meantime.

Yes, I reset configuration on every flash, including that one.

It wasn't my tree: the official CI artifact for this PR head (`update_package_v3.15-110-g4632a731`, run 33755108702) bricks the machine too. The cause is `FENCE` — that build carries the pre-`883f608d9` core, which decodes `FENCE` as illegal, so the soft CPU dies on the first barrier and the screen stays black.

`8a4e57714` and `883f608d9` fix exactly that. The current `test-merge` build, `ac2fe909`, boots fine here.

**LEDs**, as flashed and before JTAG recovery: https://drive.google.com/file/d/1W6PeFO3ag5C0SbXKXFLEA6WYyFuIdQtC/view?usp=sharing

The FT232H recovery worked — thank you for that.

Best,
Enver

UltimateC64Ultimate64fan:
Hello @chrisgleissner and @enver-haase,  
Cc: @GideonZ 

I was a bit worried about exposing mac addresses for no justifiable reason. I had a friend have a look at it and he provided the following evaluation.

---

# Evaluation of PR #827: Unauthenticated MAC Address Exposure and Hardware Realities

Adding hardware MAC addresses to `GET /v1/info` is **unjustified feature creep, architecturally broken across most hardware, and an unauthenticated privacy leak by default**.

The foundational premise of [PR #827](https://github.com/GideonZ/1541ultimate/pull/827), that callers require MAC addresses to send Wake-on-LAN (WoL) magic packets, collapses under hardware scrutiny:

* **Standby WoL Physically Exists ONLY on ESP32-S3 Boards (`U64 Elite MK2`):** 
Only the Ultimate 64 Elite II and Commodore 64 Ultimate supply an always-on 3.3V standby power rail to the co-processor. On older boards with the ESP32-WROOM (U64 Elite MK1 and original U64) and all cartridges (U2, U2+, U2+L), the chip loses power completely when the host machine is turned off. Exposing `wifi_mac` on non-S3 hardware provides 0% WoL capability.
* **Wired WoL Is Physically Impossible Everywhere:**
The LAN8720A RMII PHY and FPGA soft core are unpowered during standby across the entire product family. Exposing `ethernet_mac` is 100% dead weight on every board revision.
* **Native On-Device UI Already Covers User Needs:**
When disconnected, the root File Browser explicitly displays 
`MAC xx:xx:xx:xx:xx:xx Link Down`, 
and the `F2` menus show `Interface MAC` at all times for router/DHCP setup.
* **Unauthenticated Privacy Leak by Default:**
While `GET /v1/info` is gated when a `Network Password` is explicitly configured ([routes.h#L299-L312](https://github.com/GideonZ/1541ultimate/blob/test-merge/software/api/routes.h#L299-L312)), the factory default on all boards is empty (`""`), and companion software relies on zero-auth discovery. The PR added raw MAC addresses without checking `Product::is_unique_id_hidden()`, creating a permanent Layer 2 leak and WiGLE geolocation vector.

| Product / Hardware | Co-Processor | Standby Power Rail | Wake-on-LAN Capability | MAC Exposure in API |
| :--- | :--- | :--- | :--- | :--- |
| Ultimate 64 Elite II | ESP32-S3 | Always-on 3.3V Rail | Supported (Wi-Fi STA Only) | Justified if opt-in |
| Commodore 64 Ultimate | ESP32-S3 | Always-on 3.3V Rail | Supported (Wi-Fi STA Only) | Justified if opt-in |
| Ultimate 64 Elite (MK1) | ESP32-WROOM | Unpowered in Standby | Impossible (Hardware Off) | Dead Weight / Flaw |
| Ultimate 64 (Original) | ESP32-WROOM | Unpowered in Standby | Impossible (Hardware Off) | Dead Weight / Flaw |
| Ultimate-II+ / II+L | None / S3 Add-on | Unpowered (C64 Off) | Impossible (Cartridge Off) | Dead Weight / Flaw |
| Wired Ethernet (All) | LAN8720A RMII | Unpowered in Standby | Impossible (PHY Off) | Dead Weight (Revert) |

---

**1. Hardware Standby Power Reality: WoL Exists Only on ESP32-S3 (`U64 Elite MK2`)**

The Wake-on-Wi-Fi implementation ([wol_magic.c](https://github.com/GideonZ/1541ultimate/blob/test-merge/software/u64ctrl/main/wol_magic.c)) merged in [PR #815](https://github.com/GideonZ/1541ultimate/pull/815) relies on a dedicated standby power plane that exists strictly on newer motherboard designs:

* **ESP32-S3 Power Domain (`U64 Elite MK2`):** 
On the Ultimate 64 Elite II and Commodore 64 Ultimate, the ESP32-S3 is connected to an isolated 3.3V standby rail that remains energized while the main 12V/5V rails and FPGA core are off. This allows the ESP32-S3 to stay connected to Wi-Fi, filter for the AMD Magic Packet, and trigger the power controller.
* **ESP32-WROOM Power Domain (`U64 MK1`):** 
On the original Ultimate 64 and Ultimate 64 Elite (MK1), the ESP32-WROOM shares the main power plane. When the system enters standby, the ESP32-WROOM powers down entirely. It cannot receive radio packets or wake the system.
* **Cartridge Hardware (U2, U2+, U2+L):** 
Cartridges derive power from the C64 expansion port. When the C64 is powered off, the cartridge is unpowered. Wake-on-LAN is physically impossible.
* **Conclusion:** 
Broadcasting `wifi_mac` across all products under the guise of Wake-on-LAN ignores the physical board schematics. On non-S3 hardware, it serves no WoL function whatsoever.

---

**2. Wired Ethernet WoL Is Physically Impossible Across the Entire Line**

PR #827 explicitly stated:
> *"ethernet_mac is there for the wired side of the same question."*

This statement is contradicted by the hardware design:

* **Unpowered LAN8720A PHY:** 
The RMII Ethernet PHY is connected to the primary system power domain. In standby, the PHY loses power, the Ethernet magnetics drop carrier link, and the link LEDs extinguish.
* **Unpowered Soft MAC:** 
The Ethernet MAC is implemented inside the FPGA soft core ([rmii_interface.cc](https://github.com/GideonZ/1541ultimate/blob/test-merge/software/io/network/rmii_interface.cc)), which is halted and unclocked during standby.
* **Documented Concession in PR #842:** 
Downstream documentation [PR #842](https://github.com/GideonZ/1541ultimate/pull/842) was subsequently forced to admit: *"ethernet_mac cannot wake the device (the wired PHY is not powered in standby)"*.
* **Conclusion:** 
Exposing `ethernet_mac` provides 0% utility for Wake-on-LAN on any board. It is purely dead weight.

---

**3. Native UI Visibility: Hardware MACs Are Already Accessible Where Needed**

The assertion that the MAC address was inaccessible without digging through submenus or ARP tables is false. Gideon deliberately implemented clean on-device visibility:

* **File Browser Root (`/`):** 
When the Ethernet cable is unplugged or Wi-Fi is disconnected, [network_interface.cc#L296](https://github.com/GideonZ/1541ultimate/blob/test-merge/software/io/network/network_interface.cc#L296) and [network_esp32.cc#L94](https://github.com/GideonZ/1541ultimate/blob/test-merge/software/io/network/network_esp32.cc#L94) format the root row to show the MAC address with a red/brown `Link Down` label:  
  `Net0    MAC 02:15:41:F8:3C:87     Link Down`  
  `WiFi    MAC 24:0A:C4:1A:2B:3C     Link Down`  
  This allows users setting up static router DHCP leases or MAC filters to read the address directly on the C64 monitor without navigating menus.
* **Configuration Menus (`F2`):** 
`F2 -> Ethernet Settings` ([network_interface.cc#L35](https://github.com/GideonZ/1541ultimate/blob/test-merge/software/io/network/network_interface.cc#L35)) and `F2 -> WiFi Settings` ([network_esp32.cc#L56](https://github.com/GideonZ/1541ultimate/blob/test-merge/software/io/network/network_esp32.cc#L56)) permanently display `Interface MAC` regardless of link status.

---

**4. Privacy by Design Violation: Bypassing `Hide Unique ID` by Default**

While `GET /v1/info` is technically protected when a `Network Password` is configured ([routes.h#L299-L312](https://github.com/GideonZ/1541ultimate/blob/test-merge/software/api/routes.h#L299-L312)), the default state creates an unauthenticated leak:

* **Default Unauthenticated State:** In [network_config.cc#L18](https://github.com/GideonZ/1541ultimate/blob/test-merge/software/network/network_config.cc#L18), the password defaults to empty (`""`). Because zero-configuration companion apps operate without authentication, the vast majority of devices run with no password.
* **Circumventing Existing Privacy Gates:** In [routes.cc#L206-L209](https://github.com/GideonZ/1541ultimate/blob/test-merge/software/api/routes.cc#L206-L209), the device serial number is guarded:
  ```cpp
  const char *unique_id = get_unique_id();
  if (unique_id && !Product::is_unique_id_hidden()) {
      resp->json->add("unique_id", unique_id);
  }
  ```
  PR #827 appended raw hardware MAC addresses directly below this check with zero gating. A user who explicitly enables `Hide Unique ID` still has their permanent Ethernet and Wi-Fi MACs broadcast in the clear.
* **WiGLE Geolocation Mapping:** 
Wi-Fi BSSID/MAC addresses are indexed in public databases. Exposing `wifi_mac` unauthenticated allows any local script or browser cross-origin probe to pinpoint the user's physical street address.
* **1-Day Exploit Fingerprinting via `git_commit_hash`:** 
The reviewer-requested addition of `git_commit_hash` gives remote attackers exact commit-level precision to identify whether known open-source vulnerabilities remain unpatched on the target.

---

**5. Actionable Remediation Plan**

To eliminate dead weight and restore privacy compliance before 3.15 is tagged:

1. **Revert `ethernet_mac` Entirely:** 
Remove `ethernet_mac` from [routes.cc#L214-L236](https://github.com/GideonZ/1541ultimate/blob/test-merge/software/api/routes.cc#L214-L236). It is physically non-functional and serves no technical purpose.
2. **Restrict `wifi_mac` to S3 Hardware and Enforce Privacy Parity:** 
Only compile/emit `wifi_mac` on boards with actual standby hardware support (`U64 Elite MK2`), and strictly gate it behind the privacy setting:  
   `if (!Product::is_unique_id_hidden()) { resp->json->add("wifi_mac", mac_str); }`  
   Or provide a dedicated toggle in `Network Settings` (`Expose MAC in API: Disabled/Enabled`), defaulting to `Disabled`.
3. **Preserve Clean Architectural Boundaries:** 
Third-party companion apps requiring a device MAC for Wake-on-Wi-Fi should prompt the user to input the address once during setup or use an authenticated pairing handshake, rather than broadcasting Layer 2 identifiers over unauthenticated public REST API  end points/routes.

UltimateC64Ultimate64fan:
My friend also has a proposal which provides a solution which works for all U64* devices.

---

# Proposal: Universal Remote Power Management via "Auto Power-On on AC Connect"

**Summary**

Recent pull requests ([PR #815](https://github.com/GideonZ/1541ultimate/pull/815) and [PR #827](https://github.com/GideonZ/1541ultimate/pull/827)) demonstrate a clear and growing interest among users and companion-app developers in remotely powering on the Ultimate 64. 

However, implementing remote power-on via network-based Wake-on-Wi-Fi introduces substantial hardware incompatibilities, continuous standby power draw, and privacy concerns surrounding the unauthenticated broadcast of hardware MAC addresses.

We propose a simpler, universally compatible, and standard embedded alternative: introducing an **"Auto Power-On on AC Connect"** configuration option in the power supervisor. This approach mirrors the standard "AC Power Loss / Restore" setting found in personal computer BIOSes, providing immediate, secure remote power management across **100% of Ultimate 64 products** through standard off-the-shelf smart plugs.

---

**Architectural Comparison: Wake-on-Wi-Fi vs. AC Power Management**

| Evaluation Dimension | Wake-on-Wi-Fi (PR #815 / PR #827) | "Auto Power-On on AC" (Proposed) |
| :--- | :--- | :--- |
| **Hardware Compatibility** | Restricted strictly to ESP32-S3 boards (`U64 Elite MK2`). Incompatible with older U64 boards and all cartridges. | **100% Universal:** Operates on all Ultimate 64 boards (Original, Elite MK1, Elite II, C64 Ultimate) and requires no Wi-Fi hardware. |
| **Standby Power Consumption** | Requires continuous 3.3V power to the Wi-Fi radio 24/7 while the machine is powered off. | **True 0.0 Watts:** Entire system and radio remain completely unpowered until energized by the smart plug. |
| **Security & Privacy Footprint** | Exposes Layer 2 MAC addresses on unauthenticated HTTP endpoints (`GET /v1/info`) and opens local network WoL listeners. | **Zero Network Exposure:** Power switching occurs entirely out-of-band at the mains level; no API exposure or MAC harvesting required. |
| **Smart Home Ecosystems** | Proprietary to specific companion apps; unsupported by mainstream home automation platforms. | **Universal Compatibility:** Works seamlessly with Philips Hue, Apple HomeKit, Home Assistant, Zigbee, Z-Wave, Matter, and Shelly relays. |
| **Implementation Complexity** | High: Custom UDP listeners, packet filter hooks, REST API routes, and documentation adjustments. | **Minimal:** 5-10 lines of configuration logic in the ESP32 power supervisor ([button_handler.c](https://github.com/GideonZ/1541ultimate/blob/test-merge/software/u64ctrl/main/button_handler.c)). |

---

**This Direction is Architecturally Superior**

**1. Universal Parity Across All Product Generations**  
Wake-on-Wi-Fi is physically limited by board-level power domains. On original Ultimate 64 and Ultimate 64 Elite (MK1) motherboards, the ESP32-WROOM co-processor is powered down completely when the system enters standby. Standby packet filtering is physically impossible on these boards. In contrast, an AC power-on setting executes immediately upon power supply initialization, delivering complete feature parity across all hardware revisions regardless of which ESP32 variant or board revision a user owns.

**2. Restoring the Authentic Mechanical Switch Experience**  
The original Commodore 64 utilized a mechanical rocker switch. Users who connect authentic hardware to a smart plug can switch their systems on and off remotely with zero latency. On the Ultimate 64, the momentary push-button design defaults to standby (`machine_on = 0`) when mains power is applied. Adding an "Auto Power-On" option restores the intuitive behavior of the original mechanical switch, allowing users to integrate their retro setup into standard home automation routines with zero friction.

**3. Eliminating Network and Privacy Complications**  
PR #827 added `ethernet_mac` and `wifi_mac` to `GET /v1/info` primarily so external applications could cache the MAC address required to craft WoL magic packets. As documented during review, this creates two issues:
* `ethernet_mac` is dead weight because the wired PHY is unpowered in standby.
* Broadcasting permanent hardware MAC addresses over unauthenticated REST routes circumvents the existing `Hide Unique ID` privacy setting and exposes devices to WiGLE-based geolocation indexing.

Handling power switching at the mains level eliminates the need to expose Layer 2 hardware identifiers over the REST API altogether.

**4. True Zero-Watt Energy Conservation**  
Maintaining an active Wi-Fi receiver in standby requires continuous power dissipation and persistent RF activity. With smart-plug power management, the machine draws strictly 0.0 Watts when turned off, prolonging component lifespan and aligning with modern energy efficiency expectations.

---

**Proposed Implementation Outline**

The core logic can be implemented directly within the power supervisor task in [button_handler.c](https://github.com/GideonZ/1541ultimate/blob/test-merge/software/u64ctrl/main/button_handler.c). 

Currently, cold boot defaults to standby:
```c
// Current behavior: defaults to soft-off (0)
int initial_power_state = *((int *)arg);
int machine_on = initial_power_state;
regulator_enable(initial_power_state, 0);
```

By reading an NVS-backed preference (e.g., `auto_power_on`):
```c
// Proposed enhancement:
int auto_power_on = nvs_get_auto_power_on(); // Defaults to 0 (Standby)
int machine_on = initial_power_state || auto_power_on;

if (machine_on) {
    regulator_enable(1, 0);
    ESP_LOGI(TAG, "Auto Power-On triggered on AC connect");
} else {
    regulator_enable(0, 0);
}
```

A corresponding menu item can be exposed in the firmware's setup interface under `System Settings`:
* **Power State on AC Connect:** `[ Standby | Power On ]` (Default: `Standby`)

---

**Conclusion**

Remote power control is a valuable feature, but attempting to achieve it through network-layer Wake-on-LAN introduces significant software complexity, hardware fragmentation, and privacy trade-offs. 

Providing a standard **"Auto Power-On on AC Connect"** toggle addresses the root requirement cleanly, reliably, and securely for **all** Ultimate 64 users, while seamlessly integrating with the broader smart home ecosystem.

GideonZ:
Fair. I think for now the quickest route is removing the MAC address from the route, and move the MAC address into the configuration of the E2E tests. Agreed, @chrisgleissner ?

enver-haase:
This was a very creative (or shall I say generative) friend:
```
/* Circumventing Existing Privacy Gates: In [routes.cc#L206-L209](https://github.com/GideonZ/1541ultimate/blob/test-merge/software/api/routes.cc#L206-L209), the device serial number is guarded: */

const char *unique_id = get_unique_id();
if (unique_id && !Product::is_unique_id_hidden()) {
    resp->json->add("unique_id", unique_id);
}
```
The cited lines read entirely different, but so be it.

I already corrected "myself" (as mentioned, I am using a para-social friend entity myself) sayng that Wired WOL does not exist. Putting MAC retrieval here and not behind authentication, and including Wired Ethernet across all platforms including those that cannot Wake On WiFi, is debatable and I did for a moment hesitate myself.

However, may I re-iterate: I would like my Ultimate Assembly64 App ( iOS: https://apps.apple.com/en/app/ultimate-assembly64-app/id6768698983 , Android: get access here: https://groups.google.com/g/ultiassy-testers ) to be able to wake up the C64U, if that means authentication is required for that so be it. Neither platform offers a supported way to obtain the MAC of another host on the LAN — Android blocks reading /proc/net/arp for ordinary apps, and iOS has no public API for it.

`/v1/info` is exactly as protected as `PUT /v1/machine:poweroff` so we can turn the machine off unathenticated at the moment, I don't see why we should introduce an asymmetry here.

Maybe, however, we should question any unauthenticated access.

How about
```
// Only when a password is set, which means this request was authenticated.
if (*networkConfig.cfg->get_string(CFG_NETWORK_PASSWORD)) {
    resp->json->add("wifi_mac", mac_str);
}
```
?

PR #831, opened 2026-09-02
Fix Firefox compatibility. Remember selected theme for the current session.

radius75:
`html/index.html` and `html/api.html` each followed the browser's `prefers-color-scheme` setting on load. The theme buttons switched the theme, but the choice was lost on the next navigation, so moving between the web UI and the API explorer reset it.

## What changed

Both pages now store a picked theme in `sessionStorage` under the key `theme`, with the values `darkmode` and `lightmode`, so a choice made on either page holds for the rest of the browser tab and is honoured by the other page.

Only a theme the reader picked is stored. Following the browser's setting is not a choice, so it is applied without being written, which leaves a later change of that setting free to take effect.

Reads and writes of `sessionStorage` are wrapped in try/catch. A browser set to block on-device site data throws a `SecurityError` on access, and both pages read the value above the code that installs their handlers: unguarded, that left `api.html` on its offline notice with Swagger UI never started, and left `index.html` with its navigation hidden and no login or welcome screen.

`api.html` carries a Content-Security-Policy that admits its inline script by SHA-256 digest, so the digest is recomputed with `tools/openapi/explorer.py` to match the edited script.

In `index.html`, the themed-selector list drops two duplicate entries and `div#left-nav a`, which the plain `a` entry already covers.

## New test

`tests/e2e/web/theme_test.py` drives both pages in an installed Chrome and Firefox through Selenium. It needs no device: it serves `html/` from a local HTTP server, and `--base-url` points it at a running device instead. Registered in `run-tests` as the `web-theme` suite, marked manual because CI has no browsers.

Every condition is a real browser preference rather than anything injected into the page. Blocked site data is `profile.default_content_setting_values.cookies = 2` in Chrome and `network.cookie.cookieBehavior = 2` in Firefox, both verified to raise a real `SecurityError`. The colour scheme is CDP `Emulation.setEmulatedMedia` in Chrome and the `ui.systemUsesDarkTheme` preference in Firefox. The clicks and the page navigation are real WebDriver actions.

A missing browser or WebDriver is reported as one skip naming the reason, not a failure. `index.html` loads jQuery from a CDN and does nothing without it, so on a host with no route to that CDN the `index.html` checks skip and the `api.html` checks still run.

## Testing

```
./run-tests -s web-theme    26 checks, 26 OK, 34.5s
```

13 checks in each browser: the browser setting is followed on first load and not stored; a picked theme is stored, carries between the two pages in both directions, and is honoured on the next load; both pages stay usable with site data blocked, including their theme buttons; and the inline script of `api.html` runs under that page's own CSP.

The suite fails on the code it is meant to catch:

- Against the pages before the fix: `[01] index.html follows a light browser and stores nothing ... FAIL (what is stored is 'lightmode', expected None)`, in both browsers.
- Against the fixed pages carrying the previous CSP digest: `[04] api.html follows a dark browser and stores nothing ... FAIL (the theme is 'lightmode', expected 'darkmode')`, in both browsers. A refused inline script produces no console error, which is why this check is worth having.

`tools/openapi/test_explorer.py` passes, 16 tests, which is what pins the CSP digest to the script beside it. The device-free runner gates still pass with the new suite registered: `runner_policy_test` (99 checks), `check_transport_usage` (97 files) and `observability_test` (172 cases).

Chrome 152.0.7977.64 and Firefox 154.0.1 on Linux. Nothing was flashed to a device.

## Open point

The CSS change from `div#left-nav a:hover.darkmode` to `div#left-nav a.darkmode:hover` is described in its commit message as replacing a WebKit-specific selector. There is no WebKit-specific selector in either file, and the two forms behave identically in Chrome 152 and Firefox 154: both survive parsing, both apply burlywood on dimgray under a real hover, and they have the same specificity. The Selectors specification constrains only the type selector to come first in a compound selector. The change reads better, so it is left in, but it is a style preference rather than a compatibility fix, and it can be dropped if the diff should stay narrow.

### DIFF
--- a/html/api.html
+++ b/html/api.html
@@ -29,7 +29,7 @@
 -->
 <meta http-equiv="Content-Security-Policy" content="
     default-src 'none';
-    script-src 'unsafe-inline' https://unpkg.com/swagger-ui-dist@5.32.14/;
+    script-src 'sha256-LZqNxYBbhieAv4SV0EsDhOQ4IIXf7gtaYlYPOppWrtk=' https://unpkg.com/swagger-ui-dist@5.32.14/;
     style-src 'unsafe-inline' https://unpkg.com/swagger-ui-dist@5.32.14/;
     img-src 'self' data:;
     font-src 'self' data:;
--- a/html/api.html
+++ b/html/api.html
@@ -29,7 +29,7 @@
 -->
 <meta http-equiv="Content-Security-Policy" content="
     default-src 'none';
-    script-src 'sha256-LZqNxYBbhieAv4SV0EsDhOQ4IIXf7gtaYlYPOppWrtk=' https://unpkg.com/swagger-ui-dist@5.32.14/;
+    script-src 'sha256-aqK8YQoFANsW5zAcGrfy1JSRJISHSHowXSVLT070ZNo=' https://unpkg.com/swagger-ui-dist@5.32.14/;
     style-src 'unsafe-inline' https://unpkg.com/swagger-ui-dist@5.32.14/;
     img-src 'self' data:;
     font-src 'self' data:;
@@ -105,14 +105,38 @@
 // Follows the browser's setting on load and switches for the session, which is
 // what html/index.html does. Outside window.onload below, so the offline notice
 // is themed too, on a page where the bundle never arrives.
-function setDarkMode(dark, save = true) {
-    document.documentElement.classList.toggle("dark-mode", dark);
-    if (save) {
-        sessionStorage.setItem("theme", dark ? "darkmode" : "lightmode");
+
+// Reading or writing sessionStorage throws a SecurityError in a browser that is
+// set to block site data. This script installs the theme buttons and the
+// window.onload handler that starts Swagger UI, so an unguarded throw here would
+// leave the page on its offline notice for good. The theme is not worth that.
+function readTheme() {
+    try {
+        return sessionStorage.getItem("theme");
+    } catch (e) {
+        return null;
     }
 }
 
-const savedTheme = sessionStorage.getItem("theme");
+function writeTheme(themeName) {
+    try {
+        sessionStorage.setItem("theme", themeName);
+    } catch (e) {
+        // A browser that refuses to store the theme still gets to show it.
+    }
+}
+
+function setDarkMode(dark, save = true) {
+    document.documentElement.classList.toggle("dark-mode", dark);
+    if (save) {
+        writeTheme(dark ? "darkmode" : "lightmode");
+    }
+}
+
+// Only a theme the reader picked is stored. Following the browser's setting is
+// not a choice, so it is applied without being stored, which leaves a later
+// change of that setting free to take effect.
+const savedTheme = readTheme();
 
 if (savedTheme === "darkmode") {
     setDarkMode(true, false);
--- a/html/api.html
+++ b/html/api.html
@@ -29,7 +29,7 @@
 -->
 <meta http-equiv="Content-Security-Policy" content="
     default-src 'none';
-    script-src 'sha256-jG/elDqAeYYU5kU5fxHvBLgsUPuUbQEjzMVi0CC3Rz0=' https://unpkg.com/swagger-ui-dist@5.32.14/;
+    script-src 'unsafe-inline' https://unpkg.com/swagger-ui-dist@5.32.14/;
     style-src 'unsafe-inline' https://unpkg.com/swagger-ui-dist@5.32.14/;
     img-src 'self' data:;
     font-src 'self' data:;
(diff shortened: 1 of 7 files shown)


chrisgleissner:
I reviewed this and pushed one commit (35a2f15) to the branch, because two problems in the diff would have reached users. Details below so you can check my reasoning.

### Both pages stopped working when the browser blocks site data

`sessionStorage` throws a `SecurityError` when a browser is set to block on-device site data (Chrome "Block all cookies", Firefox with `dom.storage.enabled` false, a sandboxed frame). Neither page touched Web Storage before this PR, so that setting was harmless.

In `api.html` the throw happened at the top-level `sessionStorage.getItem("theme")`, which sits above the two button handlers and above `window.onload`. Neither got installed, so Swagger UI never started and the page stayed on its offline notice.

In `index.html` the throw happened inside `initDisplayMode()`, which runs before `await ensureLoggedIn()`. The left navigation stayed at `visibility: hidden` and no login or welcome screen appeared.

Measured in Chromium and Firefox with the storage accessor made to throw, before the fix:

```
index.html  {theme: '',         nav: 'hidden'}  pageerror: The operation is insecure.
api.html    {handler: false, onload: false}     pageerror: The operation is insecure.
```

Both reads and writes are now wrapped in try/catch. A browser that refuses to store the theme still shows it.

### index.html stored the browser's own setting as if it were a choice

`initDisplayMode()` fell through to `setDarkMode()` / `setLightMode()`, which use the default `save = true`. Loading the page therefore wrote the current `prefers-color-scheme` value into `sessionStorage`. `api.html` passes `save = false` on all three of its startup branches, so the two pages disagreed about what is worth storing.

The visible effect: load `index.html` while the browser is light, let the browser switch to dark, reload, and the page stays light. `api.html` then also shows light, because it reads the value `index.html` froze. Only a theme the reader picked should survive.

`initDisplayMode()` now calls `setTheme(..., false)` on both browser-setting branches. The buttons still save.

### The CSS selector change is not a Firefox fix

The commit message says "Replace WebKit-specific link selector with standard anchor selector". There is no WebKit-specific selector in either file. The change is `div#left-nav a:hover.darkmode` to `div#left-nav a.darkmode:hover`.

I loaded both forms in Firefox and Chromium. Both parse, both stay in `document.styleSheets[0].cssRules`, and both apply `burlywood` on `dimgray` when the link is hovered. The two selectors have the same specificity and the same meaning, because a compound selector may interleave classes and pseudo-classes in any order.

I left the reorder in, it reads better. But if you saw a real theme problem in Firefox, it is still there and nothing in this diff addresses it. What did it look like?

### Verification

32 checks against the two pages, run in Chromium and in Firefox, driving real clicks and real reloads: the browser setting is followed on first load and not stored; a change of that setting after a visit takes effect; a picked theme is stored, carries between `index.html` and `api.html` in both directions, and outlives a change of the browser setting; both pages stay usable with site data blocked. All 32 pass in both browsers after the fix. Before the fix, 7 fail in each browser.

The inline script of `api.html` changed, so its CSP digest is recomputed to `sha256-aqK8YQoFANsW5zAcGrfy1JSRJISHSHowXSVLT070ZNo=`. `tools/openapi/test_explorer.py` passes, 16 tests. To show that digest check is real rather than passing by default, I served the page with the previous digest: Chromium and Firefox both refuse to run the inline script, with no console error to say why, and the page loses its theme buttons and Swagger UI.

Nothing was flashed to a device. The pages were served over plain HTTP from a local directory, with `unpkg.com` blocked so the offline path was exercised too.

https://claude.ai/code/session_01R7YpUbzaH9SATN3KBxcVMh

chrisgleissner:
Two more commits on the branch.

**29fbef3d** shortens the comments I added to both pages.

**6a66c3d1** adds `tests/e2e/web/theme_test.py`, a manual E2E suite that drives both pages in an installed Chrome and Firefox through Selenium. It needs no device: it serves `html/` over a local HTTP server, or takes `--base-url` to test what a running device serves. 13 checks per browser, covering the two defects above and the CSP digest.

Every condition is a real browser preference rather than anything injected into the page. Blocked site data is `profile.default_content_setting_values.cookies = 2` in Chrome and `network.cookie.cookieBehavior = 2` in Firefox; both were verified to make `sessionStorage` throw a real `SecurityError`. The colour scheme is CDP `setEmulatedMedia` in Chrome and `ui.systemUsesDarkTheme` in Firefox. The clicks and the navigation are real WebDriver actions.

The suite discriminates:

- Against the pages as they were before 35a2f150: `[01] index.html follows a light browser and stores nothing ... FAIL (what is stored is 'lightmode', expected None)`, in both browsers.
- Against the fixed pages carrying the previous CSP digest: `[04] api.html follows a dark browser and stores nothing ... FAIL (the theme is 'lightmode', expected 'darkmode')`, in both browsers. That is the failure mode worth having a test for, because a refused inline script produces no console error.

Registered as `manual` because it needs installed browsers that CI does not have. A missing browser or WebDriver is one SKIP naming the reason, not a failure; if the host cannot reach the CDN `index.html` loads jQuery from, those checks skip and the `api.html` checks still run.

```
./run-tests -s web-theme    26 checks, 26 OK, 34.5s
```

`runner_policy_test`, `check_transport_usage`, `observability_test` and `tools/openapi/test_explorer.py` all still pass with the suite registered.

### One thing to decide

I checked the CSS selector change with real browsers, because the commit message calls it a Firefox fix. `div#left-nav a:hover.darkmode` and `div#left-nav a.darkmode:hover` behave identically in Chrome 152 and Firefox 154: both survive parsing, both apply burlywood on dimgray under a real hover, and they have the same specificity. The Selectors spec only requires the type selector to come first in a compound selector; classes and pseudo-classes may interleave. There is no linter rule for the order either.

So it is a readability preference, not a compatibility fix. I left it in because it changes nothing, but I am happy to revert that hunk if you would rather keep the diff to the two defects. And if you did see a real theme problem in Firefox, it is still there and nothing in this branch addresses it.

PR #833, opened 2026-09-03
CI only: rebuild PR 776 heap tracking

barryw:
Temporary diagnostic rebuild for PR #776 after the caller-list fix. Enables HEAP_TRACK only in the U64 CI command; this PR will be closed after its artifact is downloaded.


barryw:
Corrected HEAP_TRACK U64 artifact downloaded and verified for PR #776 hardware testing; temporary build PR no longer needed.

Issue #807, opened 2026-08-27
UCI network target: a UDP datagram larger than 892 bytes cannot be sent


chrisgleissner:
Follow-up to #802, which fixed the receive direction. This is the send direction, filed separately because #802 was about `READ_SOCKET` and this needs a new command rather than a fix to an existing one.

Reported by @JC-000 in https://github.com/GideonZ/1541ultimate/issues/802#issuecomment-5440267496, where the cause was also identified correctly. Everything below was measured on an Ultimate 64 Elite, firmware 3.15, FPGA 123, core 1.4E.

## The limit

`WRITE_SOCKET` sends whatever a single command carries, so the ceiling is the command queue rather than anything in `network_target.cc`. Measured against a peer on the same LAN, on the released 3.15 build:

| `WRITE_SOCKET` payload | bytes on the wire | 2-byte reply | status |
| ---: | ---: | ---: | --- |
| 892 | 892 | 892 | `00,OK` |
| 893 | 892 | 892 | `00,OK` |
| 1000 | 892 | 892 | `00,OK` |

892 is the fencepost in `command_protocol.vhd`: `command_pointer` stops incrementing at `c_cmd_if_command_buffer_end` while `command_length <= command_pointer - c_cmd_if_command_buffer_addr`. With a 896-byte buffer at offset 0 the end is 895, so the command length saturates at 895 and the payload at 895 - 3.

The write path does no length validation, so a larger request is truncated without a status. The 2-byte reply does carry the true send count, 892 for the 893-byte request above, so a client that reads it can detect this on released firmware today.

Two further measurements on the same device:

- UDP cannot be split on the client side. 500 + 500 sent as two `WRITE_SOCKET` commands arrives as two datagrams of 500, not one of 1000.
- TCP is unaffected. 500 + 500 arrives as one contiguous 1000-byte stream, byte for byte.

So the gap is specific to the datagram boundary. It is also the only place in the UCI where one logical unit has to fit in one command: `DOS_CMD_WRITE_DATA` writes into a file, which has no record boundary, so repeated commands are enough there.

## Why this is a firmware change and not an FPGA one

[...]

PR #112, opened 2019-09-28
U2p and u2 fix

markusC64:
Makes firmware ready for testing on U2+.
On U2, it is also ready for testing.

You may want to add
```
--- a/target/software/mb_ultimate/Makefile
+++ b/target/software/mb_ultimate/Makefile
@@ -91,6 +91,7 @@ SRCS_CC        =  small_printf.cc \
                        filetype_crt.cc \
                        filetype_sid.cc \
                        filetype_bin.cc \
+          filetype_cfg.cc \
             host_stream.cc \
             keyboard_vt100.cc \
             screen_vt100.cc \
@@ -128,6 +129,7 @@ SRCS_CC      =  small_printf.cc \
                        socket_dma.cc \
             home_directory.cc \
             reu_preloader.cc \
+            configio.cc \
             $(PRJ).cc

 #            sock_echo.c \
```

on the U2, but then the Application is too large for current flash configuration.

PR #835, opened 2026-09-03
Spike for #807: NET_CMD_WRITE_SOCKET_CHUNK ($16), a payload larger than one command as one datagram

JC-000:
## What this is

A spike for #807: `NET_CMD_WRITE_SOCKET_CHUNK` ($16), which assembles a payload
from several UCI commands and sends it to a socket as **one** datagram. Today
`NET_CMD_WRITE_SOCKET` can carry at most 892 bytes, so a larger datagram cannot
be sent at all — splitting it across two commands sends two datagrams, which a
peer expecting one rejects. Receiving already reaches 1472; this makes sending
match.

It is offered as a spike, not for merge as it stands, per @chrisgleissner's note
on #807. The specification it implements is included as
`doc/uci_network_chunked_write.md` so the design can be reviewed on its own.

## Three commits

| | |
| --- | --- |
| `30a88b4d` | the E2E tests, which fail on unpatched firmware |
| `8c922633` | the firmware that makes them pass |
| `82902682` | the specification, as `doc/uci_network_chunked_write.md` |

## Evidence

Bench: Ultimate 64 Elite, unique id `601A96`, firmware 3.15, FPGA 124, core 1.4F,
as `/v1/info` reports it. Both routes, REST and native (6502 agent). The suite
registers 20 scenarios run on both routes — 40 combinations, 20 of them of this
command.

| Firmware | Result |
| --- | --- |
| `ed8bc28d`, unpatched | 20 of 20 chunked combinations fail, every one citing `21,UNKNOWN COMMAND`; the 20 pre-existing combinations pass |
| `d33b7802` + this branch | 40 of 40 combinations pass, 197 checks |

The two rows come from different bases. The branch was rebased from `ed8bc28d`
onto `d33b7802` after the baseline run; the baseline row is a claim about the
absence of this command, and nothing in those seven commits adds it. The spike
row was measured on the base this branch now sits on.

> One caveat recorded rather than tidied away: nineteen of the twenty baseline
> pre-existing passes came from the suite run. The twentieth,
> `reset-closes-uci-sockets` on the native route, did not — `tools/64tass/64tass`
> was replaced with the tracked Linux binary while the suite was running, so the
> native route could not assemble its 6502 agent. It was re-run on its own,
> against the same unpatched firmware, and passed with 32 checks.

Baseline run 937 s; patched run 1323 s.

The two properties the command exists for, measured on the wire:

- a 1472-byte payload carried as 888 + 584 arrived as `datagrams [1472]` — one
  datagram — the completing chunk replying `c0 05`, being 1472 little-endian;
- a 893-byte payload, one byte past what `NET_CMD_WRITE_SOCKET` can carry,
  arrived as `datagrams [893]`, with nothing on the wire after the first chunk.

## Verified by an independent client, and adopted

`c64-wireguard` implemented this command in its own adapter — a second client on
unrelated code — and ran it on the same bench.

- Echo sweep 8/8: 888, 889, 891, 892, 893, 1452, 1472 bytes each left as exactly
  one wire datagram of the right length; 1473 refused before reaching the device.
- §2.1 asserted from a 6510-side debug-bus trace: a non-completing part leaves
  the response count at zero and the buffer untouched, answering `00,OK`; only
  the completing part writes the 2-byte total.
- Standard-MTU tunnel run, 48 MHz, peer left at its default 1420 MTU with no
  per-peer configuration: **60/60 checks, twice**. Outbound 888-1472 bytes each
  one datagram; inbound 892, 893, 1452, 1472 bytes received whole.
- Against **Cloudflare WARP's production WireGuard edge**: handshake in 48.5 s
  and 48.4 s, ICMP answered through the tunnel, and a 1278-byte DNS response
  received whole over the internet. Sizes above 1280 are capped by WARP's own
  MTU, not by this command.

That work is **merged** (`c64-wireguard` PR #112), so the first consumer has
shipped support rather than holding a branch.

## Byte weight

Measured baseline vs patched at `ed8bc28d` with `make u2_rv_swonly`:

| | Size | Free | Free % |
| --- | --- | --- | --- |
| `ed8bc28d` | 743.1 KiB | 48.9 KiB | 6.2 % |
| + this branch | 743.6 KiB | 48.4 KiB | 6.1 % |

**+504 bytes** on the U2 application image, of a 792 KiB partition. The U64 image
grows by 524 bytes and its size gate stays PASS at 20.5 % free.

## The one design decision worth arguing about

Each chunk carries an **explicit offset** rather than being positioned by an
accumulator the target keeps. It costs two bytes per chunk, taking the largest
chunk from 890 to 888. It buys the failure of two silent-corruption paths that
the implicit design has:

- A C64 reset does not reset the command interface. With an implicit accumulator
  a payload left in progress is silently continued by the next program's chunks,
  and the handle does not save you — lwip's `alloc_socket` is a first-fit from
  zero, so the low descriptors are reissued and a stale handle names a live
  socket.
- `NET_CMD_READ_SOCKET` between two chunks discards the accumulation, and every
  command still answers `00,OK`, so a client cannot see that it happened.

Appendix B of the specification sets both out with the code they rest on.

## Command number

`0x12` is not free. `xlar54/ultimateii-dos-lib` defines the TCP listener commands
at `0x12`-`0x15` and its `u-echoserver.c` sample sends `$03 $12`. This uses
`0x16`. See Appendix C.

## Not in this PR

While splitting `write_socket` I found that `NET_CMD_WRITE_SOCKET` has no
minimum-length guard: a two-byte command `$03 $11` gives `length = -1`, which
reaches `lwip_send`'s `size_t` as `0xFFFFFFFF`. It predates this work and is
reachable without ever sending `$16`, so it is filed separately rather than
bundled here.


JC-000:
**Rebased onto `883f608d`, re-measured, unchanged in content.** The code patch is byte-identical to what was reviewed; only Appendix D moved.

`uci-net-target` at `d33b7802`: 40/40 combinations, 197 checks. At `883f608d`: 40/40, 197 checks. Both UI transports, 0 failures, via `./run-tests --profile deep -s uci-net-target`. The bench moved FPGA 124 → 125 between the runs, so that is the same result on different gateware rather than a repeated one.

Separately, the `c64-wireguard` lane ran its live Cloudflare WARP interop test against this image (`a474a7ed`, FPGA 125): 4 handshakes reached ACTIVE, including 2 rekeys with `hs_timestamp` strictly increasing, plus ping, chat and a 1278 B / ancount 15 DNS reply through the tunnel — the chunked `$16` send path against a real peer, end to end, matching its reference on the previous image. Reproducible with seed `775774406`; PRG sha256 `d4dfb981…5d41e262` (msg_port 9999) and `1920eb23…b47ada9b` (msg_port 53), both `BACKEND=uci REU=0 UCI_CHUNKED_WRITE=1`.

Neither result alone separates a firmware fault from a WireGuard one; together they say the rebase changed nothing on either side. Both are claims of parity with the previous image, not a general clearance of `883f608d`.

Grrrolf:
> unique id `601A96`

Be careful with sharing this type of information. It could be considered PII when it can be linked to you as a person.

chrisgleissner:
Hi @JC-000 ,

just an FYI - the target branch for all future PRs should be `master`. More details in https://github.com/GideonZ/1541ultimate/pull/616#issuecomment-5667717676

I have already amended this PR here.

Thanks

JC-000:
Rgr, I saw the change but wasn't sure what your flow was so I didn't know if you were graduating the PR or I had missed the original target.

PR #838, opened 2026-09-03
Build 64tass on macOS, and rebuild it when the one in the tree cannot run here (#837)

### LINKED ISSUE #837: E2E: the new build-on-first-use of 64tass fails on macOS
# E2E: the new build-on-first-use of 64tass fails on macOS

**Introduced by:** #834 (`43a30ceb`), which made `tests/e2e/lib/assembler.py`
build the assembler on first use now that the binary is no longer committed.
**Affects:** any macOS checkout that has not built firmware — i.e. exactly the
case #834 added the build for.

## What happens

`ensure_assembler()` runs `make -C tools/64tass` with no flags. On macOS that
fails:

```
In file included from 64tass.c:36:
./wchar.h:46:12: error: conflicting types for 'wcwidth'
   46 | extern int wcwidth(uchar_t);
      |            ^
MacOSX.sdk/usr/include/_wchar.h:183:5: note: previous declaration is here
  183 | int     wcwidth(wchar_t);
1 error generated.
make: *** [64tass.o] Error 1
```

so `assemble()` raises and every suite that needs a 6502 stimulus — the UCI
native route among them — cannot run.

## Why

`tools/64tass/wchar.h:45-46`:

```c
#ifndef _XOPEN_SOURCE
extern int wcwidth(uchar_t);
```

The vendored header supplies its own `wcwidth` unless `_XOPEN_SOURCE` is
defined. The macOS SDK declares `wcwidth(wchar_t)` unconditionally, so the two
declarations conflict. On glibc the vendored declaration is compatible, which is
why Linux and CI are unaffected.

## Fix

Defining `_XOPEN_SOURCE` suppresses the vendored declaration and the build
succeeds:

```
make -C tools/64tass CFLAGS='-O2 -D_XOPEN_SOURCE=700'
```

Tested on macOS 15 (arm64, Apple clang) against `tools/64tass` at
`43a30ceb`: exit 0, and the resulting binary reports
`64tass Turbo Assembler Macro V1.53`. `-D_DARWIN_C_SOURCE` alone does **not**
work — I tried it separately and the conflict persists, so `_XOPEN_SOURCE` is
the load-bearing one.

Either adding it to `tools/64tass/Makefile` for Darwin, or passing it from
`ensure_assembler()`, would do. I have not proposed a patch because which of
those two you would prefer is a house-style question, and because the flag may
belong with the vendored source rather than with its caller.

## Note

Not urgent for CI, which is unaffected. It matters for contributors on macOS,
for whom the E2E suites went from "works from a fresh checkout" to "cannot
assemble" with #834.

JC-000:
Fixes #837.

Two commits: one makes 64tass build on macOS, one stops a binary built for
another platform being accepted and then failing later.

## 1. The build fails on macOS

Since the binary stopped being committed, 64tass is built from source on demand
— by `make -C tools` for a firmware build, and by the E2E suites for their 6502
stimulus. Both fail on macOS, as does building it directly:

```
./wchar.h:46:12: error: conflicting types for 'wcwidth'
   46 | extern int wcwidth(uchar_t);
MacOSX.sdk/usr/include/_wchar.h:183:5: note: previous declaration is here
  183 | int     wcwidth(wchar_t);
```

`wchar.h:45` declares `wcwidth` only `#ifndef _XOPEN_SOURCE`, and the macOS SDK
declares it unconditionally. Defining the macro is 64tass's own porting knob
rather than a workaround around one: `wchar.c:452-617` guards the bundled
implementation and its width tables the same way, so an XSI-conformant platform
links its libc's instead.

**Why in `tools/64tass/Makefile` rather than in the callers.** I first put it in
`tools/Makefile` and in the E2E helper, and that was wrong: it left
`make -C tools/64tass` still failing, which is both a reasonable thing to run
and what the helper's own error message tells you to try. Appending it in the
vendored Makefile fixes every caller from one place, and is a smaller change
than patching each of them. It is appended, so the existing `-O2` and
`-DREVISION` survive, and guarded on `uname`, so no other platform changes.

**One thing to weigh, since it is yours to decide rather than mine.** This is
the first local source change to vendored 64tass in the repository's history:
every previous touch has been a wholesale version drop (`34172882`, 163 files)
or the binary artifact. So the next "New version of 64tass" would drop these
three lines silently, and macOS would break again with nothing to say why. If
you would rather keep the vendored tree pristine, the same flag works from
`tools/Makefile` and the E2E helper instead — that survives a re-vendor, at the
cost of leaving a direct `make -C tools/64tass` broken. I have put it here
because it is the complete fix, but you do those updates and are better placed
to judge which trade you want.

**Why 700.** `isnprintf.c:62` reads `_XOPEN_SOURCE >= 500` as "the platform has
`snprintf`"; below that, a fallback replaces it and drops the size bound at its
one call site — on macOS masked only by the SDK having already defined
`snprintf`. `500` itself does not build, the SDK exposing `snprintf` only above
the level it selects. `600` and `700` give a byte-identical binary here; `700`
is the current POSIX level.

## 2. A binary from another platform is accepted, then fails

`ensure_assembler()` tested only whether the binary *existed*. A firmware build
under docker on a bind-mounted tree leaves a Linux binary and Linux objects
behind, so a later run on the host accepted them and failed in `assemble()` with
`Exec format error` — which says nothing about where the binary came from. The
reverse happens too: a host-built binary is not runnable inside the container.

It now tests whether the binary actually runs, and clears both it and the
objects before building, since `make` would otherwise keep the objects and
relink them into `unknown file type in 'str.o'`. On a fresh checkout there is
nothing to clear and this costs one no-op `make`.

Not macOS-specific — it is symmetric — but it is what makes the first fix hold
in a tree that has seen both platforms, which any macOS contributor who also
builds firmware will have.

## Against your four conditions

**Minimal.** Two files, and the first fix is three lines in one of them.

**Does not break Linux.** Guarded on `uname -s`, so no other platform's build
changes at all. Verified rather than assumed, from a fresh clone of this branch:

| | macOS 15, arm64 | `ghcr.io/gideonz/riscv:latest` |
| --- | --- | --- |
| `make -C tools/64tass` | builds | builds |
| `make -C tools` | builds | builds |
| the E2E helper, from clean | builds | builds |
| after the *other* platform built in the same tree | rebuilds, runs | rebuilds, runs |

All produce `64tass Turbo Assembler Macro V1.53.1515?`, so the revision string
survives everywhere. Before these changes the macOS column fails at `64tass.o`,
and the last row fails with `Exec format error` on both.

**Backwards compatible.** Nothing any non-Darwin platform executes has changed.

**No tests**, as you asked.

### DIFF
--- a/tests/e2e/lib/assembler.py
+++ b/tests/e2e/lib/assembler.py
@@ -33,16 +33,33 @@ ASSEMBLER = os.path.join(REPO_ROOT, "tools", "64tass", "64tass")
 ASSEMBLER_SOURCE_DIR = os.path.dirname(ASSEMBLER)
 ASSEMBLE_TIMEOUT_SECONDS = 120.0
 BUILD_TIMEOUT_SECONDS = 600.0
+VERSION_TIMEOUT_SECONDS = 30.0
+
+
+def assembler_runs() -> bool:
+    """Whether the binary at ASSEMBLER is one this machine can actually run.
+
+    Its presence is not enough. A firmware build under docker on a bind-mounted
+    tree leaves a Linux binary at that path, which the host then cannot execute,
+    and the failure surfaces from assemble() as `Exec format error` rather than
+    as anything about the build.
+    """
+    try:
+        return subprocess.run(
+            [ASSEMBLER, "--version"], capture_output=True,
+            timeout=VERSION_TIMEOUT_SECONDS).returncode == 0
+    except (OSError, subprocess.SubprocessError):
+        return False
 
 
 def ensure_assembler() -> None:
-    """Build `tools/64tass/64tass` from the source beside it when it is absent.
+    """Build `tools/64tass/64tass` when it is absent, or cannot run here.
 
     The same make that a firmware build runs, so the binary is the one the
     repository would have produced anyway, and a fresh checkout does not fail
     every assembling suite for want of a build step nobody was told about.
     """
-    if os.path.exists(ASSEMBLER):
+    if assembler_runs():
         return
     # A run naming several targets starts one process per target, and on a
     # fresh checkout two of them can reach an assembling suite together. The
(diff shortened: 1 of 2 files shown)


chrisgleissner:
Great work @JC-000 ! Thanks for that.

PR #840, opened 2026-09-03
Bound what a G64 tells the drive about its tracks (#823)

### LINKED ISSUE #823: G64 track metadata programs the drive's memory bounds unchecked
Found while sweeping for the transfer-bounds problems in #822, but a separate class: the 1541 emulation takes track lengths straight from a G64 file and uses them to program the hardware's memory bounds, with no validation.

**Why it matters.** `max_offset` is what the drive engine uses to bound its accesses, and it both reads and writes there: [`floppy_mem.vhd#L58-L82`](https://github.com/GideonZ/1541ultimate/blob/871ad0346f3e5ac259cd5947ca351ddb6b0f509e/fpga/1541/vhdl_source/floppy_mem.vhd#L58-L82) issues `mem_req.address <= unsigned(track_start) + offset_count;` with `read_writen` set both ways, wrapping at `offset_count >= unsigned(max_offset)`. `max_offset` is 14 bits ([`floppy_param_mem.vhd#L36`](https://github.com/GideonZ/1541ultimate/blob/871ad0346f3e5ac259cd5947ca351ddb6b0f509e/fpga/1541/vhdl_source/floppy_param_mem.vhd#L36)), so it accepts values up to 16383. Both buffers it can be pointed at are smaller than that, and neither has slack.

Four problems, all in the same path. The first needs only a malformed file.

| # | Problem | Trigger | Fix |
|---|---|---|---|
| 1 | Parameter RAM programmed from uninitialised locals | Any file that fails the G64 header check | Use an explicit safe length, as `remove_disk()` already does |
| 2 | Track length never validated against the buffer | A G64 declaring a length above 7928 | Clamp at load |
| 3 | Half-tracks get the previous track's length | Every G64, amplified by problem 2 | Same as problem 1 |
| 4 | Divide by zero on a zero-length track | A G64 with a non-zero offset and length 0 | Reject or skip length 0 |

---

## 1. Parameter RAM is programmed from uninitialised locals

**Problem.** In `C1541::insert_disk()`, `bit_time` and `track_len` are declared without initialisers and only assigned inside the `if` branch, but the `else` branch reads them: [`c1541.cc#L456-L500`](https://github.com/GideonZ/1541ultimate/blob/871ad0346f3e5ac259cd5947ca351ddb6b0f509e/software/drive/c1541.cc#L456-L500).

```c
uint32_t bit_time;
uint32_t track_len;
...
    if (tr->track_address) {
        bit_time = rotation_speed / tr->track_length;
        track_len = tr->track_length;
        *(param++) = (uint32_t)tr->track_address;
        *(param++) = (track_len-1) | (bit_time << 16);
    } else {
        *(param++) = (uint32_t)dummy_track;
        *(param++) = (track_len-1) | (bit_time << 16);
    }
```

If the first track has no address, both values are read before they are ever written.

**Trigger.** Trivially reachable, and it needs no crafted metadata. `GcrImage::load()` returns `false` on a bad signature after `invalidate()` has zeroed the whole track table ([`disk_image.cc#L697-L704`](https://github.com/GideonZ/1541ultimate/blob/871ad0346f3e5ac259cd5947ca351ddb6b0f509e/software/drive/disk_image.cc#L697-L704), [`disk_image.cc#L152-L159`](https://github.com/GideonZ/1541ultimate/blob/871ad0346f3e5ac259cd5947ca351ddb6b0f509e/software/drive/disk_image.cc#L152-L159)), but `mount_g64()` ignores the return value and inserts the image anyway: [`c1541.cc#L608-L624`](https://github.com/GideonZ/1541ultimate/blob/871ad0346f3e5ac259cd5947ca351ddb6b0f509e/software/drive/c1541.cc#L608-L624). So mounting any `.g64` whose first eight bytes are not `GCR-1541` or `GCR-1571` programs all 84 parameter slots with `dummy_track` and whatever those two stack slots happened to hold.

**Impact.** `dummy_track` is `new uint8_t[GCRIMAGE_DUMMYTRACKLEN]`, exactly 7692 bytes with no slack ([`c1541.cc#L218`](https://github.com/GideonZ/1541ultimate/blob/871ad0346f3e5ac259cd5947ca351ddb6b0f509e/software/drive/c1541.cc#L218)). A stale `track_len` above that makes the drive engine read and write past the end of the heap block.

**Fix.** The correct pattern is already in the same file. `remove_disk()` writes a hard-coded safe length for exactly this case: `*(param++) = (bit_time << 16) | 0x100;` at [`c1541.cc#L443-L448`](https://github.com/GideonZ/1541ultimate/blob/871ad0346f3e5ac259cd5947ca351ddb6b0f509e/software/drive/c1541.cc#L443-L448). Use the same constant in `insert_disk()`'s `else` branch, and initialise both locals. Separately, `mount_g64()` should check what `load()` returned and fall back to `blank()` or refuse the mount.

---

## 2. Track length is never validated against the buffer

**Problem.** `GcrImage::load()` takes the length from the image and masks it to 14 bits, with no upper bound check: [`disk_image.cc#L709-L727`](https://github.com/GideonZ/1541ultimate/blob/871ad0346f3e5ac259cd5947ca351ddb6b0f509e/software/drive/disk_image.cc#L709-L727).

```c
offset = le_to_cpu_32(pul[i+3]);
if(offset > GCRIMAGE_MAXSIZE) { ... return false; }
tr = gcr_data + offset;
if(offset) {
    w = tr[0] | (uint16_t(tr[1]) << 8);
    tracks[i].track_address = tr + 2;
    tracks[i].track_length = (int)(w & 0x3FFF);
```

`w & 0x3FFF` allows 16383. `GCRIMAGE_MAXTRACKLEN` is `0x1EF8`, that is 7928, so a declared length can be more than twice the largest legitimate track, and nothing compares it against `GCRIMAGE_MAXSIZE - offset - 2`. The buffer it points into is `new uint8_t[GCRIMAGE_MAXSIZE]` with no slack ([`disk_image.cc#L116`](https://github.com/GideonZ/1541ultimate/blob/871ad0346f3e5ac259cd5947ca351ddb6b0f509e/software/drive/disk_image.cc#L116)).

There is also an off-by-one on the offset check: `offset > GCRIMAGE_MAXSIZE` admits `offset == GCRIMAGE_MAXSIZE`, and `tr[0]` and `tr[1]` are then read one and two bytes past the end. And `offset` is never compared against `bytes_read`, so lengths can be taken from parts of the buffer the file never filled.

**Impact.** A track pointed near the end of `gcr_data` and declared 16383 bytes long makes the drive engine read and write up to about 8 KB past the allocation. Note `add_blank_tracks()` in the same file does bound-check correctly, `if (gcr + length <= address_limit)` at [`disk_image.cc#L182-L188`](https://github.com/GideonZ/1541ultimate/blob/871ad0346f3e5ac259cd5947ca351ddb6b0f509e/software/drive/disk_image.cc#L182-L188), so the missing check in `load()` is an omission rather than a design choice.

**Fix.** At load, clamp `track_length` to `GCRIMAGE_MAXTRACKLEN` and to the space actually remaining (`GCRIMAGE_MAXSIZE - offset - 2`), reject the track otherwise. Change the offset test to `>=`, and require `offset + 2 <= bytes_read`.

---

## 3. Half-tracks are given the previous track's length

**Problem.** `add_blank_tracks()` only fills even indices, `for(int i=0; i < GCRIMAGE_MAXUSEDTRACKS; i+=2)` ([`disk_image.cc#L174`](https://github.com/GideonZ/1541ultimate/blob/871ad0346f3e5ac259cd5947ca351ddb6b0f509e/software/drive/disk_image.cc#L174)). Every odd index therefore still has `track_address == NULL` when `insert_disk()` runs, and takes the `else` branch shown in problem 1, pairing the 7692-byte `dummy_track` with the length carried over from the preceding track.

**Impact.** For normal images this is safe by exactly zero margin: the largest entry in `track_lengths[]` is `0x1E0C`, which is `GCRIMAGE_DUMMYTRACKLEN` exactly. Combined with problem 2, a crafted image sets it to 16383.

**Fix.** Covered by the fix to problem 1.

---

## 4. Divide by zero on a zero-length track

**Problem.** `bit_time = rotation_speed / tr->track_length;` in `insert_disk()`. `track_address` is set to `tr + 2` whenever `offset != 0`, independently of the length, so a G64 declaring a non-zero offset and a length of 0 reaches the division with a zero divisor.

**Fix.** Treat a length of 0 as an unused track at load time.

---

## Reachability

`mount_g64()` is called from the file-type handler ([`c1541.cc#L1032`](https://github.com/GideonZ/1541ultimate/blob/871ad0346f3e5ac259cd5947ca351ddb6b0f509e/software/drive/c1541.cc#L1032)), so any route that puts a file on storage the machine can browse reaches this: USB stick, SD card, FTP upload, network share. No physical access is needed beyond getting the file onto the medium.

## How to test

**Host unit test, which is where the red/green belongs.** `GcrImage::load()` needs only a `File` that can return bytes, so it fits the existing host harness pattern under `software/io/usb/tests/` (Makefile compiles one source file plus `host_test_main.cpp`, wired into `make host_tests`). Feed it a corpus of crafted headers and assert, for every track: `track_length <= GCRIMAGE_MAXTRACKLEN`, `track_length > 0` or the track is unused, and `track_address + track_length <= gcr_data + GCRIMAGE_MAXSIZE`. Every problem above fails at least one of those assertions today.

**A second host test for `insert_disk()`'s parameter values.** The loop is pure arithmetic over `image->tracks[]`, so extracting the per-track computation into a small function lets a test assert that no slot is ever programmed with a length exceeding the buffer it points at, including the all-tracks-empty case from problem 1.

**On the device.** Mount a `.g64` with a deliberately wrong signature and confirm the drive still responds and the machine stays up. Then mount a well-formed G64 whose first track declares length 16383 and check the same. These are quick manual checks; the assertions above are what should run in CI.

## Suggested order

1. Problem 1. Reachable with a malformed file, and the fix already exists a few lines above in the same function.
2. Problem 2, including the off-by-one and the `bytes_read` check.
3. Problems 3 and 4 fall out of the first two.

enver-haase:
Fixes #823. Two commits, one per defect pair, each with the host tests that pin it down. No hardware was involved and none is needed: the checks are free functions over plain integers, and `make -C software/drive/tests` runs them in about a second.

## What was wrong

`insert_disk()` programs the drive engine's parameter RAM from the track table, and `floppy_mem.vhd` both reads and writes memory bounded by what it finds there. Everything in that table came from the file unchecked.

**Problems 2 and 4 — the length is never validated.** `w & 0x3FFF` admits 16383 while `GCRIMAGE_MAXTRACKLEN` is 7928, and nothing compared the length against the space left behind the track's own offset in a `new uint8_t[GCRIMAGE_MAXSIZE]` with no slack. Three smaller holes on the same path: `offset > GCRIMAGE_MAXSIZE` admitted `offset == GCRIMAGE_MAXSIZE`, so `tr[0]` and `tr[1]` were read past the buffer; `offset` was never compared against `bytes_read`, so a length could be taken from a part of the buffer no read had filled; and a track declared zero bytes long reached `rotation_speed / tr->track_length`.

**Problems 1 and 3 — the parameter pair is built from locals nobody wrote.** `bit_time` and `track_len` were declared without initialisers, assigned only in the branch for a track that is present, and read in the branch for one that is not. `mount_g64()` ignored what `load()` returned, so mounting any `.g64` whose signature is wrong programmed all 84 slots with `dummy_track` — 7692 bytes, no slack — and whatever those two stack slots held. And on *every* image the odd indices take the absent branch, because `add_blank_tracks()` fills only even ones, so half-tracks were handed `dummy_track` with the preceding track's length. That is safe today by exactly zero margin: the largest entry in `track_lengths[]` equals `GCRIMAGE_DUMMYTRACKLEN`.

## What this does

A new header, `software/drive/gcr_track_bounds.h`, holds three free functions over plain integers. They are free functions on purpose: none of the drive headers compile on a build host, and this is the part worth testing there. It is header-only, so no target Makefile changes.

- `gcr_track_header_is_readable()` — the two-byte length word has to lie inside what the file actually delivered.
- `gcr_validated_track_length()` — returns the length to use, or 0 when the track cannot be used. Rejects zero, anything above the format maximum, and anything that runs off the buffer.
- `gcr_track_parameters()` — builds one parameter pair. An absent track gets the dummy address with the same short, safe length `init()` and `remove_disk()` already program, and the function cannot see a neighbour at all.

A track that fails a check is left as `invalidate()` left it, so `add_blank_tracks()` gives it a blank track and the mount goes on rather than failing. `add_blank_tracks()` in the same file already bound-checks this way, so the missing check in `load()` was an omission rather than a design choice. `mount_g64()` now acts on `load()`'s answer: the file is closed and the drive is left empty, the state `remove_disk()` has already put it in.

## Red, then green

`software/drive/tests/gcr_track_bounds_test.cpp`, 18 cases, wired into CI next to the existing UCI palette test.

Against this branch: **18 passed.** Against a header reproducing the pre-fix behaviour of `027822696` — no length checks, `offset != 0` as the only header test, and the two parameter locals kept in statics so an absent track inherits its neighbour — **10 of the 18 fail**, covering all four problems: the 16383 length, tracks running off the buffer, a header straddling `bytes_read`, the zero-length divisor, and the absent track taking the neighbour's length and the wrong address.

The cases go past the four in the issue where the arithmetic invited it: a track ending exactly at the buffer end is accepted, one byte longer is refused, the MFM flag is not read as part of the length, and an offset of `0xFFFFFFFF` must not wrap.

## Built

`make u64ii_no_esp` in `ghcr.io/gideonz/riscv`, from a tree with every `output/` and `result/` under `target/` removed first. Zero errors, no new warning from `disk_image.cc` or `c1541.cc`, `update.ue2` and `update.cfw` produced.

## Not done here

The tests exercise the extracted functions, not `GcrImage::load()` end to end, because `disk_image.h` pulls in `userinterface.h`, `menu.h`, `subsys.h` and `iomap.h`. The manual device checks the issue suggests — mounting a `.g64` with a wrong signature, and one whose first track declares 16383 — have not been run: my C64U is bricked. Worth someone's five minutes on a live machine before this is trusted, and I would rather say so than imply otherwise.


chrisgleissner (changes requested):
Blocking: the PR confirms only that each two-byte track header was read, not that the declared track data was read. A 14-byte GCR-1541 file with track 0 offset 12 and declared length 0x1E0C passes the new header check and capacity check, then sends gcr_data+14 through gcr_data+7705 to the drive even though those bytes were never read and can be stale after a prior mount. Please require offset + 2 + length <= bytes_read without overflow, and add a red/green truncated-file regression. I am resolving the workflow-only merge conflict separately, retaining both the current HTTP integer test and this G64 test.

enver-haase:
You are right, and the hole is exactly where you put it: the check was against the
size of the buffer, so a track had to fit the allocation but not the file. Fixed in
`b1e60409`.

`gcr_validated_track_length()` now takes `bytes_read` in place of the capacity and
requires `offset + 2 + length <= bytes_read`, written as subtractions so no sum can
wrap, and `load()` passes what `f->read()` reported. Since `bytes_read` never exceeds
`GCRIMAGE_MAXSIZE`, this is strictly the stronger bound — the buffer cases the earlier
tests pin down still hold, unchanged.

Red and green, both measured:

```
against this commit                        23 passed
against the header restored to the old bound  4 failed
```

The four are the truncated cases, your 14-byte file among them, each returning 7692
where it has to return 0. The fifth new case is the positive boundary — a track ending
exactly where the file does — which must stay accepted either way, and does.

CI is green on the same commit, and it runs `make -C software/drive/tests` in the
container, so those cases ran there too and not only here.

Leaving the workflow-only conflict to you, as you proposed. Ready for another look.

software/drive/disk_image.cc
```diff
@@ -709,23 +710,32 @@ bool GcrImage :: load(File *f)
...
-    		tracks[i].track_used = true;
-    		tracks[i].in_image_file = true;
-    		tracks[i].track_is_mfm = (w & 0x8000);
-//            printf("Set track %d.%d to 0x%6x / 0x%4x.\n", (i>>1)+1, (i&1)?5:0, tracks[i].track_address, w);
-            if (i >= GCRIMAGE_FIRSTTRACKSIDE1) {
-                double_sided = true;
-            }
-        }
+    	w = tr[0] | (uint16_t(tr[1]) << 8);
+    	int length = gcr_validated_track_length(w, offset, bytes_read, GCRIMAGE_MAXTRACKLEN);
```
chrisgleissner, line 725: Blocking: this accepts every positive length, but `map_gcr_image_to_mfm()` later subtracts `MFM_TRACK_HEADER_SIZE` (162) from every accepted track into unsigned `reservedSpace`, and an MFM-marked track then reads up to that full 162-byte metadata area without checking `track_length`.

A concrete file can be exactly `GCRIMAGE_MAXSIZE` bytes, point track 0 at `GCRIMAGE_MAXSIZE - 3`, declare `0x8001`, and put sector count 32 in the final byte. This helper accepts the one delivered data byte; mounting then advances past the allocation and reads up to 160 more bytes in the sector loop. Without the MFM flag, `1 - 162` still wraps `reservedSpace` to a huge `uint32_t`, so a later MFM track update can pass the capacity check for data that does not fit.

Please make the mapping reject/skip MFM metadata unless the track contains at least `MFM_TRACK_HEADER_SIZE`, and avoid the unsigned subtraction for every shorter track (or reject such tracks at load if they are invalid for all supported uses). Add a regression with the `0x8001` end-of-buffer case so the downstream assumption is covered, not only the declared extent.

chrisgleissner (changes requested):
The new range check is necessary, but one downstream size assumption remains unchecked. I left one blocking inline finding: a short accepted track can still make the MFM mapping path read beyond `gcr_data` or underflow its writable-space bound.

enver-haase:
Confirmed, and closed in `ad56e8e8`. The finding was right; the mechanism sits one layer lower than either of us thought.

**The out-of-bounds read.** `map_gcr_image_to_mfm()` now refuses any track shorter than the metadata area, whatever its marker claims. Such a track is left exactly as `init()` zeroed it — no sectors, no reserved space — so nothing downstream reads a header that was never delivered. Your end-of-buffer file is pinned down directly in the tests.

**The underflow could not have happened, which is the more interesting half.** `MFM_TRACK_HEADER_SIZE` was defined without outer parentheses:

```c
#define MFM_TRACK_HEADER_SIZE   2 + (WD_MAX_SECTORS_PER_TRACK * 5)
```

so the one place that subtracts it, `gtr->track_length - MFM_TRACK_HEADER_SIZE`, expanded to `gtr->track_length - 2 + 160`. That is `track_length + 158`, not `track_length - 162`. `1 - 162` never wrapped because the subtraction was not one — instead **every mapped track reserved 320 bytes more than it had**, and `UpdateTrack()` admitted track writes that much too large. The macro's three other uses — an addition, a comparison, and a `- 2` — come out right either way, which is why this sat unnoticed.

Parenthesising alone would have created the underflow you predicted; the guard alone would have left the 320 bytes. Both are in.

The `static_assert` next to the macro is spelled `0 - MFM_TRACK_HEADER_SIZE == -162` deliberately: comparing the macro against 162 passes in either spelling, since `2 + 160` is 162, so it would not catch the brackets being lost again. `disk_image.h` now also includes `mfmdisk.h`, which defines the `WD_MAX_SECTORS_PER_TRACK` the macro had always used without including its source.

**All 29 tests pass on this commit.** To show that is not vacuous, I put each guard back the way it was, one at a time, and reran:

```
ad56e8e8, unmodified                        29 test(s) passed
reserved space back to a bare subtraction    3 test(s) failed
header check back to believing the marker    4 test(s) failed
```

In the first revert the one-byte track reserves 4294967135 bytes instead of 0; in the second, the reported file is mapped again. Both reverts were thrown away afterwards; they exist only to prove the new cases bite.

CI is green on this commit and ran `make -C software/drive/tests` in the container, so the six new cases ran there as well. The firmware builds too: `make -C target/u64ii/riscv/ultimate` in `ghcr.io/gideonz/riscv` on a native x86_64 host, exit 0, no errors and no warnings.

**Adjacent finding, reported separately as #921.** While measuring this I noticed that the WRITE TRACK handler discards `UpdateTrack()`'s rejection and then writes at a stale offset. Nothing in this PR touches `wd177x.cc`, so it is not a regression from it — but correcting the reservation here makes that rejection path genuinely reachable, which is why it seemed worth having on record now rather than later.

**Scope, so this is not mistaken for an endorsement.** Whether a G64 should be carrying MFM tracks at all is a separate question, now raised as #920: bit 15 of the length word is not part of the G64 specification, and VICE rejects such a track outright rather than loading it. This PR takes the mechanism exactly as it stands and only makes it safe to walk over.

For the record, my own view is that it should absolutely not stay. A file whose signature reads `GCR-1541` has no business carrying MFM content, and the gap it fills deserves a container of its own rather than a bit taken out of someone else's mandatory field. That argument belongs in #920 though, not in this review — nothing here depends on how it is settled.

Ready for another look.

PR #841, opened 2026-09-03
Fix U2+L boot hang: std::atomic emits FENCE, which the rvlite core traps

chrisgleissner:
## Relationship to 8a4e5771 and 883f608d

Those two commits address the same defect at its source, by making `FENCE` a NOP in the `rvlite` decoder. That is the better long-term fix, and this PR does not duplicate it. We reached the same conclusion independently from the hardware side, and it matches exactly: `decode_comb.vhd` left `when 3 => -- FENCE` commented out.

As of this writing that work does not yet produce a U2+L image. Both CI runs fail at "Build U2+L including FPGA":

```
ERROR: Found 3 i_riscv/* signals assigned to a PRIMARY net which for some
reason prevents booting, aborting. Please fix.
```

The `PROHIBIT` preferences added in 883f608d name synthesis-generated nets such as `un1_mstatus46_1_i`, `N_726_i` and `state_RNI52AF1[0]`. Diamond reports each of them as `does not match nets in the design. This preference has been disabled.`, because those names are regenerated by synthesis and are not stable between runs. Three `i_riscv/*` signals therefore still land on PRIMARY nets and the existing guard aborts the build. The last `test-merge` run that produced a U2+L package is 02782269, which predates the FENCE work.

So there is currently no U2+L bitstream carrying the gateware fix, and no way to test it on hardware.

This PR fixes the same failure in software instead. It matters for two reasons that remain true even after the gateware fix lands:

* It works with the bitstream that is **already on every cartridge in the field**. A gateware fix only helps a cartridge that has been updated, and a cartridge stuck in this state cannot run the updater at all, so it needs JTAG first.
* It is verified end to end on real hardware today, which the gateware fix cannot be until the ECP5 build succeeds.

If the gateware fix lands and ships, the `volatile bool` change here becomes redundant and can be dropped. It is not harmful in the meantime: these flags have one writer and no read-modify-write, so `volatile` is sufficient.

An earlier revision of this PR also added a link-time check rejecting `FENCE`. That has been removed, because it contradicts the direction of 8a4e5771.

## Problem

Since `c45dae01` ("fix(network): apply service switches live (#788)") the Ultimate II+L does not boot. The C64 shows a black screen with the cartridge fitted, and the reset, menu and freeze buttons do nothing. Recovery needs JTAG, because the U2+L has no golden FPGA partition.

The first bad build is `v3.15-104-gc45dae01`. The last good build is `v3.15-103-g9191d4c1`. Both were identified by flashing GitHub Actions update packages onto a real U2+L over ECP5 JTAG and reading the C64 reset line.

## Why the cartridge stops

The U2+L holds the C64 in reset until its application finishes starting. `InitFunction::executeAll()` runs before `c64 = C64::getMachine()` in `software/application/ultimate/ultimate.cc`, and `C64::C64()` is what writes `C64_MODE = MODE_NORMAL` to release the machine. Any init function that does not return therefore leaves the C64 in reset permanently.

Boot stops in the init function registered at ordering 100, `init_socket_gui` in `software/network/socket_gui.cc`.

## Root cause

`#788` changed the `enabled` flag in the four network daemons from a plain member to `std::atomic<bool>`:

* `software/network/socket_gui.h`
* `software/network/ftpd.h`
* `software/network/httpd.h`
* `software/network/socket_dma.h` (two flags)

The U2+L application is compiled with `-march=rv32i` (`target/u2plus_L/riscv/ultimate/Makefile`), which has no A extension. GCC therefore lowers each `std::atomic<bool>` access to an ordinary `lb` or `sb` bracketed by `FENCE` instructions. The `ultimate.elf` for `9191d4c1` contains 0 `FENCE` instructions. The one for `c45dae01` contains 22.

The Ultimate II+ and II+L run the `rvlite` core (`fpga/cpu_unit/rvlite`), instantiated as `rvlite_wrapper` in `fpga/fpga_top/ultimate_fpga/vhdl_source/u2p_riscv_lattice.vhd`. Its decoder does not implement `FENCE`. In `fpga/cpu_unit/rvlite/vhdl_source/decode_comb.vhd` the case arm is commented out:

```vhdl
-- when  3 => -- FENCE
--    null;
```

The opcode falls through to `when others`, which calls the `illegal` procedure because `g_trap_illegal` defaults to true, raising an illegal instruction trap with cause 2. `software/FreeRTOS/Source/portable/risc-v/port_asm.S` routes a synchronous non-`ecall` exception to `C_exception_handler`, which is:

```c
printf("\n** GURU MEDITATION: ...");
print_tasks();
while(1)
    ;
```

That loop runs with interrupts still enabled, so the FreeRTOS tick keeps running and other tasks keep being scheduled. Only the task that executed the `FENCE` is dead. That task is the one running `InitFunction::executeAll()`.

The first `FENCE` reached during boot is the atomic store in the `SocketGui` constructor:

```cpp
enabled = cfg->get_value(CFG_NETWORK_TELNET_SERVICE) != 0;
```

This is why the cartridge appears completely dead while the CPU is in fact still servicing interrupts.

U64 and U64-II are not affected, because they do not use the `rvlite` core. That is why the existing test suites, which run against U64 and C64U hardware over the network, stayed green. The failure is also invisible to any network based test by construction, because a cartridge in this state never brings its network up.

## Change

**Use `volatile bool` instead of `std::atomic<bool>` for the four service flags.** Each flag has a single writer and no read-modify-write, so atomicity is not required. `volatile` gives the visibility the listener tasks need and matches the idiom used elsewhere in this firmware. Each declaration carries a comment recording why `std::atomic` must not be reintroduced here.

That is the whole change. An earlier revision also added a link-time check that rejected `FENCE` for `rvlite` targets; it was removed once 8a4e5771 showed that upstream intends to support `FENCE` rather than reject it.

## Verification

All results are from one physical Ultimate II+L. The ECP5 bitstream at `0x000000` was held constant across every run, so the application at `0x0A0000` is the only variable. Each write was confirmed with `VERIFY OK` before the device was restarted.

Pass or fail was read from the C64 reset line by ECP5 boundary scan, not inferred from behaviour: `SLOT_RSTn` is ball `R3` (`target/fpga/u2plus_ecp5/u2p_ecp5.lpf`), cell 352 of the 409 bit boundary register on the LFE5U-25F, read with the read only `SAMPLE` instruction. `SLOT_RSTn = 1` means the application released the C64. The oracle was calibrated against two known states first: an unconfigured ECP5 reads 1 while the C64 visibly runs, and a known bad image reads 0.

| Build under test | `FENCE` count | `SLOT_RSTn` | Result |
| --- | --- | --- | --- |
| `9191d4c1`, last good | 0 | 1 | pass |
| `c45dae01`, first bad | 22 | 0 | fail |
| `c45dae01` with `lwipopts.h` reverted to `9191d4c1` | 22 | 0 | fail |
| `c45dae01` with the four daemon files reverted, `lwipopts.h` left new | 0 | 1 | pass |
| `c45dae01` with `PRIO_NETSERVICE` lowered from 1 to 0 | 22 | 0 | fail |
| `c45dae01` with this change applied | 0 | 1 | pass |
| `test-merge` tip with this change applied | 0 | 1 | pass |

The last good and first bad builds were each confirmed twice, and the fix was confirmed twice against the unmodified `c45dae01` build in the same session.

Two rows above exist to exclude specific alternative explanations. Reverting only `lwipopts.h` still fails, which rules out the `LWIP_NETIF_LOOPBACK`, `LWIP_LOOPBACK_MAX_PBUFS` and `SO_REUSE` changes in the same commit. Reverting only the four daemon files passes while `lwipopts.h` keeps its new values, which places the fault in the daemon changes. Lowering `PRIO_NETSERVICE` from 1 to 0 still fails, which rules out the listener tasks starving the lower priority main task.

The location was measured rather than assumed. The cartridge has no usable UART and its network never comes up in this state, so an instrumented build blinked the current init function ordering onto the C64 reset line, and the host decoded it by repeated boundary scan. That reported ordering 100 consistently. Statement level instrumentation of the `SocketGui` constructor then reported the marker immediately before the atomic store.

**Demonstrating that the change is what matters:** restoring `std::atomic<bool> enabled;` in `software/network/socket_gui.h`, with everything else unchanged, puts the `FENCE` instructions back and the cartridge stops booting again. Changing that single declaration back to `volatile bool` makes it boot.

### The fixed firmware running on hardware

The branch was built, flashed onto the U2+L over JTAG and cold booted. The
cartridge that would not start now reports:

```
{ "product" : "Ultimate II+L", "firmware_version" : "3.15",
  "fpga_version" : "124", "hostname" : "u2", ... }
```

The behaviour `#788` was written to provide was then checked on that device,
because a fix that stops the hang but breaks the feature would be no good.
Switching a service off and on again is applied to the running listeners:

| Check | Result |
| --- | --- |
| Telnet port 23 open while enabled | pass |
| Telnet port 23 closed after disabling at run time | pass |
| Telnet port 23 open again after re-enabling | pass |
| FTP port 21 open while enabled | pass |
| FTP port 21 closed after disabling at run time | pass |
| FTP port 21 open again after re-enabling | pass |

The e2e test added by `#788` itself, `tests/e2e/network/ident_service_switch_test.py`,
also passes against the U2+L:

```
[01] ident answers when enabled ... OK (0.068s)
[02] ident stops answering when disabled ... OK (1.294s)
[03] ident answers again when re-enabled ... OK (0.808s)
ident_service_switch_test: OK (3 checks, 2.329s)
```

The web service was deliberately left enabled throughout, because it carries
the REST calls that perform the switching. No setting was written to flash, so
the device kept its stored configuration.

## How this was investigated

The cartridge could not report anything about itself. It has no accessible
UART, `printf` output goes to a log nobody can read in this state, and the
network never comes up. Every conclusion below therefore came from the
hardware over ECP5 JTAG.

**Establishing a pass/fail signal.** The first step was an oracle that can
tell "the bootloader stopped" from "the application finished". `SLOT_RSTn`
does that, because the cartridge holds the C64 in reset until the application
completes startup. It was calibrated against two known states before any
result was believed.

**Finding the boundary.** Twenty-three `test-merge` update packages were
downloaded and extracted offline, then binary searched on the cartridge. Builds
whose head never landed on `test-merge` were removed with
`git merge-base --is-ancestor`, because a branch build listing includes PR head
builds whose `v3.15-NN` counts collide with real merges. The build found on the
cartridge was identified byte for byte by matching the application container at
`0x0A0000` against each extracted image, rather than inferred from dates.

**Reproducing locally.** A local build of `c45dae01` was flashed and failed,
and a local build of `9191d4c1` was flashed and passed. Only then were
experiments trusted, because the local toolchain had been shown to reproduce
both sides of the boundary.

**Splitting the commit.** The ECP5 bitstream was held constant for every run,
so the application was the only variable. Reverting only `lwipopts.h` still
failed and reverting only the four daemon files passed, which located the fault
without any guesswork about mechanism.

**Locating the hang.** An instrumented build blinked the current init function
ordering onto the C64 reset line, and the host decoded it by repeated boundary
scan, turning `SLOT_RSTn` into a one bit telemetry channel out of a cartridge
that can otherwise say nothing. That reported ordering 100. Statement level
instrumentation of the `SocketGui` constructor then narrowed it to the atomic
store.

**Identifying the mechanism.** Further instrumented builds showed that no lwIP
`LWIP_ASSERT` fired, and a FreeRTOS `configASSERT` was excluded because it calls
`portENTER_CRITICAL()` first, which would have stopped the beacon that was still
running. Lowering `PRIO_NETSERVICE` excluded priority starvation. Comparing the
two ELF files then showed 0 `FENCE` instructions in the good build and 22 in the
bad one, which led to the `rvlite` decoder.

Three hypotheses were tested and discarded along the way: the lwIP option
changes, priority starvation, and `std::atomic` being lowered to libatomic
calls. The last was excluded by disassembly, which showed plain `sb` and
`fence` rather than libcalls.

## Alternatives considered

**Implementing `FENCE` in the `rvlite` decoder.** `FENCE` can be a no-op on this single core CPU, so uncommenting the case arm is a small change. It was rejected as the fix for this defect because it requires a new FPGA bitstream, and every cartridge already in the field runs the current one. A software only fix repairs those devices; a gateware fix does not. Implementing `FENCE` in `rvlite` is still worth doing separately, and the build check added here stays useful either way, because it keeps the software honest about what the deployed cores actually execute.

**Keeping `std::atomic` and adding `-mno-fence` or similar.** No such option exists for RISC-V, and suppressing the fences while keeping the atomic type would silently change the memory model rather than state the intent.

## Limitations

This change does not add automated coverage that would have caught the defect. The existing suites run over the network against U64 and C64U hardware, and neither condition can detect a U2+L that never starts its network. The build check closes the specific hole, that an image containing `FENCE` can be produced and packaged for an `rvlite` product, but it does not verify that a U2+L boots. A boot check would need a U2+L in the test rig, with `SLOT_RSTn` boundary scan or an equivalent signal as the oracle.

The `FENCE` guard covers `target/u2/riscv/*` and `target/u2plus_L/riscv/*`. `u2plus` has no RISC-V target directory in this tree, so nothing was added for it.

### DIFF
--- a/software/network/ftpd.h
+++ b/software/network/ftpd.h
@@ -41,7 +41,6 @@
 
 #include "FreeRTOS.h"
 #include "task.h"
-#include <atomic>
 #include "config.h"
 //#include "semphr.h"
 #include "indexed_list.h"
@@ -52,7 +51,14 @@
 class FTPDaemon : public ConfigurableObject
 {
 	static void ftp_listen_task(void *a);
-	std::atomic<bool> enabled;
+// NOTE: this flag must not be std::atomic. The Ultimate II+ and II+L run the
+// rvlite core (fpga/cpu_unit/rvlite), whose decoder does not implement FENCE:
+// decode_comb.vhd leaves "when 3 => -- FENCE" commented out, so the opcode
+// falls through to the illegal-instruction trap. GCC brackets every
+// std::atomic access with FENCE on rv32i, so a single atomic access ends in
+// C_exception_handler's endless loop. A plain volatile bool is sufficient
+// here: the flag has one writer and no read-modify-write.
+	volatile bool enabled;
 public:
 	TaskHandle_t listenTaskHandle;
 
--- a/software/network/httpd.h
+++ b/software/network/httpd.h
@@ -8,7 +8,6 @@
 
 #include "FreeRTOS.h"
 #include "task.h"
-#include <atomic>
 #include "config.h"
 #include "indexed_list.h"
 
@@ -17,7 +16,14 @@
 class HTTPDaemon : public ConfigurableObject
 {
 	static void http_listen_task(void *a);
-	std::atomic<bool> enabled;
+// NOTE: this flag must not be std::atomic. The Ultimate II+ and II+L run the
+// rvlite core (fpga/cpu_unit/rvlite), whose decoder does not implement FENCE:
+// decode_comb.vhd leaves "when 3 => -- FENCE" commented out, so the opcode
+// falls through to the illegal-instruction trap. GCC brackets every
+// std::atomic access with FENCE on rv32i, so a single atomic access ends in
+// C_exception_handler's endless loop. A plain volatile bool is sufficient
+// here: the flag has one writer and no read-modify-write.
+	volatile bool enabled;
 public:
 	TaskHandle_t listenTaskHandle;
 
--- a/software/network/socket_dma.h
+++ b/software/network/socket_dma.h
@@ -11,7 +11,6 @@
 #include "menu.h"
 #include "filemanager.h"
 #include "subsys.h"
-#include <atomic>
 #include "config.h"
 
 #define SOCKET_BUFFER_SIZE 200000
@@ -24,8 +23,15 @@ class SocketDMA : public ConfigurableObject {
 	static int  writeSocket(int socket, void *buffer, int length);
 
 	uint8_t *load_buffer;
-	std::atomic<bool> dmaEnabled;
-	std::atomic<bool> identEnabled;
+// NOTE: this flag must not be std::atomic. The Ultimate II+ and II+L run the
+// rvlite core (fpga/cpu_unit/rvlite), whose decoder does not implement FENCE:
+// decode_comb.vhd leaves "when 3 => -- FENCE" commented out, so the opcode
+// falls through to the illegal-instruction trap. GCC brackets every
+// std::atomic access with FENCE on rv32i, so a single atomic access ends in
+// C_exception_handler's endless loop. A plain volatile bool is sufficient
+// here: the flag has one writer and no read-modify-write.
+	volatile bool dmaEnabled;
+	volatile bool identEnabled;
 public:
 	SocketDMA();
 	virtual ~SocketDMA();
--- a/software/network/socket_gui.h
+++ b/software/network/socket_gui.h
@@ -10,12 +10,18 @@
 
 #include "FreeRTOS.h"
 #include "task.h"
-#include <atomic>
 #include "config.h"
 
 class SocketGui : public ConfigurableObject
 {
-	std::atomic<bool> enabled;
+// NOTE: this flag must not be std::atomic. The Ultimate II+ and II+L run the
+// rvlite core (fpga/cpu_unit/rvlite), whose decoder does not implement FENCE:
+// decode_comb.vhd leaves "when 3 => -- FENCE" commented out, so the opcode
+// falls through to the illegal-instruction trap. GCC brackets every
+// std::atomic access with FENCE on rv32i, so a single atomic access ends in
+// C_exception_handler's endless loop. A plain volatile bool is sufficient
+// here: the flag has one writer and no read-modify-write.
+	volatile bool enabled;
 public:
 	TaskHandle_t listenTaskHandle;
 
--- a/target/common/rules.mk
+++ b/target/common/rules.mk
@@ -210,18 +210,6 @@ $(OUTPUT)/$(PRJ).shex: $(OUTPUT)/$(PRJ).out
 $(RESULT)/$(PRJ).elf: $(OUTPUT)/$(PRJ).out $(RESULT)
 	@echo Providing ELF.
 	@cp $(OUTPUT)/$(PRJ).out $(RESULT)/$(PRJ).elf
-	@if [ "$(FORBID_FENCE)" = "1" ]; then \
-	    if $(OBJDUMP) -d $(RESULT)/$(PRJ).elf | grep -qE '\bfence(\.i)?\b'; then \
-	        echo "ERROR: $(PRJ).elf contains FENCE instructions." >&2; \
-	        echo "       The rvlite core used by the Ultimate II/II+/II+L does not decode" >&2; \
-	        echo "       FENCE (fpga/cpu_unit/rvlite/vhdl_source/decode_comb.vhd); it traps as" >&2; \
-	        echo "       an illegal instruction and the firmware hangs before it releases the" >&2; \
-	        echo "       C64 from reset. std::atomic is the usual source on rv32i." >&2; \
-	        $(OBJDUMP) -d $(RESULT)/$(PRJ).elf | grep -E '\bfence(\.i)?\b' | head -5 >&2; \
-	        rm -f $(RESULT)/$(PRJ).elf; \
-	        exit 1; \
-	    fi; \
-	fi
 	
 $(OUTPUT)/$(PRJ).s00: $(RESULT)/$(PRJ).bin
 	@echo Make mem...
--- a/target/common/rules.mk
+++ b/target/common/rules.mk
@@ -210,6 +210,18 @@ $(OUTPUT)/$(PRJ).shex: $(OUTPUT)/$(PRJ).out
 $(RESULT)/$(PRJ).elf: $(OUTPUT)/$(PRJ).out $(RESULT)
 	@echo Providing ELF.
 	@cp $(OUTPUT)/$(PRJ).out $(RESULT)/$(PRJ).elf
+	@if [ "$(FORBID_FENCE)" = "1" ]; then \
+	    if $(OBJDUMP) -d $(RESULT)/$(PRJ).elf | grep -qE '\bfence(\.i)?\b'; then \
+	        echo "ERROR: $(PRJ).elf contains FENCE instructions." >&2; \
+	        echo "       The rvlite core used by the Ultimate II/II+/II+L does not decode" >&2; \
+	        echo "       FENCE (fpga/cpu_unit/rvlite/vhdl_source/decode_comb.vhd); it traps as" >&2; \
+	        echo "       an illegal instruction and the firmware hangs before it releases the" >&2; \
+	        echo "       C64 from reset. std::atomic is the usual source on rv32i." >&2; \
+	        $(OBJDUMP) -d $(RESULT)/$(PRJ).elf | grep -E '\bfence(\.i)?\b' | head -5 >&2; \
+	        rm -f $(RESULT)/$(PRJ).elf; \
+	        exit 1; \
+	    fi; \
+	fi
 	
 $(OUTPUT)/$(PRJ).s00: $(RESULT)/$(PRJ).bin
 	@echo Make mem...
--- a/target/u2/riscv/reverter/Makefile
+++ b/target/u2/riscv/reverter/Makefile
@@ -3,9 +3,6 @@ ELFTYPE      = elf32-littleriscv
 ARCHITECTURE = riscv
 # CPU architecture and ABI
 MARCH ?= rv32i
-# The rvlite core on this product does not decode FENCE; reject any image
-# that contains one (see target/common/rules.mk).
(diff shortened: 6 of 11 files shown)


GideonZ:
It makes perfect sense why I suspected a compiler bug! It is an incredibly common point of confusion.

The compiler is actually behaving correctly according to the RISC-V specification. When compiling for RV32IM (without the A extension), the C++ compiler cannot emit hardware atomic instructions (like `amoswap.w` or `lr.w`/`sc.w`) because it knows the target hardware lacks them. [1] 

To satisfy a sequentially consistent thread fence or load/store wrapper required by `std::atomic<bool>`, the compiler falls back to using the base FENCE instruction accompanied by normal loads and stores. [2, 3] 

## Why FENCE belongs in RV32IM
The confusion stems from assuming FENCE is exclusive to multi-core or atomic extensions. In the RISC-V architecture, the FENCE instruction is part of the RV32I Base Integer Instruction Set. [1] 

The architects placed it in the base set so that any software compiled for any RISC-V processor could safely include memory ordering boundaries—even if running on a single-core or in-order pipeline where it serves as a NOP. [1] 

## The correct fix
Because my core lacks an A extension, it does not need to implement complex memory ordering logic, cache invalidations, or hardware bus locking. However, to prevent the illegal instruction trap and make my core RV32I-compliant, I simply need to update the decoder such that the instruction becomes a `NOP`. [2] 

Once the decoder safely ignores FENCE instead of trapping, the `std::atomic<bool>` C++ code will run perfectly! [1] 

[1] [[https://docs.riscv.org](https://docs.riscv.org/reference/isa/v20260120/unpriv/rv32.html)](https://docs.riscv.org/reference/isa/v20260120/unpriv/rv32.html)
[2] [[https://groups.google.com](https://groups.google.com/a/groups.riscv.org/g/sw-dev/c/ypGP_tHB1yY)](https://groups.google.com/a/groups.riscv.org/g/sw-dev/c/ypGP_tHB1yY)
[3] [[https://gcc.gnu.org](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html)](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html)

chrisgleissner:
Closing this. The gateware fix in 8a4e5771, 883f608d and 8806a431 is confirmed working on a real Ultimate II+L, so the software change here is not needed.

This PR was opened before those commits existed. We had reached the same conclusion independently from the hardware side, and it matched exactly: `decode_comb.vhd` left `when 3 => -- FENCE` commented out, so `FENCE` fell through to the illegal instruction trap, and `C_exception_handler` ends in `while(1);` with interrupts still enabled. The task that executed the fence died silently while the tick kept running and other tasks kept being scheduled. On a U2+L that task is the one running `InitFunction::executeAll()`, which happens before `c64 = C64::getMachine()`, so `C64_MODE = MODE_NORMAL` was never reached and the C64 stayed in reset with a black screen.

## Verification of the fix

The cartridge that would not boot was flashed with the `test-merge` build at ac2fe909 and now starts normally. It reports `Ultimate II+L`, firmware `3.15`, `git_commit_hash ac2fe909`, and `fpga_version 125`, up from `124` on the build that failed.

What makes this a real test of the gateware rather than a coincidence is that the application did not change. Counting the `FENCE` encoding `0x0ff0000f` in the extracted `u2l-ultimate.app` of each update package:

| Package | `FENCE` words in the application |
| --- | --- |
| `v3.15-103-g9191d4c1`, last build that booted | 0 |
| `v3.15-104-gc45dae01`, first build that did not | 22 |
| `v3.15-114-gac2fe909`, current, boots | 22 |

The application still contains all 22 fences. The only thing that changed is the ECP5 bitstream, so decoding `FENCE` as a NOP is what repaired it.

The behaviour #788 was written to provide still works on that build. Telnet on port 23 and FTP on port 21 each stop accepting connections when the service is switched off through `PUT /v1/configs/Network Settings/...` and accept again when it is switched back on. The web service was left enabled throughout, because it carries those REST calls. Nothing was written to flash.

## On this PR

The `volatile bool` change is redundant now and should not be merged. The gateware fix addresses every `FENCE`, not only the ones `std::atomic` emits, which is the better place to fix it.

An earlier revision of this PR also added a link-time check that rejected `FENCE` for `rvlite` targets. That was removed once 8a4e5771 appeared, because rejecting `FENCE` is the opposite of the direction taken here. Mentioning it only so the commit history is not confusing.

## Two observations, offered as notes rather than requests

Nothing currently tests that a U2+L boots. This reached `test-merge` because the suites that run are network based, against U64 and C64U hardware, and a cartridge in this state never brings its network up, so no network test can observe it. Any future change that hangs an init function would brick U2+L cartridges again with CI green. The signal we used needs no network and no serial port: `SLOT_RSTn` is ECP5 ball `R3`, cell 352 of the 409 bit boundary register on the LFE5U-25F, read over JTAG with the read only `SAMPLE` instruction. It reads 1 once the application has released the C64, and 0 while the cartridge still holds it.

The second note is the pairing of bitstream and application. An application built after 8a4e5771 is only safe next to a bitstream built after it. A package carrying the newer application beside an older cached bitstream would reproduce this failure on a working cartridge, and recovering from it needs JTAG, since the U2+L has no fallback FPGA partition.

Thanks for the quick turnaround on the decoder fix.

Issue #822, opened 2026-09-01
Transfers that touch memory outside the caller's buffer: sweep following #821


chrisgleissner:
Sweep of `test-merge` (`871ad03`) following #821 and #803, looking for every place where a transfer touches memory outside the range the caller owns.

**The rule that keeps getting broken:** a USB DMA buffer must start on a 4-byte boundary and have `(-len) mod 4` spare bytes behind it. The USB engine accesses memory in whole 32-bit words in both directions, so a transfer of `len` bytes touches `4 * ceil(len / 4)` bytes. Nothing in `usb_base.cc` states this rule, and only `bulk_out` (with #821 applied) enforces it.

Seven problems below, most serious first. Each has a fix and a test.

| # | Problem | Live today? | Fix |
|---|---|---|---|
| 1 | USB Ethernet TX writes 4 bytes below the caller's buffer | **Yes**, any device with a USB Ethernet adapter | Build the header in a buffer the driver owns |
| 2 | Every USB IN transfer can write up to 3 bytes past its buffer | Latent. Nine call sites, each safe by a different accident | Bounce with slack, or fix the byte enables in RTL |
| 3 | `control_write()` has no bounce, unlike `bulk_out()` | Unproven, same condition as #803 | Copy the same fix from #821 |
| 4 | Ethernet TX reads past short packets and puts the bytes on the wire | **Yes**, every ARP request | Zero-pad the buffer, not just the length |
| 5 | `control_exchange()` can return while holding the mutex | Latent, no caller reaches it | Add the missing `xSemaphoreGive` |
| 6 | The packet BRAM write path discards byte enables | Real, but does not corrupt transmitted data | Needs a wider write port, not a one-line change |
| 7 | Four smaller latent items | No | See below |

---

## 1. USB Ethernet transmit writes below the caller's buffer

**Problem.** `UsbAx88772Driver::output_packet()` builds the adapter's 4-byte length header *in front of* the buffer it is given, then transmits from there: [`usb_ax88772.cc#L600-L614`](https://github.com/GideonZ/1541ultimate/blob/871ad0346f3e5ac259cd5947ca351ddb6b0f509e/software/io/usb/usb_ax88772.cc#L600-L614).

```c
uint8_t *size = buffer - 4;
size[0] = ...; size[1] = ...; size[2] = ...; size[3] = ...;
host->bulk_out(&bulk_out_pipe, size, pkt_len + 4);
```

[...]

PR #843, opened 2026-09-03
Keep USB and Ethernet transfers inside the buffers they were given (#822)

enver-haase:
Takes the two network defects left in #822 after Gideon's gateware fix closed problems 2, 3 and 6 (evidence in https://github.com/GideonZ/1541ultimate/issues/822#issuecomment-5530352476).

## Problem 1 — the AX88772 header lands outside the caller's buffer

`output_packet()` wrote the adapter's four byte length header at `buffer - 4`. That holds only for a buffer from this driver's receive pool, which keeps four bytes of headroom because the receive path hands the stack `usb_buffer + 4`. Neither buffer the stack supplies on transmit has any: `lwip_output_callback()` passes a `static uint8_t[1536]` for a chained pbuf, so the write lands in whatever precedes that array in `.bss`, and `pbuf->payload` for a single one, where `ETH_PAD_SIZE` is 0 and `PBUF_LINK_ENCAPSULATION_HLEN` is left at its default of 0. A plain CPU write, no DMA behaviour needed, landing wherever the link order puts it.

Header and frame are now assembled in a driver-owned buffer, reached through `BIT31` like the receive pool so it stays uncached on any target that turns the data cache on. `bulk_out()` holds the USB mutex until the transfer completes, so one buffer is enough.

## Problem 4 — short frames put neighbouring memory on the wire

`RMII_TX_LENGTH = (pkt_len < 60) ? 60 : pkt_len;` padded the number, not the data. The engine transmits what that register names, so a 42 byte ARP request read 18 bytes past the frame and sent them. Nothing is corrupted — transmit only reads — but the bytes leave the device, and with `MEM_LIBC_MALLOC` the pbufs come from `pvPortMalloc`, so it is the neighbouring heap block. Every ARP request does this.

Short frames are now copied into a driver-owned buffer and zero filled to 60 there. Frames of 60 or more still transmit from the caller's buffer, so the copy costs nothing on real traffic.

## The adjacent one

`RmiiInterface::~RmiiInterface` released `ram_buffer`, a `new[]` array, with plain `delete`. Undefined behaviour. Visible only because the padding change puts a correct `delete[]` on the next line. Separate commit; drop it if you would rather have it on its own.

## Tests

Host tests over both transmit decisions, wired into CI, at the lengths @chrisgleissner named in review: 0, 1, 1500, 1536 and 1537 bytes for the AX88772 block, where 1537 is one byte too many and has to be refused, and 42, 59 and 60 for the RMII path. Every case runs with eight sentinel bytes in front of the frame **and eight behind it** — a leak past the end is what problem 4 actually was — and asserts the four byte header, the payload byte for byte, that the padding is zero rather than the neighbour, and that the caller's buffer comes back exactly as it went in.

Red and green, both measured:

```
against this branch             12 passed
against the pre-fix behaviour   11 failed
```

The red half is `eth_tx_frame_legacy.h`, a model of what the two drivers did to memory before the fix — the AX88772 header at `frame - 4`, and a short frame whose length register was raised to 60 while the data was not. `make -C software/io/network/tests` runs both halves and fails the build if the red one ever passes, so a later change that guts the checks is caught rather than going quiet. It is a model of the defect, not of the device, and no firmware builds it.

The one test that stays green either way is the control case: a frame of 60 bytes or more goes out of the caller's buffer untouched in both versions, and must.

## Measured at the wire

A C64 Ultimate on a direct cable to a capturing host, its Ethernet given a static address whose gateway does not exist, so it repeats ARP requests: 42 bytes of ARP, padded to Ethernet's 60. Same machine, same cable, same analysis script; only the firmware changed between the two runs.

```
1.1.0 line, unfixed (f0ec31cc1)   24 ARP requests    0 zero-padded   24 carrying other bytes   14 distinct patterns
this branch         (6720894a4)   17 ARP requests   17 zero-padded    0 carrying other bytes    1 distinct pattern
```

The counts differ because the two capture windows caught different numbers of retries; what matters is that not one frame in the first run was zero and not one in the second was anything else.

What the unfixed device actually put on the wire, bytes 42..59 of three of those frames:

```
00 40 ec 01 00 38 2e 38 22 0a 7d 2c 0a 20 20 22 65 72    ".@...8.8".},.  "er"
00 40 ec 01 00 01 00 00 00 00 00 00 04 70 6f 6f 6c 03    ".@...........pool."
00 40 f4 00 00 b4 00 35 00 2a e1 9b 7f 78 01 00 00 01    ".@.....5.*..x...."
```

The first is a fragment of a REST response — `8.8"` is the tail of `8.8.8.8`, then `"},\n  "er` beginning `"errors"`. The second is a DNS query: `04 70 6f 6f 6c` is the length-prefixed label `pool`, from `pool.ntp.org`. The third carries `00 35`, port 53, and a UDP length of 42.

So this is not "undefined bytes" in the abstract. Every address resolution the device performs broadcasts eighteen bytes of whatever last used that heap — configuration API responses and DNS traffic among them. After the fix, all eighteen are zero:

```
00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
```

## What this costs

**The AX88772 fix adds a memcpy of up to 1518 bytes to every transmitted frame.** The single-pbuf path previously copied nothing. The alternative from the issue — give both stack buffers four bytes of headroom — costs nothing per packet, but it needs `PBUF_LINK_ENCAPSULATION_HLEN` set globally, which touches every pbuf including the RMII driver's, and it fails silently the moment any path allocates a pbuf without the reserve. I chose the robust one. **I have not measured the throughput difference**, and if you would rather have the headroom, say so and I will rewrite it that way.

`output_packet()` now returns `ERR_ARG` above 1536 bytes. Not reachable through lwIP at MTU 1500, but it is a new failure path.

## Problem 5 moved out

The USB mutex release is now #915, with a red/green measurement from the device. It is USB and not networking, and only travelled here because the same sweep found it. The commit is reverted in this branch rather than dropped, so this PR stays a fast-forward and the review on it keeps its anchors.

## Built

`make u64ii_no_esp` in `ghcr.io/gideonz/riscv`, from a tree with every `output/` and `result/` under `target/` removed first: 431 files compiled, exit 0, zero errors, no new warning from `usb_ax88772.cc` or `rmii_interface.cc`.


chrisgleissner:
Hi @enver-haase ,

any chance for you to add a unit or ideally E2E test? As it is, without any test, this PR is not merge-ready.

Thanks
Christian

chrisgleissner (changes requested):
Request changes for 3.15b: the build proves these files compile, but no regression test exercises any changed path. Please add a host test (and wire it into CI) for: (1) AX88772 frame assembly at 0, 1, 1500, 1536 and 1537 bytes, with canaries before and after the caller buffer; assert the four-byte length/complement header and exact copied payload; (2) RMII frames of 42, 59, and 60 bytes; assert the first two are zero-padded to 60 without modifying the caller buffer; and (3) control_exchange with a non-8-byte setup; assert its semaphore is released before it returns -10. Narrow pure helpers are acceptable if that is what makes these memory-safety contracts host-testable. A hardware AX88772 transfer/capture remains useful additional evidence, but the deterministic regression tests are required before this patch can be release-ready.

enver-haase:
Will look into it!

chrisgleissner:
Thanks @enver-haase . Once you got some tests added, thanks for also updating the PR here with red/green TDD results, demonstrating that the tests fail against the unpatched firmware and pass against the patched one.

enver-haase:
@chrisgleissner — all three items are answered, and problem 4 is now measured at the wire as well as on a build host.

**(1) and (2), the two transmit paths:** host tests at exactly the lengths you named, canaries on **both** sides of every buffer — a leak past the end is what problem 4 actually was — with the four-byte header and complements checked, payload compared byte for byte, padding asserted zero, and the caller's buffer asserted unchanged. Wired into CI. **12 passed** against the branch, **11 failed** against `eth_tx_frame_legacy.h`, a model of the pre-fix drivers. The twelfth is the control case: a 60 byte frame transmits from the caller's buffer in both versions and must stay green. `make` fails the build if the red half ever passes, so the checks cannot be quietly gutted later.

**The ARP capture you asked for.** A C64 Ultimate on a direct cable to a capturing host, its Ethernet given a static address whose gateway does not exist, so it repeats ARP requests: 42 bytes of ARP, padded to Ethernet's 60. Same machine, same cable, same analysis script; only the firmware changed between the two runs.

```
unfixed (f0ec31cc1)   24 requests    0 zero-padded   24 carrying other bytes   14 distinct patterns
fixed   (6720894a4)   17 requests   17 zero-padded    0 carrying other bytes    1 pattern
```

The counts differ because the two capture windows caught different numbers of retries. What matters is that not one frame in the first run was zero and not one in the second was anything else.

The leaked bytes are readable. Bytes 42..59 of three frames from the unfixed device:

```
00 40 ec 01 00 38 2e 38 22 0a 7d 2c 0a 20 20 22 65 72    ".@...8.8".},.  "er"
00 40 ec 01 00 01 00 00 00 00 00 00 04 70 6f 6f 6c 03    ".@...........pool."
00 40 f4 00 00 b4 00 35 00 2a e1 9b 7f 78 01 00 00 01    ".@.....5.*..x...."
```

The first is a fragment of a REST response — `8.8"` is the tail of `8.8.8.8`, then `"},\n  "er` beginning `"errors"`. The second is a DNS query: `04 70 6f 6f 6c` is the length-prefixed label `pool`, from `pool.ntp.org`. The third carries `00 35`, port 53, and a UDP length of 42. So this is not "undefined bytes" in the abstract: every address resolution broadcasts eighteen bytes of whatever last used that heap.

**(3) `control_exchange`:** moved to #915 rather than host-tested here. `usb_base.cc` reaches the Nano core through about fifty hardware symbols, and faking them to get it onto a build host fakes away the thing under test. It is measured on the device instead, red and green, with a debug probe that is not in either branch.

That measurement corrects something in this PR: **the wait is 25 seconds, not 5.** `xSemaphoreTake()` counts ticks and `configTICK_RATE_HZ` is 200, so the `5000` in that file is 25 s. Measured: 24997 ms unfixed, 0 ms fixed. The "five seconds" in the description above was mine and it was wrong.

**Built:** `make u64ii_no_esp` in `ghcr.io/gideonz/riscv`, from a tree with every `output/` and `result/` under `target/` removed first: 431 files compiled, exit 0, zero errors, no new warning from `usb_ax88772.cc` or `rmii_interface.cc`.

chrisgleissner:
Excellent work. Many thanks @enver-haase 

@GideonZ I think this is merge ready based on the analysis of Enver. Do you agree?

chrisgleissner (approved):
Re-reviewed the current head adversarially, including the new shared-buffer lifetime/serialization assumptions. I found no actionable issue.

The AX88772 block is bounds-checked before assembly, uses the existing uncached alias convention, and remains live through the synchronous `bulk_out()`. The RMII pad remains live until `RMII_TX_BUSY` clears, and lwIP serializes the link-output path, so the single pad buffer does not introduce a concurrent reuse window. The 0/1/1500/1536/1537 and 42/59/60 cases pass locally; the green suite also passes under ASan+UBSan with `-Werror`. CI is green and compiled both changed drivers across its target matrix. The on-wire red/green capture directly covers the RMII disclosure fix.

Approving.

PR #851, opened 2026-09-04
Fix HTTP body editing after deleting the selected node

barryw:
Deleting the selected HTTP body object, or its ancestor, leaves the next edit using freed storage. Keep the editing position at the surviving parent in `HttpBodySlot::remove`. Tests cover deleting the selection, an ancestor, an unrelated sibling and an array element.

Introduced after 3.14 with the HTTP target. Tested independently against `test-merge` at `8596576f`, in the upstream pipeline’s `my_docker_image` (host G++ 13.3.0):

```sh
make -B -C target/pc/linux/test_http_target test-body-removal
```

```text
Before: AddressSanitizer: heap-use-after-free; exit 2
After:  Body removal: 63 checks, 0 failures; exit 0
```

[Red/green evidence](https://github.com/GideonZ/1541ultimate/actions/runs/33932324740/job/101213300364); [final CI build](https://github.com/GideonZ/1541ultimate/actions/runs/33932682504) passed all five firmware targets and the application-space gate. The test target loads ASan before the CI image’s preloaded libraries. Hardware E2E was not run.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,29 +88,11 @@ jobs:
           -v /opt/build-tools:/mnt/build-tools:ro \
           -v $GITHUB_WORKSPACE:/mnt/project:rw my_docker_image /bin/bash -c "cd /mnt/project && make -C software/io/command_interface/tests"
 
-      - name: Verify http-body-removal red/green
+      - name: Test HTTP Body Removal
         run: |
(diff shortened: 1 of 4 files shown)


chrisgleissner:

barryw:
@chrisgleissner this can go with 3.16

PR #852, opened 2026-09-04
Preserve JSON parent links when replacing values

barryw:
Replacing an existing JSON integer, boolean or string loses its parent link, breaking subsequent HTTP body UP and REMOVE commands. Set the parent on replacement values in the three `JSON_Object::set` overloads.

Introduced after 3.14 with the HTTP target. Tested independently against `test-merge` at `8596576f`, in the upstream pipeline’s `my_docker_image` (host G++ 13.3.0):

```sh
make -C software/api/tests
```

```text
Before: 3 parent-link assertions fail; 1 test failed; exit 2
After:  12 validation tests and 27 state tests passed; exit 0
```

[Red/green evidence](https://github.com/GideonZ/1541ultimate/actions/runs/33931063372/job/101209573238); [final CI build](https://github.com/GideonZ/1541ultimate/actions/runs/33932631857) passed all five firmware targets and the application-space gate. Hardware E2E was not run.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,6 +17,12 @@ jobs:
           submodules: recursive
           fetch-depth: 0
 
+      - name: Test JSON Replacement
+        run: |
+          docker run --rm \
+          -v /opt/build-tools:/mnt/build-tools:ro \
+          -v $GITHUB_WORKSPACE:/mnt/project:rw my_docker_image /bin/bash -c "cd /mnt/project && make -C software/api/tests"
+
       - name: Generate Version String
         run: echo "GIT_VERSION=$(git describe --tags --always)" >>$GITHUB_ENV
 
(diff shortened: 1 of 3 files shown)


chrisgleissner:

barryw:
@chrisgleissner let's ship this for 3.16. it doesnt gate 3.15

chrisgleissner:
@barryw @GideonZ I suggest we add the fixes for the bugs that Barry identified to 3.15b so that user can benefit from them as soon as possible. 

I am talking about these 5 and very small PRs: #848, #851, #852, #853 and #854.

This in turn would mean that they could already be merged into `master` now.

Thoughts?

barryw:
I'm cool with these being rolled up into a follow-on fix release if @GideonZ is. I don't think any of them are absolutely critical, but they are nice correctness fixes.

GideonZ:
Yeah, we don't have to be over zealous to fix things overnight. It is great when we can, but it is no disaster if we don't. We should strike a good balance between releasing a new version for every fart and releasing once after implementinga a quadrillion new features.

barryw:
As a **customer** I love being surprised by big releases a few times a year **and** bug-fix rollups between them.. but, not my project, so not my rules. Put the "wow, cool!" features in the big bucket and the, "wow, thanks!" fixes in the smaller ones. None of my fixes gate anything, so they can go any time. Nobody is clamoring for them.

How we decide which-goes-where and what-gets-released-when is the hardest part of software development lol. Well, that and naming things.

PR #853, opened 2026-09-04
Decode negative HTTP body integers consistently

barryw:
Negative one-, two- and three-byte HTTP body integers decode as positive when plain `char` is unsigned. Use `int8_t` for the sign byte and multiplication instead of shifting a negative signed value.

Introduced after 3.14 with the HTTP target. Tests cover zero, maximum positive, minimum negative and -1 at all four wire widths. Tested independently against `test-merge` at `8596576f`, in the upstream pipeline’s `my_docker_image` (host G++ 13.3.0):

```sh
make -B -C target/pc/linux/test_http_target test-integer-widths
```

Same tests with unsigned `char`, changing only the firmware fix:

```text
Before: Integer widths: 99 checks, 6 failures; exit 2
After:  Integer widths: 99 checks, 0 failures; exit 0
```

[Red/green evidence](https://github.com/GideonZ/1541ultimate/actions/runs/33931063177/job/101209572784); [final CI build](https://github.com/GideonZ/1541ultimate/actions/runs/33932635667) passed all five firmware targets and the application-space gate. Hardware E2E was not run.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,34 +82,17 @@ jobs:
           -v /opt/build-tools:/mnt/build-tools:ro \
           -v $GITHUB_WORKSPACE:/mnt/project:rw my_docker_image /bin/bash -c "cd /mnt/project && make fpga_depends && make esp_depends"
 
+      - name: Test HTTP Signed Integers
+        run: |
+          docker run --rm \
+          -v /opt/build-tools:/mnt/build-tools:ro \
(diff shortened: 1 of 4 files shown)


chrisgleissner:

barryw:
@chrisgleissner I think this is a 3.16 bug. I just don't like leaving broken code lying around. We can leave it until after 3.15 ships.

PR #854, opened 2026-09-04
Format INT_MIN without signed overflow

barryw:
Formatting `INT_MIN` negates a value that cannot be represented as a positive `int`. Compute its magnitude with unsigned arithmetic in `_cvt`. The regression checks JSON rendering of both signed extremes, -1 and zero.

This is a pre-existing defect, also present in 3.14. Tested independently against `test-merge` at `8596576f`, in the upstream pipeline’s `my_docker_image` (host G++ 13.3.0):

```sh
make -C software/api/tests CXX="g++ -fsanitize=undefined -fno-sanitize-recover=all"
```

```text
Before: UBSan: negation of -2147483648 cannot be represented in type 'int'; exit 2
After:  12 validation tests and 27 state tests passed; exit 0
```

[Red/green evidence](https://github.com/GideonZ/1541ultimate/actions/runs/33931063507/job/101209573877); [final CI build](https://github.com/GideonZ/1541ultimate/actions/runs/33932634064) passed all five firmware targets and the application-space gate. Hardware E2E was not run.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,6 +76,12 @@ jobs:
            PYTHONPATH=/tmp/openapi-packages python -m openapi_spec_validator \
              --validation-errors all \$(python tools/openapi/generate.py paths)"
 
+      - name: Test INT_MIN Formatting
+        run: |
+          docker run --rm \
+          -v /opt/build-tools:/mnt/build-tools:ro \
+          -v $GITHUB_WORKSPACE:/mnt/project:rw my_docker_image /bin/bash -c "cd /mnt/project && make -C software/api/tests CXX='g++ -fsanitize=undefined -fno-sanitize-recover=all'"
+
       - name: Determine FPGA & ESP32 dependencies
         run: |
           docker run --rm \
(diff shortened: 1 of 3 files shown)


chrisgleissner:
Hi @barryw ,

Thanks for finding this. 

Did you investigate where does this defect has impact on the user experience?  And does it only affect `-2147483648` or also other values?

As it is so low-level, it may be hard to tell, but maybe you could dig into the transitive callers of the affected code. It would help to triage this better for inclusion in 3.15 despite the change freeze for that release.

Thanks
Christian

barryw:
@chrisgleissner These probably aren't good candidates for 3.15, but I wanted to put a fix in for 3.16. As of right now I don't see anything that's gating 3.15, but that's where these bugs came from. I'm extending the e2e tests to focus on things that have changed between 3.14 and 3.15 so that we can find regressions that may gate 3.15

chrisgleissner:
Thanks for clarifying @barryw . I agree, let's merge all of these 4 most recent bug fixes once the merges for 3.15 are done. Also thanks for strengthening the E2E test. You may notice I did quite a lot of cleanup yesterday - a deep dive review and fix, ensuring we are using modern Python best practices, e.g. `set` instead of `Set` or `int | None` instead of `Optional`.

Currently running all E2E tests against all devices - including the C64U - and there is PR #849  open for this that will contain further test adjustments to get them to pass against the C64U. Hopefully done later today. I discussed this with Gideon and it is expected to be the last one that lands for 3.15 as it is not about prd code changes, only test adjustments to cater for all devices (u64, u2, c64u).

PR #858, opened 2026-09-05
Fix heap leak in JSON discovery replies

barryw:
Release each UDP JSON discovery reply after sending it, and give the reply ownership of its password boolean. This fixes an existing leak of about 1.26 KB per discovery request; its role in #856 is not yet established.

Linux hardware regression, same test before and after: 20 replies retained 25,208 bytes before; zero afterward. Password-protected replies and caller-tracking checks also pass. Local checks with pipeline tooling passed on Linux: lint, registry checks and the U64 firmware build. Hardware used f3935d44 with FPGA124 plus allocation tracking and this fix.

### DIFF
--- a/software/network/socket_dma.cc
+++ b/software/network/socket_dma.cc
@@ -601,7 +601,6 @@ void SocketDMA::identThread(void *a)
 
             if (strncmp(client_message, "json", 4) == 0) {
                 client_message[36] = 0;
-                JSON_Bool password_protected(*password ? 1 : 0);
                 JSON *obj = JSON::Obj()
                     ->add("product", product)
                     ->add("firmware_version", APPL_VERSION_ASCII)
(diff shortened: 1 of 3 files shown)


chrisgleissner:
@barryw Nice one!

software/network/socket_dma.cc
```diff
@@ -616,7 +615,7 @@ void SocketDMA::identThread(void *a)
                 // Current Assembly64 ignores responses with booleans so we only send this if passwords
                 // are active, in which case Assembly won't work anyway.
                 if (*password) {
-                    ((JSON_Object *)obj)->add("password_protected", &password_protected);
+                    ((JSON_Object *)obj)->add("password_protected", true);
```
GideonZ, line 618: I don't understand this change. Ah... If *password is non-zero, yes, it's always true.

PR #859, opened 2026-09-05
Free queued filesystem events when observers unsubscribe

barryw:
Release pending filesystem events when an observer unsubscribes. Closing the FTP New Host form previously left two notifications and their strings allocated.

Linux hardware regression: eight host recreations retained 1,128 bytes before the fix and zero afterward. The test uses the existing FTP workflow and heap-slope helper. The unchanged filesystem-refresh suite also passes all 38 checks, including paste and bulk delete; discovery remains leak-free. Local checks with pipeline tooling passed on Linux: lint, registry checks and the U64 firmware build. Hardware used f3935d44/FPGA124 with allocation tracking and the separate discovery fix.

Refs #856.

### DIFF
--- a/software/filemanager/filemanager.h
+++ b/software/filemanager/filemanager.h
@@ -276,6 +276,11 @@ public:
     }
     void deregisterObserver(ObserverQueue *q) {
     	observers.remove(q);
+        // Unsubscribed observers will never consume their pending events.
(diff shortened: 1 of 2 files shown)


chrisgleissner:
Great find @barryw , thank for raising this.

PR #861, opened 2026-09-06
Correct DOS_CMD_COPY_FILE in the Ultimate DOS documents

cybersorcerer:
The readthedocs page for the Ultimate DOS target described DOS_CMD_COPY_FILE
as copying "the file specified by <source> to the file specified by
<destination>". GideonZ/1541u-documentation#34 corrects that; this does the
same to the Word originals here, so the two do not drift apart.

The firmware builds the call as:

    res = fm->fcopy(path->get_path(), filename, destination, filename, false);

against `software/filemanager/filemanager.h`:

    FRESULT fcopy(const char *path, const char *filename, const char *dest,
                  const char *dest_filename, bool overwrite);

So `<destination>` arrives as `dest`, the destination *directory*.
`dest_filename` is `filename` a second time, so the copy always keeps the
source's own name and the command cannot rename. `overwrite` is `false`,
which becomes `FA_CREATE_NEW`, so an existing destination file is refused
rather than overwritten.

Following the page as written cannot succeed, and neither failure points at
the real mistake: a `<destination>` given as a not-yet-existing filename makes
`dp->cd(dest)` enter a directory that is not there and the status channel
answers `PATH DOESN'T EXIST`, which reads as if the *source* were wrong; and
once such a copy has succeeded once, repeating it answers `FILE EXISTS` for a
name that never appeared in any listing.

Verified on a C64 Ultimate reporting `ULTIMATE-II DOS V1.2` via
DOS_CMD_IDENTIFY: with `<destination>` as a directory the copy returns
`00,OK` and the file appears under the source's name.

Changed in `doc/ultimate_dos-1.1.docx` and `doc/ultimate_dos-1.2.docx` — the
command was introduced in 1.1, and 1.0 does not describe it. One sentence is
corrected and two are added, in the paragraph that follows.

**The PDFs are not regenerated.** `ultimate_dos-1.1.pdf` and
`ultimate_dos-1.2.pdf` are built from these documents and still carry the old
wording. Rebuilding them here would have produced a different layout on every
page from a different Word installation, which would bury a one-sentence
change under an 800 KB binary diff. They are better regenerated with whatever
produced the originals.


chrisgleissner:
@GideonZ Having Word documents (or any other binary format) in Github presents challenges in terms of versioning and visibility. I cannot see the differences that were applied without opening and comparing the Word documents visually.

I would suggest to maintain the spec as Markdown going forward.

What are your thoughts on that?

PR #867, opened 2026-09-07
Fix Fast Reset on the Ultimate 64

### LINKED ISSUE #866: Fast Reset has no effect on the Ultimate 64: the KERNAL ROM aperture does not read back
## Summary

"Fast Reset" in *C64 and Cartridge Settings* has no effect on an Ultimate 64. The setting is
stored and reported back correctly, and a reboot applies it, but the machine still takes the full
KERNAL RAM test time to reach the READY prompt.

Measured on `test-merge` at `4acc148c`, which is the branch the 3.15 release is cut from.

I am already investigating this and will raise a PR once the fix is confirmed.

## Measured

Ultimate 64 Elite I, firmware built from `test-merge` `4acc148c`, FPGA 125, core 1.4F.

Reset to READY, five samples per state. The screen is blanked before each reset so a READY left
by the previous boot cannot match, and the machine is rebooted after each config change because
the setting is applied when the system ROMs load.

| Fast Reset | Samples | Median |
| --- | --- | --- |
| Enabled | 2.39s, 2.41s, 2.39s, 2.41s, 2.40s | 2.40s |
| Disabled | 2.43s, 2.42s, 2.41s, 2.39s, 2.42s | 2.42s |

Enabling the setting makes no difference at all.

## Root cause

`C64::init_system_roms()` in `software/io/c64/c64.cc` loads the KERNAL file straight into the
FPGA ROM image aperture at `U64_KERNAL_BASE`, and then reads that aperture back to decide
whether to apply the Fast Reset patch:

```c
FRESULT fres = FileManager::getFileManager()->load_file(
    ROMS_DIRECTORY, cfg->get_string(CFG_C64_KERNFILE), (uint8_t *)U64_KERNAL_BASE, 8192, NULL);
if (fres != FR_OK) {
    memcpy((void *)U64_KERNAL_BASE, (void *)_default_kernal_65_start, 8192);
} else if (cfg->get_value(CFG_C64_FASTRESET)) {
    unsigned char *kernal = (unsigned char *)U64_KERNAL_BASE;
    if (!memcmp((void *)(kernal + 0x1d6c), (void *)fastresetOrg, sizeof(fastresetOrg))) {
        memcpy((void *)(kernal + 0x1d6c), (void *)fastresetPatch, 22);
    }
}
```

That aperture is write only on this hardware. A read of it returns zero, so the `memcmp` never
matches and the patch is never applied. The KERNAL itself still reaches the machine, because
loading it only uses the write path.

I proved this with a probe built into a firmware image and run on the device. The probe loads
the same KERNAL file a second time into a heap buffer, then reads the aperture back two ways and
compares both against that buffer. Output over syslog, with `Fast Reset` set to Enabled and
`Kernal ROM` set to `kernal.901227-03.bin`:

```
FRPROBE fres=0 r2=0 cfgfast=1
FRPROBE ref@1d6c E6 C2 B1 C1 AA A9 55 91
FRPROBE blk@1d6c 00 00 00 00 00 00 00 00  nonzero=0 diff=7993
FRPROBE byt@1d6c 00 00 00 00 00 00 00 00  nonzero=0 diff=7993
FRPROBE memcmp_aperture_vs_org=-230 memcmp_ref_vs_org=0
FRPROBE afterwrite blk 00 00 00 00  byt 00 00 00 00
FRPROBE afterrestore byt 00 00 00 00
```

Reading it:

- `fres=0` and `cfgfast=1`: the KERNAL file loaded successfully and Fast Reset is enabled, so
  the `else if` branch above is the one that runs.
- `ref@1d6c` holds the unpatched RAMTAS bytes, so the file on the device is a stock KERNAL and
  `memcmp_ref_vs_org=0` confirms the patch would apply if the comparison saw the real image.
- `blk` is the aperture read with `memcpy` and `byt` is the same aperture read byte by byte
  through a `volatile uint8_t *`. Both are zero for all 8192 bytes (`nonzero=0`), so this is not
  a block-transfer problem: the aperture does not read back at all.
- `memcmp_aperture_vs_org=-230` is the comparison the firmware actually makes. It does not
  match, so the patch is skipped without any message.
- `afterwrite` writes the 22 patch bytes into the aperture and reads them straight back. Still
  zero, which rules out a coherency delay.

There is already a workaround for this same aperture elsewhere in the tree:
`software/monitor/u64_memory_backend.cc` snapshots the BASIC and KERNAL apertures and then falls
back to reloading the files when `rom_is_all_zero()` says the snapshot came back empty.

The second call site, `C64::enable_kernal(uint8_t *rom)`, has the same `memcmp`/`memcpy` shape
but is not affected: its argument is main memory (a heap buffer, or the cartridge ROM area in
SDRAM), not the aperture, and it copies to the aperture afterwards. The non-U64 branch of
`init_system_roms()` is not affected either, for the same reason.

## How to reproduce

1. On an Ultimate 64, set *C64 and Cartridge Settings* / *Fast Reset* to Disabled.
2. Reboot the machine, then blank screen RAM at $0400 and reset. Time how long the READY prompt
   takes to appear.
3. Set *Fast Reset* to Enabled, reboot again, and repeat the measurement.

Both measurements come out at about 2.3 seconds. On a machine where the feature works, the second
one is about 0.2 seconds.

chrisgleissner:
Fixes #866.

## Previous behaviour

On an Ultimate 64, *C64 and Cartridge Settings* / *Fast Reset* had no effect. The setting was
stored and reported back correctly and a reboot applied it, but the machine still took the full
KERNAL RAM test time to reach the BASIC prompt.

Measured on an Ultimate 64 Elite I running this branch with `software/io/c64/c64.cc` reverted to
`test-merge` at `4acc148c`. Reset to READY, five samples per state, the screen blanked before each
reset so a prompt left by the previous boot could not match: a median of 2.40s with Fast Reset
Enabled against 2.42s with it Disabled.

## The defect

`C64::init_system_roms()` loads the KERNAL file straight into the FPGA ROM image aperture at
`U64_KERNAL_BASE`, and then reads that same aperture back to decide whether the image it just
loaded is a stock KERNAL that the Fast Reset patch applies to:

```c
FRESULT fres = load_file(ROMS_DIRECTORY, cfg->get_string(CFG_C64_KERNFILE),
                         (uint8_t *)U64_KERNAL_BASE, 8192, NULL);
if (fres != FR_OK) {
    memcpy((void *)U64_KERNAL_BASE, (void *)_default_kernal_65_start, 8192);
} else if (cfg->get_value(CFG_C64_FASTRESET)) {
    unsigned char *kernal = (unsigned char *)U64_KERNAL_BASE;
    if (!memcmp((void *)(kernal + 0x1d6c), (void *)fastresetOrg, sizeof(fastresetOrg))) {
        memcpy((void *)(kernal + 0x1d6c), (void *)fastresetPatch, 22);
    }
}
```

That aperture is write only on this hardware. The `memcmp` reads zeros, never matches, and the
patch is skipped without any message. The KERNAL itself still reaches the machine, because
loading it only uses the write path.

I did not take that from the code. I built a probe into a firmware image and ran it on the
device. The probe loads the same KERNAL file a second time into a heap buffer, reads the
aperture back two ways, and compares both against that buffer. With `Fast Reset` set to Enabled
and `Kernal ROM` set to `kernal.901227-03.bin`, over syslog:

```
FRPROBE fres=0 r2=0 cfgfast=1
FRPROBE ref@1d6c E6 C2 B1 C1 AA A9 55 91
FRPROBE blk@1d6c 00 00 00 00 00 00 00 00  nonzero=0 diff=7993
FRPROBE byt@1d6c 00 00 00 00 00 00 00 00  nonzero=0 diff=7993
FRPROBE memcmp_aperture_vs_org=-230 memcmp_ref_vs_org=0
FRPROBE afterwrite blk 00 00 00 00  byt 00 00 00 00
FRPROBE afterrestore byt 00 00 00 00
```

`fres=0` and `cfgfast=1` say the file loaded and the setting is on, so the `else if` branch is
the one that ran. `ref` is the file in heap memory and holds the unpatched RAMTAS bytes, and
`memcmp_ref_vs_org=0` confirms the patch would apply if the comparison saw the real image.
`blk` is the aperture read with `memcpy` and `byt` is the same aperture read byte by byte
through a `volatile uint8_t *`. Both are zero across all 8192 bytes, so this is not a
block-transfer problem: the aperture does not read back at all. `afterwrite` writes the 22 patch
bytes and reads them straight back, still zero, which rules out a coherency delay.

The same aperture behaviour is already worked around elsewhere in the tree:
`software/monitor/u64_memory_backend.cc` snapshots the BASIC and KERNAL apertures and falls back
to reloading the files when `rom_is_all_zero()` says the snapshot came back empty.

## The change

`init_system_roms()` now stages the KERNAL image in a buffer in main memory. The file is loaded
into that buffer, the Fast Reset comparison and patch happen there, and the finished image is
written to the aperture in one block copy. No read of the aperture is left on this path.

One behaviour change comes with it, stated separately because it is not just a move:

- A KERNAL file whose length is not exactly 8192 bytes is rejected and the built-in default is
  used. `load_file` answers `FR_OK` for a short file and reports the length it transferred, so
  the length is now checked. A short image has no reset vector at `$FFFC` and the machine could
  not start from it, so there is nothing to gain by keeping it, and rejecting it leaves no
  uninitialised bytes in the staging buffer to write into the aperture.

The Fast Reset patch stays in the `else` of the file-load test, so it is considered only for an
image that came from a file. The `memcmp` guard is unchanged, so an image that does not carry the
stock RAMTAS bytes is left alone.

`C64::enable_kernal(uint8_t *rom)` has the same `memcmp`/`memcpy` shape and is left alone. Its
argument is main memory in both call paths, a heap buffer or the cartridge ROM area in SDRAM,
and it copies to the aperture afterwards. The non-U64 branch of `init_system_roms()` is
unaffected for the same reason.

## The test

`tests/e2e/io/c64/fast_reset_test.py`, registered in `run-tests` as `fast-reset`. It measures
the observable a user cares about: how long the machine takes to print READY after a reset. It
sets Fast Reset, reboots (the setting is applied when the system ROMs load, so a reset alone
cannot show it), takes three samples, and requires the enabled state to be at least twice as
fast as the disabled state.

The threshold comes from the measurements. A machine that applies the patch is about ten times
faster and a machine that does not is not faster at all, so 2.0 sits between the two populations
with a wide margin either side. That margin is also what absorbs a slow link: one 400-byte
screen read costs 12 to 18 ms on this bench, and even a poll costing 0.5s would leave the
enabled measurement at roughly 0.7s against a 1.20s threshold. The suite measures that poll cost
and skips, rather than failing, when it is above 0.5s.

Two things stop a meaningless pass:

- The disabled measurement is the control. If the machine reaches READY quickly with Fast Reset
  off then it is not running the RAM test the setting removes, and the comparison says nothing.
  That is reported as a failure of the control and named as such.
- A machine whose KERNAL the firmware does not supply cannot be patched, so Fast Reset genuinely
  does nothing there. An Ultimate 64 serves a `Kernal ROM` item and always loads a file; an
  Ultimate II serves an `Alternate Kernal` item and leaves it empty when the computer's own
  KERNAL is in use. The suite skips in that case.

## How it was verified

All on an Ultimate 64 Elite I, firmware 3.15, FPGA 125, core 1.4F, deployed over JTAG.

To show the suite can fail, I built this branch with `software/io/c64/c64.cc` checked out from
`4acc148c`, so `init_system_roms()` is exactly what it is on test-merge and everything else is
this branch. The suite failed on that firmware:

```
[03] the machine runs the KERNAL RAM test when Fast Reset is off ... OK
     Fast Reset=Disabled: reset to READY 2.31s, 2.31s, 2.33s, median 2.31s
[04] the machine reaches READY sooner when Fast Reset is on ... FAIL
     Fast Reset=Enabled: reset to READY 2.31s, 2.32s, 2.32s, median 2.32s
     2.31s disabled against 2.32s enabled, 1.0x
```

The same suite against the branch as it stands passed:

```
[03] the machine runs the KERNAL RAM test when Fast Reset is off ... OK
     Fast Reset=Disabled: reset to READY 2.32s, 2.35s, 2.31s, median 2.32s
[04] the machine reaches READY sooner when Fast Reset is on ... OK
     Fast Reset=Enabled: reset to READY 0.17s, 0.17s, 0.18s, median 0.17s
     2.32s disabled against 0.17s enabled, 13.6x
```

Independently of the suite, a five-sample measurement gives a median of 0.18s enabled against
2.40s disabled on this branch, where the reverted build gave 2.40s against 2.42s.

Host gates: `tests/lib/lint_test.py`, `tests/lib/registry_test.py`,
`tests/lib/observability_test.py` and `make host_tests` all pass.

Builds: u64, u64ii and u2pl build clean and every app-space gate passes (U64 20.2% free,
U64E2_50T 40.0%, U64E2_100T 35.8%, U2 9.0%, U2+ 22.1%, U2+L 30.2%). u2 and u2plus fail at the
updater packaging step with `No rule to make target 'rv700dd.chk'`, which is a missing cached
FPGA blob on this machine. I confirmed it is pre-existing by running the same two builds on the
tree with these commits stashed: they fail identically there.

### On an Ultimate 64 Elite I

- `./run-tests u64`, the default gate: 25 of 25 suites passed, 266s. `fast-reset` is not in that
  profile, for the same reason `reu-turbo` and `freeze-menu` are not.
- `./run-tests u64 --profile=standard`, which does include the new suite: 38 of 38 suites passed,
  862s. The new suite reported `fast-reset: OK` with 2.32s disabled against 0.17s enabled, 13.4x.
- After the review changes, on the rebuilt firmware: `fast-reset: OK`, four checks, 12.2s, 2.32s
  disabled against 0.17s enabled.

### On an Ultimate II+L in a C64 Ultimate

The cartridge was flashed with the firmware this branch builds and rerun after the review
changes:

- `./run-tests u2@c64u --profile=standard`: **38 of 38 suites passed, 816s.**
- `fast-reset` skips on that machine and says why: an Ultimate II+L takes its KERNAL from the
  computer unless an Alternate Kernal is configured, so the setting has nothing to patch there.
  The suite declines to measure rather than reporting a verdict it cannot support.

That run is the hardware evidence that this change leaves the Ultimate II family working. The
object comparison in the next section is the static evidence for the same claim.

## Review changes

Three points from review, all applied:

- A short KERNAL is rejected by length rather than the staging buffer being cleared. Clearing
  cost a `memset` on every ROM load and did not make a short file usable, since without the
  reset vector at `$FFFC` the machine cannot start. `load_file` reports the length it
  transferred, so anything other than 8192 falls back to the default.
- The comments were cut back to the reason the staging copy exists. They no longer describe what
  the code did before.
- The suite measures one reset per state rather than a median of three. The two states are about
  10x apart and a machine that cannot patch is 1.0x, so a single reset each separates them with
  a wide margin. The suite runs in 12.2s.

## Effect on the Ultimate II family

Fast Reset exists on an Ultimate II as well, but it reaches the KERNAL by a different route.
`CFG_C64_FASTRESET` is outside the `#if U64`, so the setting is offered there, and the non-U64
branch of `init_system_roms()` loads the *Alternate Kernal* file into a heap buffer and hands
that buffer to `enable_kernal()`, which applies the patch and then scatters the image into the
KERNAL replacement shadow RAM at `__kernal_area`. That path never touches an FPGA ROM aperture,
and the comparison it makes is against ordinary memory, so it worked before this PR and is
unchanged by it. When no alternate KERNAL is configured, `load_file` fails, `disable_kernal()`
runs, and the computer supplies its own KERNAL, which the firmware cannot patch at all; Fast
Reset genuinely does nothing in that case, before and after this change.

I could not test on a U2, so I measured the risk instead of asserting it. `-DU64=1` appears only
in the three `target/u64` makefiles, so an Ultimate II build compiles the `#else` branch, which
this PR does not touch. To confirm that at the object level I compiled `software/io/c64/c64.cc`
for U2 (RISC-V), U2+ (Nios II) and U2+L (RISC-V) twice, once from this branch and once with that
file checked out from `4acc148c`, and compared the results. The disassembly of the resulting
`c64.o` is identical in all three cases, and so are the `.rodata` and `.data` section contents.
The object files themselves differ only in DWARF line information, because the added lines shift
the line numbers of everything below them in the file. So the machine code an Ultimate II boots
is bit for bit what it was, and the risk to U2 boot from this change is zero rather than small.

## What I could not verify

Everything measured here is on an Ultimate 64 Elite I. I have not run this change on a C64
Ultimate or on any Ultimate II, so for those the argument is from the code and from the object
comparison above, not from a measurement.

For the U64 Elite II, which compiles the same `#if U64` branch, I can say what the change does
rather than what the hardware does. It removes a dependency rather than adding one: before, the
patch was applied only when a read of the aperture returned the image that had just been
written; after, the decision is made on the copy in main memory and the aperture is only
written. On hardware where the aperture does read back, both versions apply the patch to the
same bytes and the machine sees the same image. On hardware where it does not, only the new
version applies it.

I also did not test the second call path into `enable_kernal()`, the one a cartridge requiring
`CART_KERNAL` takes. That path is unchanged by this PR.

## Alternatives considered

Patching without the `memcmp` guard would have removed the aperture read too and needed no
staging buffer. Rejected: the guard is what stops 22 bytes being written over $FD6C in a KERNAL
that is not the stock one, and a user who has put a custom KERNAL in `/flash/roms` would get a
corrupted image instead of a machine that simply boots at the normal speed.

Reading only the 47 bytes at $1D6C from the file for the comparison, to avoid the 8 KB
allocation, would need a second pass over the file and `load_file` takes no offset, so it would
have read the whole 8 KB anyway. The allocation matches what the non-U64 branch of the same
function already does.

### DIFF
--- a/run-tests
+++ b/run-tests
@@ -425,6 +425,13 @@ SUITES: Sequence[Suite] = (
     # back the wrong bytes. Measured red on core 1.4E and green on 1.4F.
     Suite("e2e", "reu-turbo", "tests/e2e/io/c64/reu_turbo_test.py",
           "-H @HOST@ -p @PASS@ -t @TIMEOUT@"),
+    # Fast Reset patches the KERNAL's RAM test out of the image the machine
+    # boots, so the observable is how long a reset takes to reach READY. The
+    # setting is applied when the system ROMs load, so the suite reboots
+    # between the two states. Measured red at 1.0x on an Ultimate 64 before the
+    # fix and green at about 10x after it.
+    Suite("e2e", "fast-reset", "tests/e2e/io/c64/fast_reset_test.py",
+          "-H @HOST@ -p @PASS@ -t @TIMEOUT@"),
     # Reproduces #733: with the FPGA decoder fix in the core, "mirroring off" is
     # the regression guard. Runs last, because a device whose core predates that
     # fix stalls the whole Nios on this scenario and needs a JTAG recovery.
--- a/software/io/c64/c64.cc
+++ b/software/io/c64/c64.cc
@@ -1083,15 +1083,14 @@ void C64::init_system_roms(void)
     extern uint8_t _default_kernal_65_start[];
     extern uint8_t _default_chars_bin_start[];
 
-    // The ROM image aperture at U64_KERNAL_BASE is write only: a read of it returns
-    // zeros, so the memcmp below never matched there and Fast Reset was never
-    // applied. Build the image in RAM instead and write it to the aperture once.
-    // Cleared first because load_file reports FR_OK for a file shorter than the
-    // buffer and leaves the remaining bytes untouched.
+    // U64_KERNAL_BASE is write only, so the Fast Reset patch below cannot compare
+    // against it. Stage the image in RAM and write it to the aperture once.
+    // A KERNAL shorter than 8192 bytes has no reset vector at $FFFC, so it is
+    // rejected in favour of the default.
     unsigned char *kernal = new unsigned char[8192];
-    memset(kernal, 0, 8192);
-    FRESULT fres = FileManager :: getFileManager()->load_file(ROMS_DIRECTORY, cfg->get_string(CFG_C64_KERNFILE), (uint8_t *)kernal, 8192, NULL);
-    if (fres != FR_OK) {
+    uint32_t kernal_bytes = 0;
+    FRESULT fres = FileManager :: getFileManager()->load_file(ROMS_DIRECTORY, cfg->get_string(CFG_C64_KERNFILE), (uint8_t *)kernal, 8192, &kernal_bytes);
+    if (fres != FR_OK || kernal_bytes != 8192) {
         printf("Failed to load KERNAL ROM; loading default.\n");
         memcpy((void *)kernal, (void *)_default_kernal_65_start, 8192);
     } else if (cfg->get_value(CFG_C64_FASTRESET)) {
--- a/software/io/c64/c64.cc
+++ b/software/io/c64/c64.cc
@@ -1083,26 +1083,20 @@ void C64::init_system_roms(void)
     extern uint8_t _default_kernal_65_start[];
     extern uint8_t _default_chars_bin_start[];
 
-    // The FPGA ROM image aperture at U64_KERNAL_BASE is write only. A read of it
-    // on an Ultimate 64 Elite I returns zero for all 8192 bytes, both through a
-    // block memcpy and byte by byte through a volatile pointer, and a read back
-    // straight after a write returns zeros as well. So the KERNAL image is staged
-    // in a buffer in main memory, where the Fast Reset patch can compare it
-    // against the unpatched original, and only then written to the aperture.
-    // Cleared first because load_file answers FR_OK for a file shorter than the
-    // buffer and leaves the rest of it untouched. Writing to the aperture used
-    // to leave the bytes past the end of such a file as they were; staging in a
-    // fresh allocation would put uninitialised memory there instead.
-    uint8_t *kernal = new uint8_t[8192];
+    // The ROM image aperture at U64_KERNAL_BASE is write only: a read of it returns
+    // zeros, so the memcmp below never matched there and Fast Reset was never
+    // applied. Build the image in RAM instead and write it to the aperture once.
+    // Cleared first because load_file reports FR_OK for a file shorter than the
+    // buffer and leaves the remaining bytes untouched.
+    unsigned char *kernal = new unsigned char[8192];
     memset(kernal, 0, 8192);
-    FRESULT fres = FileManager :: getFileManager()->load_file(ROMS_DIRECTORY, cfg->get_string(CFG_C64_KERNFILE), kernal, 8192, NULL);
+    FRESULT fres = FileManager :: getFileManager()->load_file(ROMS_DIRECTORY, cfg->get_string(CFG_C64_KERNFILE), (uint8_t *)kernal, 8192, NULL);
     if (fres != FR_OK) {
         printf("Failed to load KERNAL ROM; loading default.\n");
-        memcpy(kernal, (void *)_default_kernal_65_start, 8192);
-    }
-    if (cfg->get_value(CFG_C64_FASTRESET)) {
+        memcpy((void *)kernal, (void *)_default_kernal_65_start, 8192);
+    } else if (cfg->get_value(CFG_C64_FASTRESET)) {
         if (!memcmp((void *) (kernal+0x1d6c), (void *) fastresetOrg, sizeof(fastresetOrg))) {
-            memcpy((void *) (kernal+0x1d6c), (void *) fastresetPatch, sizeof(fastresetPatch));
+            memcpy((void *) (kernal+0x1d6c), (void *) fastresetPatch, 22);
         }
     }
     memcpy((void *)U64_KERNAL_BASE, kernal, 8192);
--- a/software/io/c64/c64.cc
+++ b/software/io/c64/c64.cc
@@ -1083,16 +1083,30 @@ void C64::init_system_roms(void)
     extern uint8_t _default_kernal_65_start[];
     extern uint8_t _default_chars_bin_start[];
 
-    FRESULT fres = FileManager :: getFileManager()->load_file(ROMS_DIRECTORY, cfg->get_string(CFG_C64_KERNFILE), (uint8_t *)U64_KERNAL_BASE, 8192, NULL);
+    // The FPGA ROM image aperture at U64_KERNAL_BASE is write only. A read of it
+    // on an Ultimate 64 Elite I returns zero for all 8192 bytes, both through a
+    // block memcpy and byte by byte through a volatile pointer, and a read back
+    // straight after a write returns zeros as well. So the KERNAL image is staged
+    // in a buffer in main memory, where the Fast Reset patch can compare it
+    // against the unpatched original, and only then written to the aperture.
+    // Cleared first because load_file answers FR_OK for a file shorter than the
+    // buffer and leaves the rest of it untouched. Writing to the aperture used
+    // to leave the bytes past the end of such a file as they were; staging in a
+    // fresh allocation would put uninitialised memory there instead.
+    uint8_t *kernal = new uint8_t[8192];
+    memset(kernal, 0, 8192);
+    FRESULT fres = FileManager :: getFileManager()->load_file(ROMS_DIRECTORY, cfg->get_string(CFG_C64_KERNFILE), kernal, 8192, NULL);
     if (fres != FR_OK) {
         printf("Failed to load KERNAL ROM; loading default.\n");
-        memcpy((void *)U64_KERNAL_BASE, (void *)_default_kernal_65_start, 8192);
-    } else if (cfg->get_value(CFG_C64_FASTRESET)) {
-        unsigned char *kernal = (unsigned char *)U64_KERNAL_BASE;
+        memcpy(kernal, (void *)_default_kernal_65_start, 8192);
+    }
+    if (cfg->get_value(CFG_C64_FASTRESET)) {
         if (!memcmp((void *) (kernal+0x1d6c), (void *) fastresetOrg, sizeof(fastresetOrg))) {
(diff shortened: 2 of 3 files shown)


chrisgleissner:
Hi @GideonZ , as discussed, this fixes a major user facing regression. Thanks.

chrisgleissner:
Merging this despite change freeze on `test-merge`, as agreed with @GideonZ , due to the impact of this bug.

Issue #670, opened 2026-03-29
Request Feacture: add ReST API for remote control Joystick


Zilog80c:
Hi,
   thanks for your incredibile work!

    **Is your feature request related to a problem? Please describe.**
    New REST API Request for remote control Joystick Port 1 and Port 2.

    **Describe the solution you'd like**
    In the  ReST API there is “Machine” section , where is possbile , for example, reboot the machine “PUT  /v1/machine:reboot”
    If possibile i request to add this Rest API for remote control joystick:
    machine:JoySend (“PORT” 1 or 2 , “ACTION” direction or fire, “STATE” enable o disable)
    Sample:    machine:JoySend ( 1 , UP , ENABLE)
                     machine:JoySend ( 1 , UP , DISABLE)
	            machine:JoySend ( 2 , LEFT , ENABLE)
                    machine:JoySend ( 2 , LEFT , DISABLE)
                     
                    Keep the diagonals in mind too
 	            machine:JoySend ( 1 , UPRIGHT , ENABLE)
	            machine:JoySend ( 1 , DOWNRIGHT , ENABLE)	 
	            machine:JoySend ( 1 , UPLEFT , ENABLE)
	            machine:JoySend ( 1 , DOWNLEFT , ENABLE)				
	
**Describe alternatives you've considered**
I tried using RESTAPI “write mem” to write directly to the CIA address, but it doesn't work properly or at all.

**Additional context**
Motivation: Play to Other friend over Internet 😊

Thanks a lot.

enver-haase:
Likewise, I would love to remote-inject into the keyboard matrix, not just into the KERNAL's cooked buffer of characters.
Adding this here as it's the same CIA virtualization matter.

chrisgleissner:
Thanks for raising this. I would find this useful as well. I have run into similar limitations while working on U64 REST clients, and a dedicated input API would make remote control, automation, and testing much cleaner.

I looked at the firmware's existing REST API before suggesting a shape for this. The proposal below tries to stay close to the current `/v1/*` conventions, especially lowercase enum values, snake_case fields, JSON request bodies, and explicit validation.

Thanks also to @Zilog80c for the valuable input.

## Suggested REST API

The API uses one endpoint with two methods:

```http
POST /v1/machine:input    submit an ordered batch of input events
GET  /v1/machine:input    read the current REST-injected input state
```

This covers keyboard input, joystick input, batching, recovery, and test verification without introducing separate endpoint families for each input device.

[...]

PR #869, opened 2026-09-07
Let a `restore` tap carry matrix keys in one machine:input request

chrisgleissner:
## Previous behaviour

`PUT`/`POST /v1/machine:input` refused `restore` in any keyboard event that named another input. The rule was one condition in `software/api/input_api.h`:

```c
if (has_restore && ((count != 1) || (out.transition != INPUT_PARSED_TAP))) {
    input_api_set_error(err, err_size, "`restore` must appear alone in `inputs` and only with transition `tap`.");
```

So a caller wanting C= plus RESTORE had to send three requests: press `commodore`, tap `restore`, release `commodore`. Same for RUN/STOP plus RESTORE.

## Why it was that way, as far as I could establish

The rule has two halves and only one of them has a stated reason.

The transition half comes from the API specification in #670: "Unless the firmware has a clear and safe way to model a held RESTORE state, `restore` should only be valid with `transition: tap`." That is correct and stays.

The "must appear alone" half is not in the specification, is not explained in the code, and I found no discussion of it in #698 or #670. The guard and the message were added in the one commit that added the endpoint (0f2d10ee).

The delivery path below the parser already handled the combined form. `apply_keyboard_event` in `software/api/route_input.cc` splits a tap event into a matrix and a restore flag and passes both to one `Keyboard_USB::restQueueTap()` call. `RouteInputResponseKeys::addKeyboardTap` does the same split for the response snapshot. Nothing under the parser needed a change.

## The change

`input_api.h` now rejects `restore` with `press` and with `release`, and accepts it in a `tap` alongside matrix keys. The error text is now "`restore` is only valid with transition `tap`."

`press` and `release` stay rejected because `route_input.cc` skips the restore entry when it builds the live matrix, in the press path (line 505) and in `release_live_keyboard_event`. Accepting either would answer HTTP 200 and do nothing.

The `API_DOC` description in `route_input.cc` is updated and `doc/api/rest_api_openapi_u64.yaml` is regenerated from it. `make openapi_check` and `make openapi_validate` both pass.

## Why the ordering works

RESTORE is an edge on the NMI line, not a matrix column. C= plus RESTORE is recognised because the CBM key is already down in the keyboard matrix when that edge arrives, so a combined request is only correct if it produces that order.

It does, and no `setup_matrix` staging is needed. `Keyboard_USB::startRestTap()` sets the matrix overlay bits and `rest_restore_overlay` inside one critical section. `Keyboard_USB::applyMatrixState()` then writes the eight matrix rows at `matrix[0..7]` before it writes the restore register at `matrix[9]`, so the columns are asserted at the hardware register block before the restore line rises. Both stay asserted for `REST_KEYBOARD_TAP_HOLD_QUEUE_TICKS`, which is two ticks of the 20ms REST input timer.

## How it was verified

Host tests: `make -C software/api/tests` passes, 15 validation tests and 28 state tests. New tests cover a `restore` press and a `restore` release being rejected, a `restore` tap alone still being accepted, a tap carrying `restore` with a matrix key in either order being accepted, and a duplicate `restore` still being refused. `RestKeyboardStateTest.QueuedTapDrivesMatrixAndRestoreTogether` asserts the hardware register array rather than the REST snapshot: after one tick a combined queued tap has the CBM column bit set in `matrix[7]` and the restore bit set in `matrix[9]`, and both clear together.

Hardware: an Ultimate 64 Elite, firmware 3.15, running upstream `test-merge` tip `b3c76093` plus this commit, downloaded over JTAG with `nios2-download`.

An NMI handler installed at $C000 through NMINV ($0318) drove keyboard matrix row 7 and stored the column byte read from $DC01 at every NMI, along with counters for how many NMIs it saw and whether CBM (row 7, bit 5) was down. Five runs of each form, with the counters zeroed before each:

| Stimulus | NMIs | Row 7 column byte |
|---|---|---|
| `restore` tap alone | 1 | `$FF` (CBM up) |
| three requests: press `commodore`, tap `restore`, release `commodore` | 1 | `$DF` (CBM down) |
| one request: tap `["commodore", "restore"]` | 1 | `$DF` (CBM down) |

Every run of each form gave the same result. The single request produces the same stimulus at the NMI as the three-request sequence.

A second observable, from the KERNAL rather than from a program of mine: `$D020` was set to 5 over DMA, then `["run_stop", "restore"]` was tapped in one request. The KERNAL warm start ran, cleared the screen, printed READY. and put `$D020` back to 14. The three-request form gave the same result. Before this change the single request answered HTTP 400.

Rejections and the unchanged case, on the same device: `restore` with `press` and with `release` both answer HTTP 400 with the new message, `["commodore", "restore"]` with `press` and with `release` likewise, and a `restore` tap on its own answers HTTP 200 with `"inputs": ["restore"]`.

## What this does not claim

It does not claim that one request opens a C64 Ultimate's menu.

C= plus RESTORE to menu is not implemented in this repository's sources. `U64_RESTORE_REG` has two users, both the safe-mode check at boot. The menu button reaches software through the ITU (`fpga/io/itu/vhdl_source/itu.vhd`), and the mk1 top levels wire that signal straight to physical pins, with no keyboard matrix and no RESTORE line involved. `fpga/fpga_top/` carries no C64 Ultimate top level; that bitstream is prebuilt in `external/`. So the mapping lives in a bitstream that is not in this tree, on a machine whose firmware this tree cannot build.

What the evidence above establishes is narrower and is the part the API controls: the combined request delivers the same stimulus to the C64 as the sequence that works today. Whether a given core reads that stimulus as a menu request is the core's business and is unchanged by this commit.

## E2E coverage

`tests/e2e/lib/menu.py` gains a small facade for getting the menu open, because there are two signals for it and they are not the same thing. `ButtonMenuOpener` is `machine:menu_button`, which every machine answers. `CommodoreRestoreMenuOpener` is the keyboard route, and it holds the only copy of the request sequence, so the three-request form becomes one request in one place when a machine's firmware carries this change. A table says which machines have a keyboard route; a machine that is not in it has none, which is a statement about its core rather than about its firmware version.

`tests/e2e/api/input_test.py` gains one check, selected by `--test menu-open`, that opens the menu through the keyboard opener. It reports SKIP on a machine with no keyboard route, naming the machine and saying that `machine:menu_button` is the only signal there. Measured on the Ultimate 64 Elite:

```
[01] a keyboard combination opens the menu ... SKIP (Ultimate 64 Elite 3.15 has no
     keyboard route into its menu; machine:menu_button is the only signal that opens
     it, 0.000s)
input_test: OK (1 checks, 3.443s)
```

The check deliberately still sends the three-request sequence. The one machine on this bench with a keyboard route into its menu runs a separate 1.2 release line whose commit is not in this repository, so it cannot take this change, and a check that required the combined form would fail there rather than measure anything. The facade is where that becomes a single request once the firmware lines meet.

## Alternative rejected

Staging the matrix keys through `restQueueTap`'s `setup_matrix` argument, so the CBM key goes down one tick before the restore edge. That argument exists for the menu's string editor, where a modifier has to be down for a scan before the key it modifies. It is not needed here: `applyMatrixState()` already writes the matrix rows before the restore register within one call, and the measurement above shows the CBM key down at the NMI in every run. Adding a tick of staging would have slowed a combined tap for no observable gain.


chrisgleissner:
For reproducing the ordering measurement, here is the NMI handler the table in the description came from. It is not committed, because it needs firmware carrying this change and the only bench machine that can take that has no keyboard route into its menu, so there is nothing for a suite to assert on it beyond what the host tests already cover.

The handler is installed at `$C000` and reached through NMINV (`$0318`), so it runs instead of the KERNAL's own NMI routine. On each NMI it drives keyboard matrix row 7, reads the column byte from `$DC01`, and records it along with two counters.

```
C000  48        PHA
C001  8A        TXA
C002  48        PHA
C003  A9 7F     LDA #$7F        ; row 7 low
C005  8D 00 DC  STA $DC00
C008  AD 01 DC  LDA $DC01       ; settle
C00B  AD 01 DC  LDA $DC01
C00E  8D 00 C9  STA $C900       ; the column byte at this NMI
C011  29 20     AND #$20        ; CBM is row 7, bit 5
C013  D0 06     BNE $C01B
C015  EE 01 C9  INC $C901       ; CBM was down
C018  4C 1E C0  JMP $C01E
C01B  EE 02 C9  INC $C902       ; CBM was up
C01E  EE 03 C9  INC $C903       ; NMIs seen
C021  A9 00     LDA #$00
C023  8D 00 DC  STA $DC00       ; rows as the KERNAL scan leaves them
C026  68        PLA
C027  AA        TAX
C028  68        PLA
C029  40        RTI
```

The 42 bytes go in with `PUT /v1/machine:writemem`, NMINV is set to `00 C0` the same way, and the counters are read back with `GET /v1/machine:readmem?address=C900&length=4`. `$C901..$C903` are zeroed before each run. Afterwards NMINV goes back to `47 FE` and the machine is reset.

Only `$C900` and `$C903` are strictly needed; the two counters are there so a run that took more than one NMI, or none, is visible rather than being read off a single sampled byte.

chrisgleissner:
Hi @GideonZ ,

at the moment, pressing `C=` + `RESTORE` on the C64U via our REST API requires three separate REST calls:
1. Press `C=`
2. Tap `RESTORE`
3. Release `C=`

With this PR, only a single REST call is needed.

As you see, the prd code changes are minimal since the only change needed was to remove an explicit rejection of combining `RESTORE` with other keys in the same REST call. Further down the stack, it was already supported.

Thanks
Christian

PR #865, opened 2026-09-06
Preserve Assembly64 query fields after dropdown selection

barryw:
Selecting a dropdown in the Assembly64 query form erased the selected value and every previously entered field. Form states now opt out of the automatic reload after context actions; directory and result browsers retain their existing reload behavior, and explicit CLEAR still resets the form. The shared FormUI editor receives the same protection.

Fixes #863.

The production change adds five lines and changes one. The regression scenario is part of the existing Assembly64 hardware suite and checks populated text fields, +/- changes, cancellation, and confirmation for Repo, Category and Subcat, including retention of earlier selections.

Validation on Ultimate 64 Elite, FPGA 124/core 1.4F:
- **TDD red:** before the fix, population, +/- and cancellation passed; confirming Repo failed with `Name: changed from 'turrican' to ''`.
- **TDD green:** all 10 regression checks pass on the flashed fix.
- Full Assembly64 overlay suite: **27 checks pass**, including a real search returning 20 matching results, aborted edits, empty-query handling and repeated reopening.
- Full browser filesystem refresh suite: **38 checks pass**, with 140 successful observer comparisons across Menu, Telnet, FTP and REST, including event-queue pressure, deletion, copying and moving.
- Separate hardware check confirms explicit CLEAR still empties populated text and preset fields. Menu closed and temporary test files absent afterward.
- Full Nios II application/updater builds, Python syntax, CRLF-aware diff checks, and GitHub build CI pass.

The hardware application was built from base `abd0888c` plus the exact production changes in `94a3d455`; it reports the base hash. Application SHA256: `6ab20214e558d756da53018140908c47aa4c614c2b43f232674fd9f350928318`. The updater's FPGA payload matches the board's existing FPGA 124 image. The reporter's Elite-II model was not separately tested.

Issue #863, opened 2026-09-06
Assembly 64 Query Form broken in latest "test-merge" builds?


telcirion:
**Describe the bug**
In the Assembly 64 query form, selecting a value from a list clears the entire form. (using + and - seem to work fine though). I don't know if is this already on the radar but just in case it isn't.

**To Reproduce**
1. From starting screen, go to ultimate menu with commodore+restore or power-switch ' up'.
2. Press ' F6' -> empty Assembly Query form is shown as expected.
3. Fill in name, group, handle, and/or Event.
4. Navigate with cursor keys to Repo, Category, Subcat, or any other field which value may be selected from a list.
5. Press 'Return'.
6. Navigate with cursor keys to a value to be selected.
7. Press ' Return'.
8. The entire form is empty again.

**Expected behavior**
The chosen field is filled with selected value, all other field keep their previous filled in values. Just like with firmware 3.14.

**Screenshots**
None.

**Applies to (please complete the following information):**
 - Hardware type: U64E2
 - Firmware Version 3.15 (update_v3.15-121-g7e30b13e.ue2) 

**Additional context**
None.

barryw:
@telcirion thanks for the bug report. i'll take a look.

chrisgleissner:
@telcirion The fix was merged into the test-merge branch. 

Thanks for retesting and confirming that it is fixed. Also many thanks for spotting it and to @barryw for the quick fix.

telcirion:
Yes, it works properly again! @barryw and @chrisgleissner, thanks for the quick fix and merge.

PR #870, opened 2026-09-07
Improve E2E tests for use against C64U

chrisgleissner:
Three changes to the E2E harness, all under `tests/` and `run-tests`. No firmware is touched. Two are defect fixes found by running the suites on a C64 Ultimate; the third is a new suite guarding a change that is not in this branch yet.

Branched from `test-merge` at `b3c76093` ("Fix Fast Reset on the Ultimate 64 (#867)"). Ten files, +1569/-349, all under `tests/` and `run-tests`, and no binary entries in `git diff --numstat upstream/test-merge..HEAD`.

## 1. The online search suite named the search form's fields

Previous behaviour. `scenario_dropdown_preserves_fields` in `tests/e2e/io/c64/assembly64_test.py`, added by `4acc148c` (#865), iterated the literal tuple `("Repo:", "Category:", "Subcat:")` and raised a `Failure` when `row_of` could not find one of them. The rest of the suite named `"Name:"` and `"<<"` in the same way.

The defect. Those labels are not the firmware's. Four text fields are fixed, and after them `BrowsableAssemblyRoot::getSubItems` appends one `BrowsableQueryField` per `{type, values}` entry in the presets the remote service answered with (`software/userinterface/assembly_search.h`), so the field set follows the service. #865 landed after #849 ("Support all Ultimate devices in every E2E test") and was never run on a C64 Ultimate, which searches CommoServe rather than Assembly 64.

Evidence, harness commit `b3c76093`, overlay mode. On `u64` (Ultimate 64 Elite, firmware 3.15) the suite passed with 27 checks. On `c64u` (C64 Ultimate, firmware 1.2RC) it failed on all three attempts, and again in the freeze pass, with the first failing check `[07] Repo: +/- preserves existing fields` and the message `the form has no 'Repo:' field`. The check before it, "populate the text fields", passed, so the form was reachable and drivable and only the preset fields differed.

The change. `tests/e2e/lib/search_form.py` is a new page object for the query form, written against `Backend` in the same idiom as `browser.py`, so the same steps work over REST and Telnet. It opens the form, reads the fields off the rendered screen (label, row and current value, in the order the form draws them), classifies each field by driving it, and offers focus, type, cycle, cancel, confirm, clear and submit, plus a whole-form value snapshot and an "unchanged except for these" assertion.

Classification is behavioural rather than tabular: a preset field takes a value when it is cycled with `+`, a free-text field does not. `+` reaches `AssemblySearchForm::increase`, which `BrowsableQueryField::updown` ignores unless the field's target is a list, and it is not a character the string editor can receive, so probing a free-text field with it changes nothing.

The suite no longer holds any field label. `row_of`, `field_value`, `enter_field`, `select_row`, `open_query_form`, `leave_form`, `submit_query`, `empty_field`, `telnet_field_row`, `no_cursor_reason` and `_box_interior_bounds` moved into the new module. What the form is called and where the menu keeps it stay in `tests/lib/machine.py`: that is navigation, and it is a property of the machine. What the form contains is not, and is now discovered.

Discovery is guarded so it cannot degrade into covering nothing. The scenario reports the drawn fields, the free-text fields and the dropdown fields through `detail()`, so every run record names what was covered on that machine. It fails, rather than skipping, when no field takes a value when cycled, and it fails when every field is a dropdown. A field that discovery found and that cannot then be driven is still a hard failure: `SearchForm.field` raises rather than answering `None`.

The concrete effect on coverage: `assembly64` on `u64` runs 46 checks where it ran 27. Discovery finds 13 fields on the Assembly 64 form (Name, Group, Handle, Event, Repo, Category, Subcat, Rating, Type, Date, Latest, Sort, Order), classifies 4 as free text and 9 as dropdowns, and drives all 9 rather than the 3 that were named.

Two checks changed shape, and both changes are visible on `u64`.

The `+/-` check used to require the value to come back to where the sequence `+ + -` started. `updown()` clamps at both ends, so that only holds for a preset list of a particular length. The check now reads the field's own value back after each of the three keystrokes, requires it to be non-empty, and asserts that no other field moved. That is the #863 behaviour and it does not depend on how many presets a service listed.

The "confirming preserves its value" check passed only by accident. The context menu opens on its own first item (`AssemblySearchForm::change` calls `browser->context(0)`) rather than on the value the field holds, so DOWN, UP, RETURN applies the first preset. The old key sequence always left the field on that first preset, so the value happened not to change. Measured here, a field left on any other preset is reset to the first one by that sequence: `Repo: changed from 'Gamebase64' to 'CSDB'`. The check now rewinds the field to its first preset before opening the menu, which makes "confirming without moving preserves the value" an assertion about the firmware rather than about the order the service listed its presets in.

### A firmware defect the discovery reached

With every field of the CommoServe File Search form holding a value, confirming any item in a preset field's context menu empties the whole form on a C64 Ultimate 1.2RC. Measured in overlay mode on 2026-09-07:

```
before {'Name:': 'name0', 'Group:': 'group1', 'Handle:': 'handle2',
        'Event:': 'event3', 'Category:': 'Apps', 'Date:': '1980',
        'Type:': 'crt', 'Sort:': 'Name', 'Order:': 'Ascending'}
after  every one of those nine ''
```

Three key paths do it: RETURN on the item the menu opens on, DOWN then UP then RETURN, and DOWN then RETURN. Leaving the menu with RUN/STOP keeps the fields, and so does cycling the same field with `+` and `-`, so only the context-menu action is affected. An Ultimate 64 Elite on 3.15 keeps every field on all three paths, which is the behaviour #865 gave the Assembly 64 form; the CommoServe form is drawn by a different class in that release line.

That is a firmware gap rather than a harness one, so it is recorded in the `FIXES` table in `tests/lib/machine.py` as `query-form-survives-dropdown-confirm` with the C64 Ultimate listed as lacking it, and the "confirming ..." checks are tagged with it. They report SKIP naming the fix and the machine on a C64 Ultimate and run everywhere else. Skipping rather than failing there is what lets the rest of the suite run on that machine: the defect empties every field, so a form left in that state would fail every later check for a reason unrelated to what those checks assert.

## 2. The CFG loader log was parsed by its first quote

Previous behaviour. `loader_report` in `tests/e2e/filemanager/cfg_single_group_test.py` took a store name with `line.split("'", 2)[1]`, requiring only that the line started with the loader's prefix.

The defect. The device writes that debug log with `printf`, which emits one character at a time through `outbyte()` and takes no lock (`software/system/small_printf.cc`, `software/application/ultimate/ultimate.cc`). A task switch part way through a line splices another task's output into the middle of it. The REST poller that overlay mode runs prints `Accept client 0 on socket 7.` for every connection, so it is the task whose text lands in the loader's lines. A prefix-only parse then reports the spliced text as a store name, or misses the record entirely when the splice lands before the prefix.

Evidence, from one exhaustive run of `b3c76093` against `c64u` (C64 Ultimate, firmware 1.2RC), overlay mode. `cfg-partial-effectuate` failed twice and passed on the third attempt:

- attempt 1: `the loader effectuated 0 store(s): []` — the record had been appended to another task's line, so it no longer began with the prefix and was not counted at all;
- attempt 2: `1 store(s) the file never mentioned were applied: ['Audio MAccept client 0 on socket 7.  192.168.1.185:49362']` — the real store is `Audio Mixer`, and the splice ends at the interrupting task's own newline, so the rest of the record (`ixer' after loading.`) landed on the following line;
- attempt 3: OK, 1 effectuated and 21 left clean.

The same suite also failed once in the freeze pass of that run, with the attempt-1 shape. The suite was therefore flaky rather than broken, and its failure message printed spliced text where a reader expects a store name.

The change. A line is read as a loader record only when it starts with the prefix, ends with the matching suffix, and the name between them holds no quote. Lines that hold a prefix or a suffix anywhere without being a whole record are collected into a third list, `damaged`, rather than dropped: dropping them would let an interleaved log read as a load that touched fewer stores than it did, which is the failure `cfg_partial_effectuate_test` exists to detect. Containment rather than `startswith`/`endswith`, because attempt 1 above shows a splice can land before the prefix. The two prefixes cannot be confused: `Effectuating settings of store '` spells "store" in lower case and `Store '` in upper case.

`cfg_partial_effectuate_test` decides in that order. An unexpected store read from an intact line still fails the check, so a real loader defect stays visible whatever else the log holds. Only when the intact lines carry no unexpected store and the log also holds a damaged line is the result reported as SKIP, naming the count and printing each unreadable line: a spliced line may have carried a store the `.cfg` never named, so from that log neither a pass nor a fail can be justified.

Scope note. `loader_report` is defined in `cfg_single_group_test.py` but only `cfg_partial_effectuate_test.py` calls it. `cfg-single-group` loads the `.cfg` and asserts nothing about the log, so it was not failing on this; it shares the module, not the call.

## 3. New suite for the deferred machine actions in commodore#1

`GideonZ/commodore#1`, commit `c66ef028`, changes two things. `C64_Subsys::executeCommand` released the host and reset the machine while the menu was still drawn; it now queues the action (`C64::defer_machine_action`), answers `MENU_HIDE`, and `UserInterface::run_once` performs it after `release_host` and `release_ownership`. Separately, `C64::checkButton` kept its edge detector in a function-static `button_prev`, which a reset left holding the pre-reset level, so the next press produced no edge; `button_prev` is now a member that `C64::syncButtonState` resyncs after `run_once` returns.

`tests/e2e/io/c64/deferred_actions_test.py` is registered as `deferred-machine-actions` with no explicit profile, so the standard profile and everything above it select it. Ten checks in four scenarios.

The two REST checks run everywhere. With no menu open, `cmd->user_interface` is NULL and `request_hide_then_action` takes its else branch, so `PUT /v1/machine:reset` and `PUT /v1/machine:menu_button` must act on the machine straight away.

The three screen-cleanliness checks run everywhere as well, one per route into the deferred path: "Reset C64" and "Reboot C64" from the task menu, and C= + R in the file browser, whose `KEY_CTRL_R` case in `tree_browser.cc` issues `MENU_C64_RESET` and returns `user_interface->menu_response_to_action`. They are not tagged with a firmware fix: measured on `u64`, whose firmware carries none of `c66ef028`, all three pass, so a vintage gate would report SKIP for behaviour the machine satisfies and would leave the invariant unguarded on the machine that holds it.

The three menu-reopening checks report SKIP everywhere, because no route this harness has reaches `C64::checkButton`'s edge detector. `PUT /v1/machine:menu_button` calls `setButtonPushed()` directly and never enters it. The other route is the hardware button, or the C64 keyboard combination that raises the same signal on a C64 Ultimate, and injecting that combination does not reach it. Measured on `c64u` with the Overlay interface selected: a press of `commodore`, a tap of `restore` and a release of `commodore` left `machine:menu_screen` answering 404 for the whole 8s it was polled, with the same result when the three requests were spaced 0.5s apart and when `restore` was tapped alone, while `PUT /v1/machine:menu_button` in the same probe opened the menu every time. The same sequence does nothing on `u64`. The mapping is in the prebuilt C64 Ultimate bitstream (`external/u64e2_*.bit`) and in no C source here. The checks stay in the suite as skipped lines so every run says the behaviour was not measured and names what stops it.

The last check is separate, and it fails on `c64u`. It sends one menu button straight after a close with nothing between them. On `c64u` that press is dropped: 1 run in 4 of the whole suite fails on it, and a direct probe put it at 1 trial in 6 with no gap after the close and 0 in 6 once a 0.25s gap was left. It is last in the suite deliberately, so a failure there does not stop the run before the deferral checks, and the steps that only need the menu up retry the button rather than carrying this defect into every check. This is the same regression `readmem-writemem` fails 3 attempts of 3 on. It is left failing.

"A clean BASIC prompt" is asserted from screen RAM at `$0400`, over the whole 40x25 screen. The KERNAL's banner and READY must be present, and no cell may hold a screen code whose low seven bits are `0x40` or above; `0x00` to `0x3F` are the letters, digits and punctuation a boot screen is made of, and `0x40` to `0x7F` are the graphics the menu draws its borders with. Testing the low seven bits lets the cursor through (a reverse space, `0xA0`) and still catches a reversed menu character. The screen is blanked before each action.

That assertion is only load bearing under one interface, and the suite says which one it is running on. `--mode freeze` puts the menu in C64 screen RAM, where a leftover would be; `--mode overlay` draws it on its own plane, and the menu never writes screen RAM at all.

Two machine facts came out of running it. Which task-menu category holds the machine actions is the machine's, and is added as `Machine.machine_task_category`: `C64_Subsys` registers them under "C64 Machine", and a C64 Ultimate lists the same actions under "Power & Reset" (read off the device, where that category offered Reset C64, Reboot C64, Reboot (Clr Mem), Power OFF, Power Cycle, Save C64 Memory and Save REU Memory). Only the category differs. Separately, the suite settles the browser listing before it opens the task menu: `Browser.overlay_items` reads an overlay by diffing against the screen captured before the key, and a C64 Ultimate refreshes its drive status unprompted, so a row that changed for its own reasons is taken for part of the overlay. Measured on `c64u` in overlay mode, opening the task menu straight after `go_to_root()` parsed its categories in 3 attempts of 5, and reading until two consecutive screens matched first parsed them in 5 of 5.

Not covered: starting a `.crt` from the menu, which the author's reproduction lists. It reaches the same `request_hide_then_action` call (`C64_START_CART`), but it needs a cartridge image fixture on the device and leaves the machine running foreign code that no BASIC-prompt assertion could then read.

## Verification

Every run below was taken with the device held under a lock, one suite at a time, with retries limited to one attempt unless stated.

**u64** (Ultimate 64 Elite, firmware 3.15, `git_commit_hash b3c76093`), overlay mode, all five suites in one run: `assembly64` OK 46 checks 71.7s, `cfg-single-group` OK, `cfg-partial-effectuate` OK ("effectuated 1, left clean 18, unreadable 0"), `cfg-loader-log` OK 6 checks, `deferred-machine-actions` OK 10 checks 5.5s. Freeze mode: `assembly64` OK 46 checks 73.2s, `deferred-machine-actions` OK 10 checks 5.7s. Telnet mode: `assembly64` OK 44 checks 431s.

`assembly64` on u64 ran 27 checks in 42.3s before this change and runs 46 in 71.7s now, because discovery finds 13 fields and drives all 9 dropdowns rather than the 3 that were named.

**c64u** (C64 Ultimate, firmware 1.2RC), overlay mode: `assembly64` OK 34 checks 57.3s, where the same suite failed all three attempts before this change with `the form has no 'Repo:' field`. Discovery found 9 fields, 4 free text and 5 dropdowns, and the 5 "confirming ..." checks reported SKIP against the new `FIXES` entry. `cfg-single-group` OK, `cfg-loader-log` OK, `cfg-partial-effectuate` OK.

The damaged-log path was exercised on real hardware. Five consecutive `cfg-partial-effectuate` runs on `c64u` gave 2 clean logs and 3 interleaved ones, each of the latter reported as SKIP with the line quoted, for example:

```
effectuated 0, left clean 21, unreadable 1
unreadable log line: "ttings of store 'Audio Mixer' after loading."

effectuated 1, left clean 20, unreadable 1
unreadable log line: "Store 'SID Sockets Configuration' is clean afterAccept client 0 on socket 7.  192.168.1.185:36902"
```

Before this change the first of those reported `the loader effectuated 0 store(s)` as a failure and the second silently lost a store from the clean list.

`deferred-machine-actions` on `c64u`: overlay OK 10 checks in 3 runs of 3; freeze OK 10 checks in 3 runs of 4, and FAIL in 1 run of 4 on the last check, `one menu button reopens the menu straight after a close`, with `one PUT /v1/machine:menu_button sent straight after the close did not open the menu within 8s, so the press was dropped`. That check is left failing.

The device-free gates pass: `make lint_test` (ruff over `tests` and `run-tests`), `registry_test` (64 suites), `stale_gates_test`.

The `cfg-loader-log` suite is what proves the log-parser fix, because the interleaving is not reproducible on demand. Reverting `store_name` to the previous prefix-only parse makes its check 3 fail with `effectuated: got ['Audio MAccept client 0 on socket 7.  192.168.1.185:49362'], expected []`.

## Alternatives considered and rejected

Naming the C64 Ultimate's preset fields in `machine.py`, beside the other `search_*` properties. That would have fixed the failure and matched the existing pattern, but it puts one service's current catalogue into the harness, so the next service change or the next machine reintroduces the same failure. The fields are discovered instead, and `machine.py` keeps only navigation: `search_service`, `search_menu_entry`, `search_in_launcher`, `search_form_title`, `task_menu_key`.

Skipping the absent preset fields in the dropdown scenario. It would have turned a red suite green without covering anything on a C64 Ultimate, and a scenario that silently covered no field looks the same as one that covered every field.

Discarding the unreadable loader log lines. That is what makes the interleaving invisible, and an interleaved log would then read as a load that touched fewer stores than it did, which is the failure `cfg_partial_effectuate_test` exists to detect.

Rejoining the two halves of a spliced loader record by concatenating the following line. Nothing guarantees the two halves are adjacent, and more than two tasks can interleave, so the reconstruction would be a guess presented as a measurement.

Tagging the `deferred-machine-actions` screen-cleanliness checks with the firmware fix. They pass on `u64` under `--assume-fix`, so tagging them would report SKIP for behaviour the machine satisfies and would make `tools/stale_gates.py` read a forced pass as the fix having landed.

Failing rather than skipping the CommoServe dropdown-confirm checks. The defect empties every field on the form, so the failure would cascade into every later check in that scenario and stop the suite before the scenarios that follow it, which have never run on a C64 Ultimate. The `FIXES` entry reports it as a named SKIP on that machine and keeps it running everywhere else.

Letting the dropped menu button fail the deferral checks at their setup. It did, in 2 freeze runs of 3, and those runs then measured nothing about the deferral. The defect is asserted by its own check at the end of the suite instead, and the steps that only need the menu up retry the button.

Covering the `.crt` launch in the new suite. It needs a cartridge image fixture on the device and leaves the machine running foreign code, which no BASIC-prompt assertion could then read.

### DIFF
--- a/run-tests
+++ b/run-tests
@@ -320,6 +320,13 @@ SUITES: Sequence[Suite] = (
           "-H @HOST@ -p @PASS@ --mode @MODE@"),
     Suite("e2e", "cfg-single-group", "tests/e2e/filemanager/cfg_single_group_test.py",
           "-H @HOST@ -p @PASS@ -t @TIMEOUT@ --mode @MODE@", profile=profiles.QUICK),
+    # Needs no device: it reads the loader's debug log parser over recorded
+    # text. The device writes that log with an unlocked, character-at-a-time
+    # printf, so another task's line splices into the middle of a loader line;
+    # the splice is not reproducible on demand, which is why the case is a
+    # recorded one rather than a device run.
+    Suite("e2e", "cfg-loader-log", "tests/e2e/filemanager/cfg_loader_log_test.py",
+          "", profile=profiles.SMOKE),
     # Manual: loading a partial .cfg is supposed to leave the stores it does
     # not name alone, and that no longer holds. Measured on a C64 Ultimate
     # 1.2.0, a file naming one store listed nineteen in its loader
--- a/run-tests
+++ b/run-tests
@@ -439,6 +439,17 @@ SUITES: Sequence[Suite] = (
     # fix and green at about 10x after it.
     Suite("e2e", "fast-reset", "tests/e2e/io/c64/fast_reset_test.py",
           "-H @HOST@ -p @PASS@ -t @TIMEOUT@"),
+    # Guards GideonZ/commodore#1 (c66ef028): a Reset, a Reboot or a C=+R taken
+    # from the menu has to run after the UI has torn down, so the C64 comes
+    # back to a boot screen with none of the menu's characters on it. The
+    # other half of that change, the menu opening again afterwards, needs the
+    # hardware button's edge detector, which no route this harness has can
+    # reach; those three checks report SKIP with the measurement. It resets
+    # and reboots the machine several times, which is why it is not in the
+    # shallow profiles.
+    Suite("e2e", "deferred-machine-actions",
+          "tests/e2e/io/c64/deferred_actions_test.py",
+          "-H @HOST@ -p @PASS@ -t @TIMEOUT@ --mode @MODE@"),
     # Reproduces #733: with the FPGA decoder fix in the core, "mirroring off" is
     # the regression guard. Runs last, because a device whose core predates that
     # fix stalls the whole Nios on this scenario and needs a JTAG recovery.
--- a/tests/README.md
+++ b/tests/README.md
@@ -268,8 +268,10 @@ Profiles, shallowest first. A suite runs from its own profile up.
 
 e2e:
   assembly64                        .      x      x      x      x
+  av-stream                         .      .      x      x      x
   browser-filesystem-refresh        .      .      x      x      x
   browser-long-filename             .      x      x      x      x
+  cfg-loader-log                    x      x      x      x      x
   cfg-partial-effectuate            .      .      .      x      x
   cfg-single-group                  .      x      x      x      x
   cfg-unknown-items                 .      .      x      x      x
@@ -277,12 +279,15 @@ e2e:
   create-disk-image                 .      x      x      x      x
   doom-release                      .      .      .      x      x
   esp-depends                       x      x      x      x      x
+  fast-reset                        .      .      x      x      x
   freeze-menu                       .      .      x      x      x
   ftp-client                        .      x      x      x      x
   ftp-server                        x      x      x      x      x
+  ident-service-switch              .      .      x      x      x
   input                             .      x      x      x      x
   input-batching                    x      x      x      x      x
   key-injection                     .      .      x      x      x
+  lint                              x      x      x      x      x
   machine-code-monitor              .      .      .      x      x
   menu-screen                       x      x      x      x      x
   navigation-keys                   x      x      x      x      x
@@ -294,15 +299,19 @@ e2e:
   prg-load-path-trim                .      x      x      x      x
   printer                           .      x      x      x      x
   readmem-writemem                  x      x      x      x      x
+  registry                          x      x      x      x      x
   rest-api-coverage                 .      x      x      x      x
   reu-turbo                         .      .      x      x      x
   runner-policy                     .      x      x      x      x
+  stale-gates                       .      x      x      x      x
   telnet-drain                      .      x      x      x      x
   telnet-stale-session              .      .      .      x      x
+  telnet-sustained-input            .      .      .      x      x
   temp-auto-cleanup                 .      .      x      x      x
   transport-usage                   x      x      x      x      x
   uci-net-target                    .      .      .      x      x
   uci-targets                       .      .      x      x      x
+  ui-backend-parse                  x      x      x      x      x
   ui-backend-smoke                  x      x      x      x      x
   usb-bulk-out-integrity            .      .      .      x      x
   wake-on-wifi                      .      .      .      x      x
@@ -327,8 +336,8 @@ soak:
   usb-keyboard-repeat               .      .      .      x      x
 
                                ------ ------ ------ ------ ------
-  suites                            8     21     43     52     52
-  suite runs                        8     21     43    104    156
+  suites                           12     26     51     63     63
+  suite runs                       12     26     51    126    189
 
 Scenario and check counts, and durations, are not shown here:
 the registry does not know them. They depend on the machine and are
--- a/tests/README.md
+++ b/tests/README.md
@@ -277,6 +277,7 @@ e2e:
   cfg-unknown-items                 .      .      x      x      x
   cfg-whitespace                    .      .      x      x      x
   create-disk-image                 .      x      x      x      x
+  deferred-machine-actions          .      .      x      x      x
   doom-release                      .      .      .      x      x
   esp-depends                       x      x      x      x      x
   fast-reset                        .      .      x      x      x
@@ -336,8 +337,8 @@ soak:
   usb-keyboard-repeat               .      .      .      x      x
 
                                ------ ------ ------ ------ ------
-  suites                           12     26     51     63     63
-  suite runs                       12     26     51    126    189
+  suites                           12     26     52     64     64
+  suite runs                       12     26     52    128    192
 
 Scenario and check counts, and durations, are not shown here:
 the registry does not know them. They depend on the machine and are
--- a/tests/e2e/filemanager/cfg_loader_log_test.py
+++ b/tests/e2e/filemanager/cfg_loader_log_test.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python3
+# Gate check: the .cfg loader's debug log is read correctly when it interleaves.
+
+"""Verify parse_loader_log over a recorded debug log, intact and spliced.
+
+The device writes its debug log with printf, which emits one character at a
+time through outbyte() and takes no lock (software/system/small_printf.cc and
+software/application/ultimate/ultimate.cc). A task switch part way through a
+line therefore splices another task's output into the middle of it. The REST
+poller that every overlay-mode suite runs prints "Accept client 0 on socket 7."
+for each connection, so it is the task that lands in the loader's lines.
+
+That is not reproducible on demand, so it is checked here against the text
+measured on a C64 Ultimate 1.2RC rather than by driving a device. The intact
+log below was captured from an Ultimate 64 Elite on firmware 3.15, harness
+commit b3c76093.
+
+Needs no device.
+"""
+
+import sys
+from pathlib import Path
+
+# The one stanza that puts the shared library on sys.path; see tests/lib/bootstrap.py.
+sys.path.insert(0, str(next(p for p in Path(__file__).resolve().parents
+                            if (p / "tests" / "lib").is_dir()) / "tests" / "lib"))
+import bootstrap  # noqa: E402,F401
+
+sys.path.insert(0, bootstrap.directory("e2e", "filemanager"))
+
+from report import (  # noqa: E402
+    Failure, add_colour_argument, apply_colour, check, suite_fail, suite_ok)
+from selftest import expect  # noqa: E402
+
+from cfg_single_group_test import parse_loader_log  # noqa: E402
+
+NAME = "cfg_loader_log_test"
+
+# Two records and the surrounding noise, as an Ultimate 64 Elite wrote them.
+INTACT = """\
+Accept client 0 on socket 7.  192.168.1.185:50898
+HTTP GET /v1/machine:menu_screen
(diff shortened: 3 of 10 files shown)


chrisgleissner:
Hi @GideonZ ,

these errors came out of the most recent run of the E2E tests against the c64u changes we discussed. This also fixes an issue in a recently added E2E which did not work with the C64U.

It would be good to merge them for 3.15. Especially since there are no prd code changes.

Thanks
Christian

PR #872, opened 2026-09-07
Mute U64II speaker SIDs during freeze

### LINKED ISSUE #868: Silence U64E2 / C64U freeze pop sound
Reported by Gideon: on the C64U the speaker still emits a low frequency clicking/popping sound when the menu is entered and left. The same is true on the U64E2. On the Ultimate 64 it does not.

## Previous behaviour

The freezer silences the SIDs while it owns the machine by writing the SID master volume at `$D418`. That register also sets the DC offset of the SID output, so each write steps the offset and is heard as a low click. Measured on an Ultimate 64 over the device's own audio stream: peak sample 19000 against a silent-machine floor of 19.

## Implemented change and the restriction in it

`freezer_mute_sids()` and `freezer_unmute_sids()` in `software/io/c64/c64.cc` keep the `$D418` writes on every target. Where an FPGA audio mixer exists, the mixer is closed around those writes, so the offset step happens where it cannot be heard. `u64_mute_sids()` zeroes the eight SID channel bytes at `U64_AUDIO_MIXER`; `u64_unmute_sids()` reprograms the mixer from the stored settings through `U64Config::restoreMixer()`, because the mixer registers are write-only and read back as zero.

The mixer close is guarded by `#if U64 == 1`:

- `software/io/c64/c64.cc:831-837` (the weak declarations)
- `software/io/c64/c64.cc:845-851` (mute)
- `software/io/c64/c64.cc:858-864` (unmute)

So on `U64 == 2` builds, which are the U64E2 and the C64U (`target/u64ii/riscv/ultimate/Makefile` defines `-DU64=2`), only the `$D418` writes run and the click is audible.

There is no hardware reason for the restriction. `software/u64/u64_config.cc` is compiled for both `U64 == 1` and `U64 == 2`, and `U64_AUDIO_MIXER` is the same address on both (`software/system/u64.h:16`). When the change was made I only had an Ultimate 64 to measure on, so I scoped it to that target rather than changing behaviour on hardware I could not verify. That was too cautious.

## What needs to change

1. Widen the guard from `#if U64 == 1` to `#if U64` in `software/io/c64/c64.cc`, so `u64_mute_sids()` and `u64_unmute_sids()` are used on `U64 == 2` as well.

2. The U64E2 and C64U have a second mixer for the internal speaker, at `U64_SPEAKER_MIXER` (`software/system/u64.h:17`), programmed by `U64Config::setSpeakerMixer()` with 20 bytes. Gideon's suggestion is to mute it inside the existing mute function, guarded by `#if U64 == 2`. Zeroing the ten SID-channel bytes there matches what `u64_mute_sids()` already does for the main mixer.

3. On the unmute side, check whether anything extra is needed. `U64Config::restoreMixer()` calls `mixercfg.effectuate_settings()`, and on `U64 == 2` `setMixer()` already calls `setSpeakerMixer()` at the end (`software/u64/u64_config.cc:1336-1339`). If that path is reached, the speaker mixer is restored with the main one. If it is not, `speakercfg.effectuate_settings()` has to be called as well. Either way, a speaker that the user has disabled through `CFG_SPEAKER_EN` must stay silent after the unmute.

## How to verify

Enter and leave the menu on a C64U and on a U64E2, with the internal speaker enabled, and confirm no click is audible. The Ultimate 64 measurement used the device's own audio stream and compared the peak sample against a silent-machine floor; the same method works on the other two targets for line-out. The speaker itself has to be judged by ear, or (optionally) using a USB mic listening to it.

Both mixers are write-only, so a test cannot read back what was programmed. Restoring settings after the unmute has to be checked by confirming that the configured volumes are in effect again, not by reading the registers.

## Request

@barryw would you be able to pick this up? The change is small: widen one preprocessor guard and add the speaker mixer to the existing mute and unmute. I am happy to review, and the measurement method above is what I used on the Ultimate 64.

Introduced in ba4ccb71 ("Mute the SIDs in the FPGA mixer where there is one").

barryw:
U64E2 and C64U builds skipped the FPGA mixer mute path when entering the freezer, leaving SID master-volume writes audible as clicks. Enable the existing mixer helper for all U64 targets and, on U64II, mute the eight SID inputs in both the main and internal-speaker mixers. The existing restore path reapplies both mixers from configuration, including a disabled speaker.

Fixes #868.

Validation:

- `make u64ii` on Beast, including ESP-IDF control firmware and both U64E2 size checks
- `make u64_no_esp` on Beast as an original-U64 regression build
- OpenAPI validation and 36 repository tests pass in both builds
- Linked U64II code writes all eight SID inputs in both mixer banks; linked U64 code retains its existing single mixer-bank writes

The audible result still needs confirmation on a U64E2 or C64U. @GideonZ @chrisgleissner, could one of you verify entry and exit from the freezer with the internal speaker enabled?

### DIFF
--- a/software/io/c64/c64.cc
+++ b/software/io/c64/c64.cc
@@ -828,7 +828,7 @@ void C64::dma_transfer_frozen(uint16_t offset, uint8_t *buffer, int length, int 
     C64_DMA_MEMONLY = saved_memonly;
 }
 
-#if U64 == 1
+#if U64
 // Defined in u64_config.cc, which owns the FPGA audio mixer and the settings
 // it is programmed from. Weak because the updater application links this file
 // without that one; there the SID writes below stand on their own.
@@ -836,13 +836,12 @@ extern void u64_mute_sids(void) __attribute__((weak));
 extern void u64_unmute_sids(void) __attribute__((weak));
 #endif
 
-// Silence the SIDs while the freezer owns the machine. The SID master volume
-// is what does it, on every target. Where an FPGA mixer is available it is
-// closed around those writes, so the click they make is not heard; see
-// u64_mute_sids().
+// Silence the SIDs while the freezer owns the machine. Where an FPGA mixer is
+// available, use it instead of stepping the SID master-volume DC offset. Other
+// targets retain the portable SID-register fallback.
 static void freezer_mute_sids(void)
 {
-#if U64 == 1
+#if U64
     if (u64_mute_sids) {
         u64_mute_sids();
         return;
@@ -855,7 +854,7 @@ static void freezer_mute_sids(void)
 
 static void freezer_unmute_sids(void)
 {
-#if U64 == 1
+#if U64
     if (u64_unmute_sids) {
         u64_unmute_sids();
         return;
--- a/software/u64/u64_config.cc
+++ b/software/u64/u64_config.cc
@@ -1312,9 +1312,9 @@ int U64Config :: setFilter(ConfigItem *it)
 // rather than instead of them: the $D418 mute stays, and stays portable, and
 // on this hardware it happens where it cannot be heard.
 //
-// The first eight bytes are the SID channels: UltiSID 1 and 2, socket 1 and 2,
-// right and left. SetMixerAutoSid zeroes the same eight to mute the SIDs while
-// it remaps them.
+// The first eight bytes of either mixer are the SID channels: UltiSID 1 and 2,
+// socket 1 and 2, right and left. SetMixerAutoSid zeroes the same eight in the
+// main mixer while it remaps them.
 void u64_mute_sids(void)
 {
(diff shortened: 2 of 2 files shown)


chrisgleissner (approved):
Looks good.

chrisgleissner:
@Retro1968 Hi there, would you be interested in testing this change on your machine? The changes are minima, thus there is no risk to your machine,

If it works, you should not hear any sound as you enter or leave the menu in Freeze mode, i.e. when the menu takes up the entire screen. 

If it does not work, you should hear a "pop"-like sound as you enter the menu, and another such sound as you leave the menu.

@GideonZ Maybe you can assess just by looking at the code whether this will work, and if it does, it would be good to merge.

chrisgleissner:
@GideonZ @barryw I tested this on the C64U and it is now completely silent on entering and leaving freeze mode.

Retro1968:
If there is still a test needed on UE2 please give me a downloadlink.

Issue #877, opened 2026-09-08
IEC Device: Partition table wrong


markusC64:
# IEC Device: Partition table wrong

The partition table of the IEC device is wrong.

It currently shows the path to the root as the partition name instead of the actual partition name. It also shows `DIR` as the type instead of the actual type:

* `NAT` for DNP images and directories on the FAT filesystem
* `41{space}` for 1541 partitions
* `71{space}` for 1571 partitions
* `81{space}` for 1581 partitions

The type for DNP/1541/1571/1581 partitions is based on the image used as the root.

Firmware version: 3.15
Device: U64 II, but probably all other Ultimates as well.

chrisgleissner:
Thanks @markusC64 ,

I am having a look at this.

Best
Christian

chrisgleissner:
Hi @markusC64 ,

the three issues you reported yesterday should be fixed now in https://github.com/GideonZ/1541ultimate/actions/runs/34338550102

Could you please retest?

Thanks
Christian

markusC64:
@chrisgleissner  Thanks. I will test it after work. Now in the lunch break I have just done a very quick test: Your patch has a regression. For brevity, I am using S-Jiffydos syntax:

```
a$="cP"+chr$(13)
@a$
```

is now bricked.

chrisgleissner:
@markusC64 Thanks. 

Just to avoid ambiguity, what is the expected result of the command you gave, and what is the actual result you experience? The term 'bricked' is very vague and could also mean your device has become completely unresponsive. :-)

Also, I will await more test results from you to ensure this can be fixed comprehensively rather than peace meal.

markusC64:
The expected result is that partition 13 is selected.

The actual result is that the partition number is lost, because `CHR$(13)` is the binary partition number here and must not be stripped. The `cP` command, in contrast to the `cp` command, expects the partition number in binary. Thus we get a syntax error.

`cP` is a well-known CMD command; the partition number is a binary byte, not ASCII text. Partition 13 is a valid partition number.

So my "bricked" was not meant to imply that the device itself becomes unresponsive — just that this command no longer works.

I am doing more testing soon.

chrisgleissner:
Hi @markusC64,

Thanks for the clarification. This issue is now fixed in #881, on top of the merged #878.

[...]

Issue #876, opened 2026-09-08
IEC Device: Block access incompatible


markusC64:
**Describe the bug**
"view bam" from the test demo disk of the 1541 is not working.

**To Reproduce**
1. Change device id of the iec device to 8.
2. Make the test demo disk image (d64) the root of a partition.
3. load "view bam" and execute

**Expected behavior**
The BAM should be decoded and displayed. But every block is shown as occupied which is not true. And at the end, it's not wairting for a key because of errors in processing.

**Applies to (please complete the following information):**
 - Hardware type: U64 II and probably all other
 - Firmware Version 3.15 Release(!)

chrisgleissner:
Thanks @markusC64 ,

I am having a look at this.

Best
Christian

chrisgleissner:
Hi @markusC64 ,

can this be closed?

Thanks

markusC64:
No, we cannot close that. BAM VIEW is still not working correctly.

markusC64:
Ok,. BAM VIEW might have an error, DISPLAY BAM from the 1570 TEST DEMO Disk is working better...

Issue #875, opened 2026-09-08
IEC Device broken: Command Channel


markusC64:
**Describe the bug**

**To Reproduce**
create an "os" directory in the root of the partition by the file browser. Execute
open 15,10,15
print#15,"cd//os"
and read the error channel. It fails with 71, directory error.

**Expected behavior**
change into that directory.

**Applies to (please complete the following information):**
 - Observed on ultimate 64 ii, but should be an issue on all supported ultimates
 - Firmware Version 3.15

chrisgleissner:
Thanks @markusC64 ,

I am having a look at this.

Best
Christian

markusC64:
Works now.

PR #878, opened 2026-09-09
Fix the Software IEC command channel for the bytes a Commodore sends

chrisgleissner:
The Software IEC drive mishandled the bytes a Commodore puts on the bus, rather than the commands themselves. `PRINT#` ends every command with a carriage return, and BASIC prints a space before and after every number. The command channel counted the carriage return as part of the command, and the block command parameter parser could not step over the colon in the usual `PRINT#15,"U1:";c;d;t;s` idiom.

Refs #875, #876, #877.

## The three reported defects

**Command channel (#875).** `PRINT#15,"CD//OS"` sends `CD//OS` and a carriage return. The firmware looked for a directory whose name ends in a carriage return and answered `71, DIRECTORY ERROR`. Every text command was affected, not only `CD`: `MD` created directories with a carriage return in the name, and the relative file position command read a record number one byte too wide. One trailing carriage return is now removed before parsing. Only the last byte is removed, so a command carrying binary parameters keeps them.

**Block access (#876).** `B-R`, `B-W`, `B-P`, `B-A`, `B-F`, `U1` and `U2` read their parameters with `sscanf("%d%d%d%d", ...)`. `PRINT#15,"U1:";2;0;18;0` arrives as `U1: 2  0  18  0 `, so the colon was taken as channel 0 and every value moved one place, which made the track number arrive as the partition number and answer `77, SELECTED PARTITION ILLEGAL`. `B-P:2,144` did not fail at all: it silently positioned channel 0 at offset 2. No format string fixes this, because `%d` skips leading whitespace and nothing else. CBM DOS separates these parameters with a space, a comma or a cursor right and allows one colon after the command word, which is what the ROM routine at $CC6F does; a small parser here does the same.

**Partition directory (#877).** `$=P` listed the path each partition is rooted at instead of its name, and typed every entry `DIR`. The name now comes from the partition and the type from the file system at its root: `NAT` for a host directory and for a DNP image, `41 `, `71 ` or `81 ` for a mounted image. The type strings and the three character field follow CMD DOS, as reproduced byte for byte in VICE's CMD hard disk emulation.

## The scanner underneath

`software/system/small_printf.cc` provides the firmware's own `sscanf`, and it had three defects. It counted a conversion whether or not it made one, so checking the return value told a caller nothing. It ignored literal characters, so `"%d/%d"` did not require a slash. It had no `%c`, and the default branch stored a whole `int` through the caller's pointer, which for a `char` is three bytes past the end.

That is the root of the time stamp defect listed below. It is fixed at the root rather than worked around: the alternative was a hand written date parser in `cbmdos_parser.cc`, which would have left the same scanner defect in place for the FTP server, the FTP client, the modem, the configuration reader and the poke dialog. `_vscanf` now follows the rules its callers were already written against. A conversion that does not happen is not counted and stores nothing, a literal has to match, and whitespace in the format matches any run of it including none. The date is parsed with a format string again, which removes 61 lines from `cbmdos_parser.cc`.

Blast radius, checked caller by caller. Three passed a destination they never initialised and used to receive zero for input that does not parse. Two now report it: the configuration reader returns `CFG_LINE_MALFORMED` for a value that is not a number, and the U64 poke dialog refuses an entry that is not an address and a value. The modem keeps the zero, because all six of its uses have always taken a missing number that way and changing them would alter AT command handling that nothing here tests. The remaining sixteen live callers either check the return value, and only improve, or already initialise what they pass. `%s` still has no field width; no caller in this firmware uses it.

A Linux build binds `sscanf` to the C library rather than to this file, so no existing host suite can see the defect or the fix. `target/pc/linux/scanf` declares `sscanf` itself and checks the scanner directly, including a guard that catches the write past a `char`.

## Further defects found while testing

Found by sweeping the documented command surface, not from the reports, and all in the same command channel.

- A command that filled the 64 byte buffer was dropped silently. The terminating zero landed on the low byte of the byte count and set it to zero. The buffer is one byte longer.
- `CD:_` and `CD/path/:_` were refused with `71, DIRECTORY ERROR`. The left arrow means the parent directory behind a colon as well as directly after the command word.
- `U9` and `U:` were refused. CBM DOS takes the user command from the low four bits of the character after the U, so `U9` is `UI` and `U:` is `UJ`.
- `UI+` and `UI-` reset the drive and rewrote the error channel to the power-up message. On a real drive they only select bus timing, so they now do nothing.
- The position command refused the documented `PRINT#15,"P"CHR$(96+2)...` form. CBM DOS masks that byte to its low four bits when it is 19 or more, so both spellings reach the same channel and a byte just outside the channel range is still refused.
- A buffer position above 255 was refused. CBM DOS keeps only the low byte, so it wraps.
- A listing filtered by a time stamp, such as `$:*=>01/02/25 03:04 PM`, answered `30, SYNTAX ERROR` on the device for every threshold, because the format string asks for `%c`. Fixed by the scanner above.
- `T-RA`, `T-RI`, `T-RD` and `T-RB` reported the wrong year. The clock counts years from 1980, the way a FAT time stamp does, and `get_current_time` passed that number to callers that print a calendar year. On 9 September 2026 the drive answered `0046-09-09` and now answers `2026-09-09`. `T-RA` also prints the month before the day, matching its own comment and every other date this drive prints.

## Test approach

Four layers, because the defects sit at four depths. The first three need no device and finish in well under a second, so they carry the bulk of the coverage.

| Layer | What only it can reach |
| --- | --- |
| `target/pc/linux/scanf` | The firmware's own scanner, which no other host target links |
| `target/pc/linux/parse` | What each command was parsed into, not just that it returned no error |
| `target/pc/linux/iecdrive` | Whole drive behaviour against real sectors, directories and five partitions formatted during the run |
| `tests/e2e/io/iec/dos_command_test.py` | The bytes the KERNAL actually puts on the bus, and the two checks that depend on the firmware's scanner |

The hardware layer exists for two things a host cannot reach. These defects are about what a Commodore sends, and a harness that builds command strings by hand reproduces none of them, which is why the existing host regression in `software/test/iecdrive` passed throughout. A host build also links the C library's `sscanf`, so the time stamp filter and the reported year cannot fail there at all. Everything else is checked on the host, where it is cheap.

The parser program now returns a failing exit status. It did not before, which is why one of its own cases had been failing unnoticed. Both hardware suites share one KERNAL agent: `rel_agent.asm` became `iec_agent.asm`, gained a fixed count read and a separate secondary address, because the KERNAL refuses logical file number 0 and a directory needs secondary address 0.

## Why the hardware suites cost what they do

The agent waits blind after each transaction and only then reads the result. A REST memory read halts the C64 for the duration of the copy, and one landing inside a serial transfer corrupts it: polling the mailbox every 20 ms through a 254 byte read spoiled one read in thirty here. There is no cheap signal that a transaction has finished, so the wait has to be an estimate of the transfer time. It used to be a flat two seconds whatever the transaction carried, and is now a fixed part per operation, between 0.08 s and 0.22 s, plus 1.8 ms a byte, plus 1.2 s for a transaction addressed to an emulated drive, which seeks and reads at the speed of the hardware it emulates.

Those numbers are measured, and the floor is real. Profiling a full run put 84% of it in the wait, against 3.8 s of REST and 0.8 s of FTP, so nothing else was worth attacking. Scaling every budget together and rerunning the suite gave 40.5 s at 1.0, 27.2 s at 0.6, 20.3 s at 0.4 and 17.1 s at 0.3, all clean; 0.22 overran three transactions and 0.16 overran five, and both were slower overall, because an overrun costs two seconds of grace. The shipped values keep a little under twice that margin. The emulated drive allowance was measured the same way: 1.8 s was clean at 206 s, 0.8 s cost 157 s and left one transaction short, so 1.2 s is what is here.

Shrinking the payloads does not help either. `rel-copy` moves about four kilobytes per file system, so the byte term is a small part of its run; the per operation part and the emulated drive allowance are the rest. Its record lengths of 1, 31, 164 and 254 are the boundary values under test, 254 being the largest a REL file allows. Merging REST round trips would save under a second. Nothing further comes off without costing coverage.

A transaction that needs longer still gets two seconds of grace, and each suite reports how many needed it, so a wait that is too tight becomes a visible number rather than a rare corrupted read. Neither run below needed it. The rate is a property of the Commodore's serial routines rather than of this machine, so it should carry to other Ultimates.

| Suite | Before | After |
| --- | --- | --- |
| `iec-dos-commands` | not present | 24 s |
| `rel-copy` | 643 s | 171 s |

Both suites are tagged `standard`, so they run in the merge gate and not in the default `run-tests` run, which is `quick`. The gate is about seven minutes shorter than before despite gaining a suite, because `rel-copy` was already in it at 643 s.

## Validation

| Check | Result |
| --- | --- |
| `target/pc/linux/scanf` | Pass. Twelve checks fail on the old scanner, including the guard for the write past a `char` |
| `target/pc/linux/parse` | Pass. Each behaviour change was proved red first |
| `target/pc/linux/iecdrive` | Pass in 40 ms. Ten separate reverts of one change each turn at least one new check red |
| Hardware, this branch with only its production changes reverted | Fails in all four areas: `71, DIRECTORY ERROR` on `CD//<dir>`, `77, SELECTED PARTITION ILLEGAL` on `U1: 2  0  18  0 `, `T-RA` reporting year 46 for a file stamped 26, and the partition listed as `/TEMP/` with type `DIR`. A time stamp filter answers `30, SYNTAX ERROR` on the same build |
| Hardware, this branch | `iec-dos-commands` 9 checks in 24 s, three consecutive runs at the same time. The partition lists as `RAMDISK` with type `NAT` |
| `rel-copy`, this branch | 55 checks in 171 s and 182 s over two runs. No transaction needed the grace period in either |
| Builds | `u64` and `u64ii` build clean. `u2` was not built here: its recipe runs `git show` inside the build container, which cannot resolve a git worktree |

The hardware runs used an Ultimate 64 Elite, FPGA 125, core 1.4F, unique id 38C1BA, over wired Ethernet, with both firmware images loaded over JTAG rather than flashed. The suite restores the Software IEC settings and working directory and removes only its own fixtures.

```sh
./run-tests 192.168.1.13 -s iec-dos-commands --no-retry -o runs/iec-dos
```

`tests/e2e/io/iec/README.md` describes both suites, the agent they share, and the timing above.

## Not covered

`B-R` still exposes the whole 256 byte block from its first byte, as `U1` does. On a real drive `B-R` takes the count of valid bytes from the first byte of the sector and starts at the second. Changing that would alter behaviour existing software may rely on, so it is left alone and the drive level suite deliberately does not assert what `B-R` leaves the buffer pointer at.

The partition types `41 `, `71 ` and `81 ` were checked on the host against images formatted during the run; on hardware only `NAT` was checked. The type comes from the file system object alone, with no device specific behaviour in that path. The modem and the poke dialog were verified by reading them, not on hardware.

`N` (format) remains a stub that answers `00, OK` without doing anything, and `V` (validate) is not implemented. Neither was touched. Only the Ultimate 64 Elite was tested on hardware.


chrisgleissner:
Hi @GideonZ ,

this is ready for review. Please see the PR description above for details. In case it is too long, I can rewrite, but I rather keep all details in there.

Also, I deliberated for a while whether to fix the root cause of some of these issues (the scanner), but opted for a clean root cause fix instead of some workarounds. 

Thanks
Christian

PR #880, opened 2026-09-09
Stop REST fire2/fire3 leaking onto the other joystick port

### LINKED ISSUE #879: U64 REST Joystick input 2nd/3rd firebutton triggering on both joystick ports
**Describe the bug**
When using the REST call /v1/machine:input with joystick and 'fire2' or 'fire3' it triggers 2nd/3rd fire button on both joystick port 1 and port 2.  first button and directions work fine - only on the joystick port that is selected in rest call.

**To Reproduce**
1.call /v1/machine:input with 'fire2' or 'fire3' as direction

**Expected behavior**
only selected joystick port in rest call should be triggered

 - Hardware type: U64
 - Firmware Version 3.15 official

chrisgleissner:
## Summary

Fixes #879.

`JoystickOutput::apply()` mirrors an asserted-low POTX/POTY value into port 1's paddle registers whenever port 2's `fire2`/`fire3` is pressed, and OR's port 2's extra-button state into port 1's mouse-enable line. Holding `fire2` or `fire3` on one REST joystick port also reads as pressed on the other port's Anykey-style POT read, matching #879. The fix removes the mirror and the OR term; each port's paddle registers now reflect only that port's own `fire2`/`fire3` state.

A second defect: `apply_joystick_event()` releases an input by updating persistent state only. A `tap` on the same input that's still mid-hold keeps its overlay bit pressed, which outranks the new persistent state in `JoystickOutput::outputSnapshot()`'s `persistent & overlay` combination, so the release has no visible effect for up to one 20ms timer tick. `JoystickOutput::cancelRestPort1Overlay`/`cancelRestPort2Overlay`, called from the release branch, clear the released bit's overlay and hold countdown immediately.

## Changes

- `joystick_output.cc`/`.h`: removes the cross-port mirror and mouse-enable OR term; adds `cancelRestPortXOverlay` + `cancel_overlay_bits`; replaces a literal `7` with a named `JOYSTICK_BUTTON_COUNT` constant matching `input_api.h`'s `INPUT_API_MAX_JOYSTICK_INPUTS`.
- `route_input.cc`: calls the new cancel methods on release.
- `input_api_state_test.cpp`: two host unit tests for the release/tap fix.
- `input_test.py`: Anykey-isolation checks collapsed into one parametrized loop; added a check holding `fire2` on port 1 and `fire3` on port 2 at once (a mirror can't satisfy this by chance the way it can a one-port-at-a-time check); added hardware checks for tap auto-release and for release-in-the-same-batch-as-tap; extended the round-trip and `release_all` checks to assert POT hardware, not just REST JSON and CIA state; `read_joystick_pots` now pauses the C64 like its siblings (unpaused, the KERNAL's keyboard-scan IRQ races the CIA1 write within ~20ms); all three pause-then-read helpers now attempt `resume()` even if `pause()` itself raises; paired-port checks now pause once via `read_joystick_pots_both` instead of twice.

## Test plan

Both defects verified red/green on an Ultimate 64 Elite (firmware 3.15, real hardware):

- Cross-port mirror: reverting only `joystick_output.cc` reproduces #879's exact symptom (`expected fire2/fire3=False/False, got True/False`); fixed, `joystick` 21/21, `contract` 7/7, `keyboard` 19/19 pass.
- Release/tap race: disabling only the two `cancelRestPortXOverlay` calls fails the new same-batch check (response still shows `fire2` pressed); restored, it passes. Also covered by 2 host unit tests (`cd software/api/tests && make input-api-state`, 30/30 pass).

`u64` and `u64ii` both build cleanly with the full change set.

## Related, not fixed here

`usb_hid.cc`'s mouse-move handler unconditionally writes `C64_PADDLE_1_X`/`C64_PADDLE_1_Y` from raw USB mouse position right after `JoystickOutput::apply()` writes the same registers from REST state, so an attached USB mouse clobbers a REST-held port-1 `fire2`/`fire3`. Same register family, a different, uncoordinated writer. No USB mouse was available to verify a fix red/green.

### DIFF
--- a/software/api/route_input.cc
+++ b/software/api/route_input.cc
@@ -268,8 +268,11 @@ static void apply_joystick_event(const InputParsedEvent &event)
         active_low |= event.joystick_mask;
         if (event.port == 1) {
             JoystickOutput::instance().setRestPort1Persistent(active_low);
+            // Cancel any in-flight tap so it can't mask this release.
+            JoystickOutput::instance().cancelRestPort1Overlay(event.joystick_mask);
         } else {
             JoystickOutput::instance().setRestPort2Persistent(active_low);
+            JoystickOutput::instance().cancelRestPort2Overlay(event.joystick_mask);
         }
         break;
     case INPUT_PARSED_TAP:
--- a/software/api/tests/input_api_state_test.cpp
+++ b/software/api/tests/input_api_state_test.cpp
@@ -812,6 +812,43 @@ TEST(RestJoystickStateTest, Fire2AndFire3PersistAndReleaseIndependently)
     EXPECT_EQ(0x5F, port1);
 }
 
+TEST(RestJoystickStateTest, ReleaseCancelsInFlightTapOverlayImmediately)
+{
+    // Tap then release of the same input, as route_input.cc applies them.
+    reset_joystick_output();
+    uint8_t hold[7] = { 0, 0, 0, 0, 0, 1, 0 };  // fire2 (bit 5) tapped
+    uint8_t port1 = 0;
+    uint8_t port2 = 0;
+
+    // Tap shows fire2 pressed.
+    JoystickOutput::instance().armRestPort1Overlay(0x7F & ~0x20, hold);
+    JoystickOutput::instance().snapshot(port1, port2);
+    EXPECT_EQ(0x5F, port1);
+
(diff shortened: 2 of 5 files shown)


chrisgleissner:
Confirmed by user to work. Merged.

PR #881, opened 2026-09-09
Software IEC: C64 OS / CMD DOS compatibility, crash fixes and logs

### LINKED ISSUE #890: $=P command adds a blocks free row when using SoftIEC
When listing the partitions with load"$=P",X, the partition list contains an "0 blocks free" line. To be more compatible with existing devices, this line should be removed.

**To Reproduce**
From BASIC, type

LOAD"$=P",[devno]
LIST

Note the last line "0 blocks free".

**Expected behavior**
No such last line, only the header and partitions.

 - Hardware type: U2+L
 - Firmware Version 3.15
 - Git Gash 68b78f50

chrisgleissner:
This PR fixes Software IEC compatibility defects affecting C64 OS, CMD DOS and sd2iec, plus crash and memory-safety defects found through randomized testing and code review. It also adds release-build diagnostics for investigating failures.

Fixes #890. Refs #875, #876, #877, #917.

**Validation:** all five host suites pass, including AddressSanitizer. Firmware passes the IEC suite, now 27 checks, and soak/stress testing with logging enabled and disabled on Ultimate 64 Elite and U2+L. Elite also passes all 50 UCI checks. The `u64`, `u64ii` and U2 targets build. The two fixes added for #917 were run red and green on an Ultimate 64 Elite; see “Issue #917 follow-up”.

**Unresolved:** the intermittent firmware hang reported in #877 has not been reproduced. These fixes address independently identified defects; they do not establish the cause of that hang.

## Scope

[doc/softiec_compatibility_spec.md](https://github.com/chrisgleissner/1541ultimate/blob/fix/iec-partition-commands/doc/softiec_compatibility_spec.md) defines requirements SI-001 to SI-153 and their provenance: the CMD HD manual, 1541 ROM, sd2iec, the C64 OS article “Gaps in Software IEC” (GAP) and the measurements reported on #917.

This PR includes defect fixes and requirements identified by #877/#890, GAP or the C64 OS boot trace attached to #877. Other features were removed or narrowed during review. The specification records those decisions, and section 18.1 describes partial implementations; none of their omissions affects commands in the boot trace.

**Review the diff against `master`:** the history includes features subsequently removed.

## Compatibility fixes

Each row has a regression test in `target/pc/linux/parse` or `target/pc/linux/iecdrive`, generally a requirement-named `Suite11` case. Tests fail when the corresponding firmware files are restored from the fix’s parent commit.

### C64 OS boot trace and GAP

| Requirement | Defect and fix |
| --- | --- |
| Binary Change Partition, #878 retest | Preserve partition byte 13 in `C` + shifted P + `CHR$(13)` instead of stripping it as a terminator. |
| SI-021, SI-022 | Increase command/name capacity from 64 to 254 bytes; reject longer input with 32. Previously, truncated commands executed and could scratch in the wrong directory. C64 OS paths reach 232 characters. |
| SI-031 | Return 31 for unknown commands, including C64 OS’s `CHR$(0)`, rather than 33. Recognised commands with invalid parameters return 30; missing names 34; illegal names 33. |
| SI-033 | `S/TEMPORARY/:*`, sent at boot, returns `01, FILES SCRATCHED,00,00` when nothing matches, rather than 62. A nonexistent path returns 71. |
| SI-105, SI-112 | Answer C64 OS’s four `M-R` probes with the requested number of zero bytes, clamped to the page, rather than a syntax error. Zero bytes match no drive signature and therefore do not advertise drive-code upload support. |
| SI-100 | Implement `U0>` + `CHR$(d)` through new `IecInterface::readdress()`. It rewrites the device-number slot without resetting the IEC processor while the command is on the bus. |
| SI-103 | `UJ` closes data channels; `U` + shifted J additionally resets every partition to its root and selects partition 1. Both return 73. Neither calls `IecDrive::reset()`, whose interface configuration would reset the IEC processor while the host still addresses it. Previously, `UJ` closed nothing and shifted-J was unknown; GAP reports bus lockups. |
| SI-045, SI-046, #890 | Partition listings now contain a header, partition entries and the two-zero-byte BASIC end marker, without `BLOCKS FREE.`. `read_dir_entry()` emits the correct ending; the header count uses `IecFileSystem::CountPartitions()`, matching sd2iec. |
| SI-134 | Apply partition-type filters `N`, `4`, `7`, `8`, `C`. Previously `$=P:*=N` listed everything. Treat `H` as non-filtering; it previously excluded every entry. |
| SI-065 | Show the listed directory’s name in the header, using the partition name only at its root. C64 OS uses this header to name favorites. |
| SI-130, SI-131 | Correct listing prefix `01 01 04 01` to `01 04 01 01`: load address `$0401`, link `$0101`. Previously `LOAD"$",8,1` loaded into the stack page. |
| SI-014, SI-015 | Interpret left arrow as parent only in the name position (`CD_`, `CD:_`). Previously every underscore became `..`, making host directories containing underscores inaccessible. |
| SI-012, SI-150 | Support wildcards in path components and remove the 16-component limit, which previously truncated deeper paths silently. |
| SI-144 | Recognise VICE/sd2iec P00/S00/U00/R00 headers. List, open, copy, rename, scratch and position using the header’s CBM name; exclude the 26-byte wrapper from file data and update the header name on rename. Name collisions return 63, or replace with `@`. UCI `GET_IECNAME` reads the header when given a full path. |

### Reporter findings on #877

| Requirement | Defect and fix |
| --- | --- |
| SI-032 | `SAVE"@:FOO*"` replaces the first match under its existing name if the type matches; otherwise returns 64, following ROM `$D8F5`. Without `@`, wildcard SAVE returns 33. Previously both returned 33. |
| SI-053 | `I` closes data channels and returns OK, as sd2iec does, instead of returning 73 without cleanup. |
| SI-071 | `N` creates/formats D64/D71/D81 images or creates DNP images, selected by extension. Previously it returned OK without doing anything. Inside a mounted image it returns 30 because formatting beneath an open filesystem is unsafe. |
| SI-072, SI-083 | Preserve disk-image/cartridge extensions. On write channels, `P` flushes, seeks and extends; insufficient space inside an image returns 72. This fixes JiffyDOS MD81 (`FOO.D81,P,W`, seek to 819199, write one byte), which produced a one-block `.D81.prg` and error 69. |
| SI-076, SI-132 | Implement `L` through new `FileSystem::file_attrib()` and `FileManager::set_attributes()`: toggle FAT read-only (`f_chmod`, `FF_USE_CHMOD 1`) or CBM type-byte bit 6. Locked entries display `<` and cannot be scratched. |
| SI-091 | Accept the documented three parameters for `B-A`/`B-F`: partition, track, sector. Previously four were required. |
| SI-041 | `G-P` returns the CMD type, partition name and actual size: image-file size or volume size through new `FileSystem::get_total()`. Previously every partition reported type 0, its root path and 16,711,680 blocks. |

### Reporter findings on #917

| Requirement | Defect and fix |
| --- | --- |
| SI-138 | The last byte of a listing now carries EOI. `read_dir_entry()` set `prefetch_max` to 31 for the 32-byte `BLOCKS FREE.` line, and to 1 for the two-byte end marker of a partition listing, so `pop_data()` took its “fetch the next entry” branch one byte before `last_byte` and stopped there. A `LOAD` never saw it, because the drive prefetches past that point inside a single talk. A program that reads a listing with `GET#` addresses the channel to talk again for every byte, so it received the stuck byte for ever and the KERNAL status never became 64. |
| SI-139 | A time stamped line is now a fixed 64 bytes in the long format and 42 in the short format, with `$01` filling the space between the end of the stamp and the zero that ends the BASIC line. The long stamp starts four characters behind the three character type instead of two, which also widens the gap between the date and the time from one space to three. The short format's columns were already correct and only its length changed. |
| SI-149 | A GEOS file inside a mounted image is now listed with the type its own directory entry carries, and a read of it over the bus now returns the file rather than the CVT container. **The read half is a regression introduced in 3.15.** `FA_OPEN_FROM_CBM` was added in 2020 by commit `8ee46e83` and passed from `IecChannel::open_file()`; the call is present at `v3.14` and `v3.14d`. Commit `76d887bd` of 23 June 2026 rewrote that open and dropped the argument, leaving the flag defined and still tested by `FileInCBM::open()` with no caller, so 3.15 puts a CVT header block in front of the file. The listed type is a separate, older defect: 3.14d listed a GEOS file as `CVT`, which is not a CBM type, and 3.15 lists it as `SEQ`; neither matches the entry's type bits. An open of a VLIR file returned nothing and logged a channel fault. |

### Other correctness fixes

| Requirement | Defect and fix |
| --- | --- |
| SI-016, SI-018 | Parse relative-record number/offset from the command as sent, including offset 13, following ROM `$E23E`. Accept record numbers without an offset or terminator. For plain-file positions, strip trailing CR/LF, including the pair emitted by `PRINT#` for logical file numbers ≥128. Previously terminators could become position bytes. A lone CR returns 31. |
| SI-036 | Return `66,ILLEGAL TRACK OR SECTOR` for out-of-geometry block access, instead of 69. |
| `B-A`, `B-F` | Allocating an allocated block now returns `65,NO BLOCK` with the next higher free track/sector, or track 0 if none exists, following the 1541-II guide and HD 9-43. Freeing a free block returns OK without changes. Both previously returned 74. Neither command is implemented by sd2iec. |
| SI-090, SI-093, SI-094 | Start `#` buffers at byte 1; retain their opening partition; use byte 0 as the `B-R`/`B-W` length. Block commands on non-`#` channels return 70, preventing `U1` from overwriting a file channel’s stream buffer. |
| SI-060, SI-063 | Return 34 for `MD` without a colon and for `RD` with a path, following HD 9-17/9-19 and sd2iec. Previously MD returned 63 and RD accepted paths. |
| SI-074 | Reject rename to an existing name with 63, including collisions with another file type. Cross-directory/partition rename still moves the file, as on `master`. |
| SI-084 | Read both one-byte and two-byte relative-file headers. A nonzero second byte or file size congruent to 1 modulo record length identifies sd2iec’s one-byte layout. Image files always use the two-byte layout. |
| SI-133 | Encode `(size mod 254) + 2` in a host file’s directory-link low byte so clients can calculate exact size. Previously every link was `$0101`. |
| SI-142 | Escape CBM-name `*`/`?` as `{2A}`/`{3F}` and match scratch/RD through directory scans, preventing unintended host wildcard deletion. |
| SI-147, SI-148 | Compare PETSCII bytes as `uint8_t`; drop trailing shifted spaces and escape interior ones. Previously signed `char` on Nios II prevented `$A0` matching, while the RISC-V path could read past the terminator. |
| Scratch, CR-8 | Match CBM names through a directory scan instead of repeatedly deleting `foo.???`, which also removed `.txt`/`.d64` files. Skip locked entries and same-name entries behind them, because deletion by name removes the first entry. |

## Crash and memory-safety fixes

These defects also exist on `master`.

| Defect and trigger | Fix | Regression evidence with fix reverted |
| --- | --- | --- |
| Negative `prefetch_more()` length after positioning beyond a record; UCI uses it as a copy length. | Offer zero bytes. | `Crash-RecordPastEnd`: negative offer detected. |
| `FileInCBM::get_size()` indexes its loop map at -1 on a damaged chain referencing an invalid track. | Stop outside disk geometry. | `Crash-DamagedChain`: ASan heap-buffer-overflow. |
| `FileInCBM::read_linear()` copies a negative length after seeking beyond EOF inside an image. | Clamp remaining bytes to zero. | `Crash-SeekPastEndImage`: ASan overflow in `memcpy`. |
| Unterminated host names in `DirectoryFAT` and 48-byte drive `FileInfo` buffers; names ≥40 characters fail during listing/open/copy/rename/replacement. | Terminate copies and use `INFO_SIZE` buffers. | `Crash-LongHostName`: ASan overflow in `strlen`. |
| Mount-cache eviction frees an image while its directory remains open and other channels access more than eight other images. | `FileManager::open_directory()` retains the filesystem until the directory is deleted. | `Crash-EvictOpenListing`: ASan use-after-free. |
| Star-heavy patterns take exponential time while holding the drive. `S:FOO*X` also incorrectly matches `FOO`. | Iterative matching, bounded by the product of pattern/name lengths; only stars may remain at end of name. | Before: 16 stars took 4.6 s; 20 exceeded 30 s. After: 40 take <1 ms. Differential test covers 171,990 inputs against the corrected recursive matcher. |
| Sector read/write/allocation traverses mounts without the FileManager lock. | Lock `fs_read_sector()`, `fs_write_sector()` and `fs_allocate_sector()`. | Not testable on the host. |
| Copy from a missing partition leaves the target open and leaks 32 KB, e.g. `C:NEW=SRC,200:X`. | Return 77 through normal cleanup. | `Crash-CopyMissingPartition`: 32 KB growth per command. |
| Reopening a UCI channel leaks its previous file, e.g. two `LOAD_SU` calls without `LOAD_EX`. | Close before reopening, matching bus OPEN. | `Crash-UciReopen`: heap growth per open. |
| `petscii_to_fat()` appends `{}` beyond a full buffer for names ending in `.PRG`. | Append only when both braces fit. | `braces bound` in `parse`. |
| Abandoned listings retain their directory; failed directory opens leave a stale pointer subsequently freed again. | Release in `close_file()` and clear the pointer before reopening. | `CR4-AbandonedListing`; `CR5-FailedDirectoryOpen`, ASan use-after-free. |
| GUI/REST/UCI/IEC tasks concurrently access partitions and channels. | Recursive drive lock across each IEC bus pass and other entry points. Closures run before taking the lock; settings changes take it. JiffyDOS transmission ends the pass after 100 ms without host consumption, avoiding an indefinite spin. | `CR6-Lock`: zero block accesses outside the lock. |

The IEC task’s 6,400-byte stack retained at least 3,352 bytes after U2+L IEC checks/10-minute soak, 3,288 after a 45-minute soak, and 3,668 on Elite after a 10-minute soak. U2+L uses the reporter’s RISC-V CPU family; Elite uses Nios II. These diagnostic measurements make IEC stack overflow an unlikely explanation for #877. They predate operation logging, whose buffers are static.

## Diagnostics and logging safety

`SoftIEC:` replaces `SOFTIEC-TRACE`.

- **Always logged:** commands leaving an error, failed opens and the first failure of a channel.
- **“Log Every Operation”, off by default:** all commands, opens and closes; data replies such as `M-R`; resolved host paths or the first 32 listing bytes; and listing endings. Available in release builds.
- Each line includes escaped input, status and post-operation device/channel/partition/directory. A trailing sequence number distinguishes lost/duplicated delivery from duplicate generation.
- `I`, `UJ` and shifted-J log closes for all channels 0-14, including unopened channels, before their command line.

Each rendered component uses a bounded, static 260-character buffer under the drive lock. Successful operations with logging disabled cost one setting read.

Three related fixes are included:

| Area | Fix |
| --- | --- |
| Settings | `IecDrive::effectuate_settings()` reconfigures the IEC processor only when device number or enable state changes. Toggling logging previously reset the processor and could drop an active transfer. |
| Syslog | `Syslog::charout()` now checks, writes and advances the buffer position within one safe section, preventing concurrent out-of-bounds writes. |
| Shared text log | `StreamTextLog::charout()` reads `offset` once and bounds-checks that value before writing, avoiding a race between repeated reads. |

`Suite11-FailureLog` verifies default failure-only output. `Suite11-OperationLog` checks line types, consecutive sequence numbers and disabling logging; it fails with logging checks forced false or sequence numbering reverted. `Suite11-OperationLogNoReconfigure` verifies zero reconfigurations for two logging changes, versus two on the parent, while device-number changes still reconfigure.

`Suite11-OperationLogBounds` covers 253-byte commands/names, 255/256-byte `M-R` replies, a 288-character directory and longer host paths. Deliberately overstating a buffer by 128 bytes triggers ASan. The two shared logging races cannot be reproduced on the host and have no tests.

**Limitations:** serial logging adds transmission latency. `printf` does not serialize whole lines, so other tasks can split/interleave them across syslog datagrams. Sequence numbers allow the test checker to account for delivery gaps and split lines.

### Coverage

Host tests run with `make` under `target/pc/linux/`:

| Suite | Coverage |
| --- | --- |
| `parse` | CBM DOS commands as sent by a Commodore. |
| `iecdrive` | FAT and D64/D71/D81 images; requirement and crash regressions. |
| `parse_unsigned`, `iecdrive_unsigned` | `-funsigned-char` variants covering CPU-dependent name handling. |
| `iecdrive_asan` | AddressSanitizer, with `ASAN_OPTIONS=detect_leaks=0`. |

Host `Suite11-Soak` randomizes boot operations, saves, relative/direct access, directory commands, abandoned listings, malformed input and GUI actions. `IECSOAK_ITERATIONS`, `IECSOAK_SEED` and `IECSOAK_LOG` control execution.

Device suites run with `./run-tests <host> -s <suite>`:

- `iec-dos-commands`: 27 checks through a C64 6502 agent using KERNAL bus operations. Three of them read the tail of a listing one byte per KERNAL transaction, which is what BASIC's `GET#` does, and check the length and the filler of a time stamped line.
- `uci-targets`: includes full-path x00 `GET_IECNAME`.
- `softiec-soak`: IEC traffic with concurrent REST/FTP load. Supports soak/stress/both and logging off/on/both/toggled every 10-20 seconds. Profiles select 10-minute stress or 4-hour soak, logging off.

**Soak mode** uses 1-6 seconds of think time and covers the #877 boot trace, filtered/40-entry listings, full readback, LOAD/SAVE secondary addresses 0/1, append, USR/x00, replacement saves, temporary files, rename/copy/lock, all four image types, DNP subdirectories, relative/direct access, clock reads, `XPWD`, `G-P`, and UCI identification/name/channel/partition operations.

**Stress mode** adds malformed commands/names, nine open channels, exhausted buffers, relative-record boundaries, eight-source copies, mount-cache eviction with open listings, invalid geometry, full-directory scratch, readdressing and resets with channels open. A continuous REST lane resets the drive every 20 seconds; three FTP lanes modify files used by the C64, including files inside its open image.

Tests fail on unresponsive firmware/IEC, data or log mismatches, excessive steady heap decline in soak (>256 bytes/iteration over the second half), or stress heap loss >64 KB after idle settling. Before declaring a hang, stress pauses load and retries REST six times.

Log verification correlates agent operations with syslog, checking fields, replies, unexpected lines, silence with logging disabled and sequence-accounted losses. `--syslog-spool` supports external collection; `--log-dump` retains operations and logs.

### Final-firmware results

Firmware `9d2bf79f`; later commits change tests/documentation only. Elite ran over JTAG; U2+L ran from flash as `u2@c64u`. Each device’s phases ran consecutively on one boot.

- All five host suites passed without ASan reports.
- IEC: **24/24 on both devices**.
- UCI: **50/50 on Elite**; not run on U2+L.
- All eight soak/stress phases below passed, including syslog correlation.

Heap values are bytes. Soak shows **warm-up / start of second half / end**; stress shows **before / after idle settling**.

| Device | Mode/logging | Minutes | Iterations | REST/FTP passes | Free heap |
| --- | --- | ---: | ---: | ---: | --- |
| Elite | Soak/on | 30 | 202 | 1,170/421 | 5,208,024 / 5,199,088 / 5,199,128 |
| U2+L | Soak/on | 30 | 155 | 1,127/402 | 5,261,968 / 5,253,632 / 5,253,552 |
| Elite | Soak/off | 10 | 72 | Not reported | 5,199,060 / 5,190,512 / 5,190,040 |
| U2+L | Soak/off | 10 | 57 | Not reported | 5,253,376 / 5,245,376 / 5,244,432 |
| Elite | Stress/off | 10 | 56 | 8,487/2,440 | 5,201,744 / 5,201,752 |
| U2+L | Stress/off | 10 | 33 | 2,757/1,462 | 5,256,096 / 5,256,160 |
| Elite | Stress/on | 10 | 67 | 9,098/714 | 5,201,748 / 5,196,524 |
| U2+L | Stress/on | 10 | 33 | 2,721/1,412 | 5,253,440 / 5,253,584 |

Logging-on phases matched 1,365/1,055 lines during Elite/U2+L soak and 642/216 during stress. Missing lines were accounted for by sequence gaps or split-line evidence; duplicate and interleaved deliveries were also detected. Logging-off phases confirmed silence for 490/392 successful soak operations and 703/389 successful stress operations.

The heaviest delivery disruption was U2+L stress/on: 112 missing complete lines, accounted for by 10 sequence gaps and 118 numbers recovered only from split lines, plus three duplicate deliveries and two interleaved lines. This passed operation correlation but demonstrates the limitations of concurrent serial/syslog output.

Soak used suite `39b80194`; stress used `1e125225`, except U2+L stress/on at `38f76a7c`. Later suite changes fixed stress cleanup, UCI steps and log checking. Every phase’s saved evidence passes the checker at `38f76a7c`.

### Issue #917 follow-up

Firmware `eeb378b1` on an Ultimate 64 Elite over JTAG. No U2+L or C64 Ultimate was available for this round.

- `iec-dos-commands`: **27/27**. On the parent build, which carries the tests but not the firmware fix, the two new areas fail: the long listing comes back 112 bytes rather than a header, whole 64-byte lines and a blocks free line, and eight single byte reads at the end of a 128-byte listing all answer status 0, returning `20 00 00 00 00 00 00 00`.
- `uci-targets`: **50/50**.
- `softiec-soak` in stress mode: 700 s, 67 iterations, 9,050 REST and 2,264 FTP passes, 21,248 bytes of heap lost after an idle settle, inside the suite's 64 KB limit.
- All five host suites pass. `Suite11-SI138-ListingEof` and `Suite11-SI139-StampedEntries` fail when the two firmware files are restored from the fix's parent commit, and each fails on its own when only its own hunks are reverted.

SI-149 was added after that round and was tested separately.

- `Suite11-SI149-GeosEntries` builds a D64 holding a GEOS PRG, a GEOS SEQ, a GEOS VLIR USR and a plain PRG, and checks the listed type of each, the bytes a read returns, and that the VLIR read returns its record block rather than nothing. The case fails on the commit that adds it and passes on the commit that adds the fix. It also fails when only the listing change is reverted, and fails differently when only the `FA_OPEN_FROM_CBM` flag is reverted: the types are then right and the read returns 256 bytes of CVT header instead of the file.
- Checked against a reference 1541 outside this tree. VICE's `c1541` lists `geos-2.0r-cenbe.d64` as `prg prg usr usr usr usr usr usr usr`, which is what this drive now lists; it listed nine `SEQ` before. `c1541` extracts `DISK COPY` from `deskpack-plus-b.d64` as 4,335 bytes and this drive now returns the same 4,335 bytes, byte for byte. That file has a load address of `$0801` and a `10 SYS(2064)` line, so it loads and runs from BASIC; it is the only such file among the 355 GEOS entries on the twenty GEOS disks that were scanned.
- All five host suites pass with the change, AddressSanitizer included.
- Measured on an Ultimate 64 Elite against three firmwares, each run from its own matching bitstream over JTAG, with `deskpack-plus-b.d64` as the Software IEC partition and the same BASIC program run from the C64 each time. The program sends `CD:DESKPACK.D64`, prints the printable bytes of `$:DISK COPY`, then prints the first eight bytes an open of `DISK COPY` returns.

  | Firmware | Listed type | First eight bytes |
  | --- | --- | --- |
  | 3.14d (`40a41caa`), FPGA 122, core 1.49 | `CVT` | `1 8 13 8 10 0 158 40` |
  | 3.15 as this PR found it (`bc3f2dc4`), FPGA 125, core 1.50 | `SEQ` | `130 0 0 68 73 83 75 32` |
  | This change (`79a53425`), FPGA 125, core 1.50 | `PRG` | `1 8 13 8 10 0 158 40` |

  `1 8 13 8 10 0 158 40` is `$01 $08 $0D $08 $0A $00 $9E $28`, the file's load address and first BASIC line, and is what `c1541` extracts. `130 0 0 68 73 83 75 32` is `$82 $00 $00` followed by `DISK `, the CVT header's copy of the directory entry, whose first two bytes a `LOAD` reads as a load address of `$0082`.
- Firmware `79a53425` on an Ultimate 64 Elite over JTAG: `iec-dos-commands` **27/27**. The device suite has no GEOS image, so this is a regression check of the existing checks against a shared change in `FileInfo` and the image file system, not coverage of SI-149 itself. No U2+L or C64 Ultimate was available for this round.

### Earlier testing and resulting corrections

| Evidence | Result or resulting fix |
| --- | --- |
| #890 device regression | Failed on U2+L `ade66d32` and pre-fix Elite firmware with an unwanted `BLOCKS FREE.` ending; passes on both after the fix. Host cases: `Suite11-SI045-PartitionDirectory`, `Suite11-SI046-PartitionCount`. |
| Host ASan soak on `04b29936` | 24 seeds × 3,000 iterations, plus eight × 3,000 with logging enabled, approximately 29,000 lines per seed. No ASan reports. All five host suites also passed. |
| Device checks across earlier builds | IEC passed on both devices; Elite UCI passed. Builds before the logging-setting fix also passed these checks with logging enabled and disabled. |
| Earlier soak/stress runs | Both devices passed 30-minute soak and repeated 10-minute stress runs. Elite additionally passed a 20-minute logging-on soak. These used earlier operation mixes and less complete log verification. |
| `B-A` regression found on `3f9336ea` | Both devices returned 74 for `B-A:0,18,0`; prompted the fix and `Suite11-BlockAllocateAnswers`. Final Elite firmware returns `65,NO BLOCK,18,02`. |
| First-session U2+L heap failure | An early 10-minute run lost 628-794 bytes/iteration over 36 iterations; most returned after deleting the test directory. Isolated drive, REST/FTP and C64-memory operations showed no leak. |
| Longer U2+L follow-up | A later 45-minute run on the same firmware passed: 158 iterations, 1,688 REST/603 FTP passes. Periodic heap samples stayed near 5.282 MB apart from one transient low sample; idle heap was 5,282,736 bytes. First-session allocation appears bounded, but its source remains unidentified. |
| Heap-check correction | Transfer buffers can depress individual samples by approximately 20 KB. Compare peak free heap in the first/last quarters instead of individual endpoints. A subsequent first-session U2+L run passed; the suite documents the remaining cold-session caveat. |

## Not included

| Item | Reason and current behaviour |
| --- | --- |
| `R-P`, `R-H`, `XH:`, `D:` | No identified requirement. Host `R-H` would rename the directory itself, unlike CMD header rename; sd2iec also rejects it. Return 30, or 31 for `D:`. |
| `EL:`, `EU:`, `EH:`, `A:` | `L` covers locking; hiding needs a listing-policy change. Return 30, or 31 for `A:`. |
| `W-0`, `W-1` | No identified software write-protect requirement; would affect every write path. Return 31. |
| `S-8`, `S-9`, `S-D` | No second drive to swap. Return 31, matching sd2iec. |
| `T-W` | System clock is set by C64 OS through UCI. Return 30, permitted by SI-120. |
| `M-W`, `M-E` | No drive-code execution; return 30 rather than falsely acknowledging an upload. |
| `V` | No image validation implemented; return 31 rather than false success. |
| `##n`, `B-P` high byte | No sd2iec-style direct 512-byte sector commands. Use standard buffers and ignore the high byte. |
| Raw directory on nonzero secondary addresses | Preserve existing `$` listing behaviour, including channel 2. |
| x00 write setting | Read existing wrappers; create plain files. No new write-format setting. |
| Stack-overflow check/memory report | Diagnostic-only changes excluded; the check would halt all images, including updaters. |

## Build size

| Measurement | Final value |
| --- | ---: |
| U2 application | 754,080 bytes |
| Headroom below its 792 KiB limit | 55.6 KiB |
| Ultimate 64 Software IEC object growth against `master` | 8,841 bytes |
| Operation-logging contribution | 862 bytes on Ultimate 64; 664 bytes on U2 |
| #917 contribution | 36 bytes on Ultimate 64, measured as the application section length before and after the fix |

## Known limitations and external observations

- **#877 remains unreproduced.** No claim is made that its intermittent hang is resolved.
- #917 also reports that a listing header prints `00 2a` where a CMD HD prints `hd 1h`, and that the low byte of a line's link pointer differs from the CMD HD's constant `01 01`. Both are kept. The header field states which device wrote the listing, and `1H` is the CMD HD's own DOS version; the link low byte is sd2iec's documented file size remainder, which SI-133 implements. Specification rows C15 and C16 carry the reasoning.
- The `C` command still copies a GEOS file out of an image as a CVT container, because `do_copy()` opens each source with a plain `FA_READ`. That keeps the interchange format on a copy to the host file system, where the receiving side needs it, and it is the one place where a copy and a read of the same file return different bytes.
- GEOS itself still cannot start from this drive, because the GEOS speeder is not implemented. SI-149 changes what a listing and a read report, not what GEOS needs to boot.
- Scratch uses 16-character CBM names. `S:GREAT GIANA SIST` can delete multiple host files sharing that displayed prefix.
- UCI copies replies after releasing the drive lock; a menu Reset in that window can change reply contents.
- `G-P` does not verify that a FAT volume remains mounted before reading its size.
- `IecChannel::push_data()` still buffers writes to a read channel at its read position, failing only when it attempts a file write at 512 bytes. This exists on `master`, was found by inspection and is unchanged.
- Stress produced 0-10 FTP upload resets per run, one U2+L PASV 425 and two agent REST resets under HTTP-slot saturation. Firmware remained responsive. FTP’s early close on target-open failure is a possible explanation, but was not verified.
- With `/Temp` full, a 254-byte REST memory write returned success without changing C64 memory. The agent now verifies mailbox writes by readback. The cause is unidentified.
- Under network saturation, a UCI command sent as individual `$DF1D` REST writes lost its final byte/push and consumed the next command as a continuation. The suite now pauses load during UCI steps. The cause is unidentified; SoftIEC logs confirmed the received bytes.


software/filesystem/file_system.h
```diff
@@ -42,11 +52,12 @@ class FileSystem
...
-    // CMD DOS calls a partition that holds its own native file system "NAT", and
-    // reports a drive emulation partition by the drive model it emulates. A
-    // directory on the host file system and a DNP image are both native.
-    virtual const char *get_partition_type(void) { return "NAT"; }
+    // CMD DOS partition type. It is reported twice, as the code in byte 0 of the
+    // G-P reply and as the three character name in the partition directory ($=P),
+    // so both come from this one answer. A partition holding a native file system
+    // is native, which covers a directory on the host file system and a DNP image;
+    // a partition rooted in a disk image reports the drive model it emulates.
+    virtual cbm_partition_type_t get_partition_type(void) { return e_partition_native; }
```
GideonZ: Reject.  File_system.h is a GENERIC base class that is not supposed to know anything about CBM.

GideonZ (changes requested):

software/filesystem/file_system.h
chrisgleissner: I will come up with another approach. Thanks Gideon.

chrisgleissner: Thanks Gideon. This is fixed now. I removed the CMD partition types and the strings from the generic FileSystem layer: the `cbm_partition_type_t` enum, `get_partition_type()`, and the CMD overrides on the D64, D71 and D81 file systems are all gone. `rg` over `software/filesystem/` finds no CMD partition code and no `NAT`, `41 `, `71 ` or `81 ` string.

Software IEC now owns the CMD mapping and derives it from the disk geometry the file system already has. `FileSystem` gained two neutral accessors in the vocabulary it already uses for `read_sector(buffer, track, sector)`:

```cpp
virtual int get_num_tracks(void) { return 0; }
virtual int get_sectors_in_track(int track) { return 0; }
```

`FileSystemCBM` answers both from its layout table; anything without track and sector addressing reports no tracks. `cbm_partition_type_of()` in `iec_channel.cc` then reads 256 sectors on track 1 as a native partition, 40 as a 1581, and 21 as a 1541 or, above 42 tracks, a 1571. `$=P` and `G-P` call the same function, so they cannot disagree.

chrisgleissner:
Hi @GideonZ,

I addressed your review comment. The CMD partition-type enum and `get_partition_type()` methods are now out of `FileSystem`, including the D64, D71 and D81 overrides.

Those methods only repeated information already present in the disk layouts: D64 and D71 both have 21 sectors on track 1 and differ by total track count; D81 has 40 sectors per track, and DNP has 256. `FileSystem` now exposes only that generic track/sector geometry. Software IEC converts it to the CMD type codes in one place, where that CMD-specific knowledge belongs. Both `$=P` and `G-P` use that mapping.

I also added trace logs behind a compile-time flag, as requested by Markus. We can keep them for future diagnosis or remove them once he has finished his analysis.

I put the PR back into Draft mode. Once Markus’s findings are in, I plan to address them together and then request review again.

Thanks,  
Christian

chrisgleissner:
Hi @GideonZ ,

thanks for re-review this PR which fixes many bugs and compatibility issues.

For full details, please read the PR descriptions above.

Thanks

GideonZ:
@chrisgleissner Wow.. this must have been a lot of work!

I am very pleased to see that there are so many fixes. On the other hand, it makes me sad that so many fixes were needed! Especially the possible crashes due to buffer overflows. That makes me a lousy programmer. This is truly humbling.

One thing I am less happy about is the support for P00, X00, etc... I deliberately avoided this and also argued with the maker of C64OS that that was a bad architectural design choice. The whole escaping of the filenames exists because of not wanting to support it. Now it feels like a monster with multiple heads. I can expect requests like: "you can read it, why not write x00 files, too?" and then we're back to ground zero.

chrisgleissner:
Thanks @Gideon for having had a look at it. I would not feel bad about the numer of issues found. This is quite normal for a piece of software with this degree of complexity.

It seems like you prefer removal of P00, X00, etc support. Could you please explain more what exactly needs changing so that I can amend the PR in order to ensure it can be merged?

chrisgleissner:
As discussed privately, I will leave the P00 / X00 read support as already added, and I am making sure that P00 can be used from the file browser consistently.

Issue #439, opened 2024-11-20
Weird issue (U2+) Cartridge on C64 Silver label.


0wwafa:
I have tried a few cartridges on the U2+ and they all seem to work on the breadbin.
Today I found out some cartidges I made, work on the breadbin but not on the silver label.
It's not about ROM differences between the two.

They are both magic desk type.

I have no idea how to debug this because in the emulator(s) they both work,

[archon.zip](https://github.com/user-attachments/files/17831797/archon.zip)

Attached are the two cartridges.
Any help apreciated (if it's my fault) if it's not.. then it's on you :D

radius75:
It looks a bit like #335

UffeJakobsen:
Have you adjusted PHI1/PHI2 settings for that Silver Label ?

radius75:
This is probably not a problem with PHI, I have the same problem with Archon as OP.

0wwafa:
> Have you tried adjusted PHI1/PHI2 settings for that Silver Label ?

Yep. I tried every possible setting of PHI1/PHI2.
Probably it's because my cart loader is the fastest possible.. but I don't yet know if it's the cart loader that causes the problem or the U2+.

0wwafa:
> It looks a bit like #335 on my U2+

Hmm.. I did not test it with 3.9... I am on 3.11 now.

radius75:
Archon 64KB crt doesn't work on fw3.9 for me either. So it probably never worked.

mrdudz:
The first thing to do here would be replacing the archon payload by something like a checksummed self checking binary (then you can see if the cartridge startup is actually broken) - like this its a pretty useless test case

radius75:
```Creating CRT type from info: WorkinginU2onSilverLabel.crt
Object level 1 returned 1.
Action set was: Run Cart
Cartridge Load.. WorkinginU2onSilverLabel.crt
CRT Hardware type: 19 Header OK. Now reading chip packets starting from 000040.
Reading chip data bank 0 ($8000) with size $2000 to 0x00F00000.
Reading chip data bank 1 ($8000) with size $2000 to 0x00F04000.
Reading chip data bank 2 ($8000) with size $2000 to 0x00F08000.
Reading chip data bank 3 ($8000) with size $2000 to 0x00F0C000.
Reading chip data bank 4 ($8000) with size $2000 to 0x00F10000.
Reading chip data bank 5 ($8000) with size $2000 to 0x00F14000.
[...]

PR #901, opened 2026-09-14
Fix COMAL 80 cartridge: blank screen on 4 MB products (#899) and garbled graphics text

### LINKED ISSUE #49: FW 3.12a - Comal CRT producing garbled text in Graphic Screen
Testing some new software written in Comal 80 2.01 I found issues in displaying text in a graphic screen, only when a CRT image of Comal is attached in the U64 or the UII+:

Image

This issue does **not** come up when a **physical** original Comal cartridge or even a Kung Fu Flash **using the exactly same CRT** image is being used. In that case, the screen is displayed like this: 

Image

Also, the very same CRT image is working in VICE.

So, from my perspective, it looks as if the Comal CRT image is somehow badly handled by the U64 and UII+ cores.

### LINKED ISSUE #899: Comal80 CRT does not work anymore with 3.15/3.15a
**Describe the bug**
With firmware 3.14d (and lower) CRT files of Comal80 used to work with all Ultimate products. With firmware 3.15/3.15a, running comal80.crt just brings up an empty screen.

**To Reproduce**
1. Upload comal80.crt to Ultimate
2. Run CRT file

**Expected behavior**
The COMAL 80 system should start running.

 - Hardware type: 1541U2, U2+(?), U2+L, U64, U64E2
 - Firmware Version 3.15(a)
 - Git Hash dddd29b2

**Additional context**
A physical Comal 80 cartridge (or e.g. a Kung Fu Flash) does work with fw 3.15(a).

chrisgleissner:
Fixes #899. Fixes GideonZ/ultimate_releases#49.

### Blank screen on products with 4 MB of cartridge memory (#899)

COMAL 80 selects its banks by writing $80 to $83 to $DE00, so bit 7 is always set. Since the large cartridge support, `c_ocean_16K` takes all eight bits as the bank number and 4 MB products decode 22 address bits, so bank $80 is at offset 2 MB. `C64_CRT::auto_mirror` still repeated the image only up to 64 banks (1 MB), and the rest of the region is $FF.

`auto_mirror` now repeats the image up to `max_cart / bank_multiplier` banks. On 1 MB products that is still 64. This is the rule described in #439 and on #793: repeat the image until the whole cartridge region is filled.

### Garbled text on the graphics screen (ultimate_releases#49)

COMAL 80 writes $41 to $DE00 and $01 for every glyph byte that `plottext` reads. On the Commodore (black) cartridge bit 6 drives GAME and EXROM, so the cartridge switches off and the character ROM appears at $D000. `c_ocean_16K` held GAME and EXROM low and ignored bit 6, so the glyphs were read from RAM.

In `c_ocean_16K`, bit 6 of a $DE00 write now drives GAME and EXROM for variants 0 and 1, which the firmware selects only for COMAL 80. This matches the black cartridge schematic, VICE and Kung Fu Flash. The older grey cartridge uses bit 6 for GAME only and bit 5 for EXROM. VICE marks it with CRT subtype 1, and the firmware now selects variant 2 for that subtype, which keeps the previous behaviour.

The CI build synthesises the FPGA images of the Ultimate II family from `fpga/`, so their update files include this change. The Ultimate 64 and C64 Ultimate images are prebuilt, so they need a core rebuild to include it.

### Register placement for the cartridge-off flag

The cartridge-off flag from bit 6 has to persist between $DE00 writes, since `io_wdata` only holds a valid value during the write pulse itself. `game_n` and `exrom_n` for `c_ocean_16K` read a dedicated register, `ocean_off`, rather than `mode_bits(0)`, which about fifteen other cart-type branches in `all_carts_v5.vhd` already read and write for their own state.

Reading and writing `mode_bits(0)` for this flag passed a VHDL analysis and an open-source ECP5 synthesis mapping, but failed Lattice Diamond's real place-and-route for the U2+L target three times in CI, each a genuine 18 to 27 minute synthesis attempt rather than a cached skip. Reverting only that one hunk, with everything else in this branch (including #905 and #908, already in master) unchanged, built clean in 29 minutes. `mode_bits(0)` is already read by roughly fifteen other branches; the extra read this flag added pushed that shared net past whatever margin Diamond's router needs on U2+L's ECP5.

`ocean_off` costs one flip-flop that nothing else reads. An open-source synthesis mapping of `all_carts_v5` (GHDL and yosys, ECP5 cell library) shows the wide-mux cell count on the `mode_bits` fan-out back at master's 69, with one new, isolated flip-flop (146 to 147) in its place.

### Effect on other cartridge types

Checked against master d115446b for every CRT type the firmware accepts (comparison scripts not in this PR):

- **Loader:** 3,516 CRTs (generated for every type, plus 114 real images) load through master's and this branch's `c64_crt.cc` with the same result, I/O flags and first 1 MB of cartridge memory. Only CRT 21 with subtype 1 gets a different cartridge type.
- **FPGA:** a GHDL simulation of master's and this branch's `all_carts_v5` side by side, for all 32 cartridge logics and 8 variants, shows output differences only for black COMAL 80 (`c_ocean_16K` variants 0 and 1). Grey COMAL 80 and Ocean (#908) are unchanged.
- **Memory above 1 MB:** only COMAL 80, Ocean 16K, EasyFlash, Megabyter and, without #905, C64GS address it. EasyFlash decodes six bank bits, so the copies match the hardware.
- **Variant gating:** `c64_crt.cc` only pairs `CART_TYPE_OCEAN_16K` with variant 0 or 1 for COMAL (from `CART_COMAL80`), or with variant 2 or 3 for the grey cartridge and a 256K Ocean cartridge (from `CART_COMAL80` and `CART_OCEAN_8K` respectively). The write to `ocean_off` is gated on `not variant(1)`; variant 2 and variant 3 both have that bit set, so the gate stays closed regardless of what the ROM writes to bit 6. A 256K Ocean cartridge (Robocop 2, Shadow of the Beast, Space Gun) has no use for bit 6 in its own bank address either: 256 KB is 32 banks, five bits.

Visible effects:

- An EasyFlash Maniac Mansion writes $78 to $7B to $DE00. On the Ultimate 64 Elite it stays on a blue screen with master and matches VICE with this branch.
- Megabyter images of fewer than 256 banks read copies instead of $FF at unused banks above 1 MB and above the image size rounded up to a power of two.
- COMAL 80 CRTs with subtype 0 get the black cartridge logic, as in VICE and Kung Fu Flash.

### Test

`comal80-cartridge` builds four-bank type 21 CRTs in code. A routine in RAM stores $EE in the RAM under $9FF0, selects banks 3 to 0 through $DE00, and copies the bytes at $9FF0 and $BFF0 after each selection to screen RAM. Each bank has its own marker bytes. The suite has three checks:

1. Writes of $83 to $80 read the markers of banks 3 to 0 (#899).
2. Writes of $C3 to $C0 read $EE at $9FF0 and no marker at $BFF0, so the cartridge is off (#49).
3. The same writes with CRT subtype 1 (grey cartridge) read the markers of banks 3 to 0.

Check 2 needs the FPGA change. It is gated on a new `tests/lib/machine.py` entry, `comal80-cartridge-off-bit`, which lists the Ultimate 64, Ultimate II+ and C64 Ultimate kinds, so it reports SKIP there; `run-tests --assume-fix comal80-cartridge-off-bit` runs it. The Ultimate II+ kind stays listed until the machines the suites run against have a firmware build that includes this change. Check 3 also depends on the #899 fix, because $C0 addresses offset 3 MB.

`software/io/c64/tests` is a host test of the CRT loader, run in CI as "Test CRT Loader". For every CRT type the firmware accepts, it checks the cartridge logic and I/O restrictions the loader selects, where each chip lands in cartridge memory, and that the image repeats up to the end of 1 MB and 4 MB of cartridge memory.

`easyflash-cartridge` builds a 64-bank type 32 CRT in code. A routine in RAM selects every bank with $00, $40, $80 and $C0 in the upper bits and copies each bank's marker to screen RAM. Each of the four rows has to read the markers of banks 0 to 63.

### Verification

Each fix was shown red by removing only that fix, and green with this branch. All firmware was built from a8ee9cd2 (this branch); a red build adds one local commit that reverts the `software/` part of one commit. Every run used `run-tests --no-retry -s comal80-cartridge`, with `--assume-fix comal80-cartridge-off-bit` unless the table says SKIP.

Ultimate 64 Elite, firmware deployed over JTAG. The cores are the flashed core 1.50 and two evaluation cores built locally from the same core sources and timing constraints, with `fpga/` from master (c7464da2) or from this branch (05ef8b50). The two local cores differ only in `all_carts_v5.vhd`.

| core | firmware | check 1 | check 2 | check 3 |
|---|---|---|---|---|
| flashed 1.50 | a8ee9cd2 | OK | FAIL: $9FF0 reads 30 31 32 33 | OK |
| flashed 1.50 | a8ee9cd2, 162ff0c0 reverted | FAIL: all $FF | SKIP | FAIL: all $FF |
| local, master `fpga/` | a8ee9cd2 | OK | FAIL: $9FF0 reads 30 31 32 33 | OK |
| local, this branch | a8ee9cd2 | OK | OK | OK |
| local, this branch | a8ee9cd2, firmware part of 05ef8b50 reverted | OK | OK | FAIL: $9FF0 reads ee, $BFF0 reads bf |

Ultimate II+L in a C64 Ultimate (`u2@c64u`). Each package was built from the listed firmware with a CI-built ECP5 bitstream and installed through the updater. The bitstream comes from the CI build of this PR (run 34845594620) or of master d115446b (run 34942624081).

| bitstream | firmware | check 1 | check 2 | check 3 |
|---|---|---|---|---|
| this PR | a8ee9cd2 | OK | OK | OK |
| this PR | a8ee9cd2, 162ff0c0 reverted | FAIL: all $FF | SKIP | FAIL: all $FF |
| master | a8ee9cd2 | OK | FAIL: $9FF0 reads 30 31 32 33 | OK |
| this PR | a8ee9cd2, firmware part of 05ef8b50 reverted | OK | OK | FAIL: $9FF0 reads ee, $BFF0 reads bf |

Without `--assume-fix`, check 2 reports SKIP and the suite passes with a8ee9cd2 on both machines.

Host test: all 525 checks pass on 61db8fdc, in the CI image as well. Reverting only 162ff0c0 fails the six 4 MB repetition checks, for example `EasyFlash, 64 banks, 4 MB: 192 of 256 banks do not repeat the image, first bank 64`. Reverting only the firmware part of 05ef8b50 fails the two grey COMAL 80 rows (`cartridge type $09, want $49` and `$29, want $49`).

`easyflash-cartridge` on the Ultimate 64 Elite, core 1.50, suite from 61db8fdc: a8ee9cd2 passes all five checks. a8ee9cd2 with 162ff0c0 reverted passes the $00 to $3F row and fails the $40, $80 and $C0 rows, which read $FF for every bank.

Real images on the Ultimate 64 Elite (not part of this PR), firmware a8ee9cd2 unless stated:

- COMAL 80 rev 2.01 (Commodore, subtype 0): blank screen with 162ff0c0 reverted; the banner `$$$ Commodore-64 COMAL 80 rev 2.01 $$$` on core 1.50 and on the local core with this branch's `fpga/`.
- COMAL 80 ApS (subtype 1): starts with `Copyright (C) 1984 by UniComal ApS` on both of those cores.
- `plottext(10,100,"hello world")` on the graphics screen: 386 pixels differ from VICE on core 1.50, none on the local core with this branch's `fpga/`.

The register move in `ocean_off` was verified the same way. CI's U2+L build for this branch's latest commit is a genuine, non-cached synthesis (25m47s, run 35005898066). `u2@c64u`, flashed with that build's package (`git_commit_hash` confirmed matching), passes `comal80-cartridge` (all three checks, `--assume-fix comal80-cartridge-off-bit` included), `ocean-cartridge` (32-bank, 256K, banks reached at both $8000 and $A000) and `easyflash-cartridge` (all five checks).

### Limitations

- The FPGA images in `external/` are not rebuilt, so the #49 change reaches Ultimate 64 and C64 Ultimate devices with the next core build.
- Not measured: 1541 Ultimate-II, Ultimate II+, Ultimate 64 Elite II and C64 Ultimate as a target.
- The grey cartridge keeps its previous behaviour rather than getting its own GAME and EXROM logic.


chrisgleissner:
Merge check, run locally because the self-hosted CI runner is offline.

This PR was merged together with #901, #902, #905 and #908 onto master b7c49f10:

- CI host steps (API, UCI and HTTP target tests, lint, OpenAPI, D81): pass. Two observability cases fail in a worktree container because git is not reachable there, and they fail the same way on master.
- u64, u64ii and u2 build, u2 including FPGA synthesis with `all_carts_v5.vhd` (timing met). U2 app space: pass.
- Ultimate 64 Elite, `run-tests --profile standard` on that firmware: 46 of 48 pass on core 1.4F. `ultimax-cartridge` fails with the known 1.4F result and passes on core 1.50. `rest-api-coverage` failed only because the worktree build had no commit hash, and passes on a build that has one. `comal80-cartridge` and `ocean-cartridge` pass; `c64gs-cartridge` skips as designed.

Not built here: the U2+ and U2+L FPGAs (no Cyclone IV or Diamond toolchain).

#901, #905 and #908 each add a suite next to `ultimax-cartridge` in `run-tests` (a small conflict) and each bump the counts in `tests/README.md`. After the second one is merged, regenerate the matrix with `tools/docs/update_test_docs.py`.

chrisgleissner:
Confirmed to work by @radius75 (thanks).

PR #902, opened 2026-09-14
Accept CRT files with data after the last chip packet (#900)

### LINKED ISSUE #900: A Pig Quest (Trex Release) don't work any more
**Describe the bug**
The Trex-Release of "A Pig Quest" doesn't work any more with Firmware 3.15 and only shows a black screen.

**To Reproduce**
1. Upload attached CRT-File to 1541U II/II+/II+L
2. Select CRT-file.

**Expected behavior**
The game should start normally

 - Hardware type: 1541U II (old version)
 - Firmware Version 3.15

**Additional context**
Also I'm still waiting for any reply regarding the 1541U II upgrade. It's not nice to keep someone waiting so long. @GideonZ Please get in touch or send me your correct email address so I know whether I need to buy new hardware or not. Thanks!

chrisgleissner:
Fixes #900.

Since bfbe3d1a (first released in 3.11), `C64_CRT::read_chip_packet` rejects a CRT that has any bytes after its last chip packet with "Error detected in file format". Up to 3.10 those bytes ended the chip list, and VICE does the same, because its loaders stop at the first header they cannot read. Kung Fu Flash rejects such a file. The copy of A Pig Quest in #900 is the csdb release plus 96 bytes of 0x1A that pad it to a 128-byte block, most likely from a BBS or XMODEM transfer. The reporter confirmed that the padding is the only difference.

A header without the `CHIP` signature now ends the chip list when at least one chip packet has already been read. When the first packet after the file header is not a chip packet, the file is still rejected, so data that is not a cartridge image keeps the error from bfbe3d1a. `load_crt` empties `chip_chunks` before parsing, so a non-empty list means a valid packet came first.

### Test

`ultimax_cartridge_test.py` gets two checks next to the existing Jupiter Lander check:

- `jupiter_lander.crt` with the same 96 bytes appended has to reach the same golden frame.
- the CRT header followed only by those bytes has to be refused with HTTP 415.

Both starts now also require HTTP 200 from `runners:run_crt`. No new cartridge image is added.

A separate commit lets the polling loops skip a capture that could not assemble a complete frame. The host socket drops datagrams while the suite is blocked in a REST call, and on the bench the existing check failed that way in 3 of 10 runs on master. Other capture errors still fail.

### Verification

Ultimate 64 Elite with `external/u64.sof` from master (core 1.50) loaded over JTAG, because the golden frame needs the Ultimax fix from #882:

- master firmware: the padded check failed with HTTP 415 in 2 of 2 runs, the other checks passed.
- this branch: all three checks passed in 6 of 6 runs.

On the stock core 1.4F, the padded A Pig Quest copy from #900 was refused on master through `runners:run_crt` and "Run Cart", and starts the Trex menu with this branch through both. The csdb copy still starts.


chrisgleissner:
Merge check, run locally because the self-hosted CI runner is offline.

This PR was merged together with #901, #902, #905 and #908 onto master b7c49f10:

- CI host steps (API, UCI and HTTP target tests, lint, OpenAPI, D81): pass. Two observability cases fail in a worktree container because git is not reachable there, and they fail the same way on master.
- u64, u64ii and u2 build, u2 including FPGA synthesis with `all_carts_v5.vhd` (timing met). U2 app space: pass.
- Ultimate 64 Elite, `run-tests --profile standard` on that firmware: 46 of 48 pass on core 1.4F. `ultimax-cartridge` fails with the known 1.4F result and passes on core 1.50. `rest-api-coverage` failed only because the worktree build had no commit hash, and passes on a build that has one. `comal80-cartridge` and `ocean-cartridge` pass; `c64gs-cartridge` skips as designed.

Not built here: the U2+ and U2+L FPGAs (no Cyclone IV or Diamond toolchain).

#901, #905 and #908 each add a suite next to `ultimax-cartridge` in `run-tests` (a small conflict) and each bump the counts in `tests/README.md`. After the second one is merged, regenerate the matrix with `tools/docs/update_test_docs.py`.

PR #906, opened 2026-09-14
Fix direct REU DMA timing on U64

### LINKED ISSUE #895: [U64 rev1.2 fw3.15a] Expand by Bonzai REU Demo does not run (flawlessly)
One initial test of the REU Demo "Expand by Bonzai" fails. 

1. Download "Expand by Bonzai - Disk1.D64" from https://csdb.dk/release/?id=192886 
2. Run it on a U64 rev1.2 board with fw 3.15(a)
3. The (last) test fails. Demo freezes with a light blue screen. 

You can get past this test by setting the U64 Turbo to e.g. 4 MHz before starting the demo.
Then this message shows up:
Image
Looks like the "REU DMA to font data" routine got timing issues with this firmware.
You can switch back to 1 MHz after that but the graphics in parts of the demo (e.g. on disk 2 the font in the last part) using this routine will look teared.

Tested on:
 - U64 rev1.2 board
 - Firmware 3.15a
 
REM if the "Bus Operation Mode" is set to "Compatibility" the test fails regardless the "higher MHz cheat" - result: light blue screen.

barryw:
Closes #895

@chrisgleissner @GideonZ

Requested labels: `bug`, `3.15`.

The U64/U64E2 no-delay path asserted `DMA_n` immediately but left the actual
REU request in the delay state until the next phi2 tick. That split timing
breaks the raster-timed REU-to-font transfers used by Expand. Dispatch the
request immediately on direct-DMA machines while retaining the initial phi2
delay for cartridge REUs.

Regression coverage:

- New `reu_dma_timing_tc` was red on `master` at 46 ns: `direct DMA waited for
  phi2 after the command executed`; it is green with this change.
- The same RTL test verifies that cartridge REUs retain their phi2 delay.
- Existing `reu_turbo_test` passes at 1 MHz and 48 MHz on the U64 Elite
  (21.1x measured speedup).
- New manual `reu-bonzai` E2E downloads the exact published D64 by pinned
  SHA-256, loads it through drive 8, and requires three stable title frames.
  It reproduces the failure on the currently flashed 3.15/core 1.50 image.
- The maintained Beast `u64` build passes.

Hardware validation note: this public tree packages the prebuilt
`external/u64.sof`; it does not contain the U64 FPGA project that rebuilds
that file. The Bonzai E2E therefore needs a U64/U64E2 FPGA image rebuilt from
this VHDL change before it can provide the hardware green result.


chrisgleissner:
@barryw I *really* like your VHD test case. Nice one.

barryw:
@GideonZ We have cycle-level red/green evidence for this fix, but the public repository only packages external/u64.sof; target/u64/nios2/updater/getbuild.sh expects the FPGA project at ../../../../ult64/target/u64_a4/work/u64.sof.

The same GHDL test passes immediately before 30278344 and fails at that commit, then passes again with 0139866. It checks both sides of the behavior: direct-DMA U64 dispatches on the DF01 execute write, while cartridge REUs retain the phi2 delay.

Could you point us to the ult64 project/build procedure, or build a U64/U64E2 image from this PR for us? We have an on-device Bonzai E2E ready and will flash and run it before marking the PR ready.

chrisgleissner:
@barryw Gideon is currently working on fixing the CI builds, but it could take a while. I don't think there is a point in creating ad-hoc builds since the time is better spent on getting CI back up.

barryw:
@chrisgleissner do you have access to the source for u64.sof? I'm happy to do all of this myself, but for whatever reason I cant find this code. I wasn't sure if this contained some secret sauce that @GideonZ didn't want publicly available.

chrisgleissner:
Yes, this is a private repo.

GideonZ:
There is no quick fix on this. It needs an architectural change and I am not going to let others fiddle with it.

PR #912, opened 2026-09-15
[DO NOT MERGE] Bisect test for #901 CI failure

chrisgleissner:
Throwaway diagnostic PR, will close once done.

This is #901's branch with only the `c_ocean_16K` game_n/exrom_n hunk reverted back to master's version, to test in isolation whether that specific hunk is what's causing #901's U2+L build to fail on real CI (3/3 real syntheses of #901's code have failed the same way: `diamondc` exits nonzero in the "Build U2+L including FPGA" step, no other change differs from a known-good master build).

If this build goes green, it confirms the hunk. If it still fails, something else is going on.


chrisgleissner:
Bisection confirmed: reverting the ocean_16K hunk fixed the U2+L build. Root cause identified and fixed forward in #901 (dedicated register instead of the shared mode_bits bit). Closing this throwaway.

PR #927, opened 2026-09-20
Keep the GMod2 EEPROM change until it is saved

Jondalar:
## Problem

`C64::get_eeprom_dirty()` clears the hardware flag every time it is read, and
System Info reads it to show "Dirty (Needs save!)".

So on a GMod2 cartridge: look at System Info, then run Save Cartridge. The save
finds a clean EEPROM, skips fetching it from the hardware, and writes the
contents from the time the cartridge was loaded. The player's save is gone, and
nothing says so.

The same function also cleared the flag when it had read it as clean, so an
EEPROM write landing between the read and the clear was dropped as well.

## Change

A file-static latch holds a change that was seen until `get_eeprom_data()` or
`set_eeprom_data()` actually takes the contents. Reading the flag is now without
effect for callers. The hardware flag is cleared only when it was seen set, and
`get_eeprom_data()` clears it before copying, so a write during the copy is
caught next time instead of being lost.

## Validation

Builds for U64, U64-II and U2+ (application and recovery app). One file, no
interface change.

Not reproduced on hardware. The path is read from the code: `system_info.cc`
reads the flag, and `save_crt()` in `c64_crt.cc` then finds it clean and keeps
the buffer it loaded.


chrisgleissner:
Hi @Jondalar ,

Thanks for raising the three PRs today: One new feature and two bug fixes.

The way I understand it, you raised the two bugs as independent PRs so that they can be merged in before we land the new feature PR, as the latter includes the bug fixes as well.

A bit more context on the two bug tickets would be useful, especially given nobody reported appears to have reported the bugs. We need to figure out whether they are safe to land in the 3.15b release.

Also, could you please add tests to this PR here? We generally try to use a more test-driven approach and have recently added a large and growing set of E2E tests. These are useful where writing a unit test would require extensive mocking.

Btw, your reverse engineering MCP server https://github.com/Jondalar/C64ReverseEngineeringMCP is very impressive. I will be sure to check it out. 

Not sure if it's of interest to you, but I also wrote an MCP server a while back: https://github.com/chrisgleissner/c64bridge I see how the two MCP servers could complement each other, as they have different use cases.

Thanks

Jondalar:
Hi @chrisgleissner 

understood, I will see that we build tests in your format. I test with my new emulator (was sick of flashing all the time) usually but repeatable BDD tests are a good thing. 

Thanks for mentioning my workbench, use it a lot and it delivers great results  (see Wasteland Enhanced Edition e.g.).

Jondalar:
Tests added, and the fix is simpler than it was.

**Test.** `tests/e2e/io/c64/gmod2_eeprom_dirty_test.py`, registered as `gmod2-eeprom-dirty`. It builds a GMod2 cartridge (CRT type 60) in the test rather than shipping a ROM, and drives the EEPROM from RAM with the runtime in `software/6502`, because the chip select in bit 6 of $DE00 switches the ROM off with it. The harness asks for the write over `machine:writemem`, so the EEPROM is read once before one as well:

```
[01] the cartridge starts and its EEPROM routine is ready ... OK
[02] the machine has an EEPROM ... OK (System Info reports 'Clean')
[03] the EEPROM starts out unchanged ... OK
[04] System Info reports the EEPROM as changed after the game saved ... OK
[05] opening System Info again still reports the change ... OK
```

Check 5 is the regression: before the fix the second reading says `Clean`, and a Save Cartridge taken then writes the EEPROM as it was loaded. The suite reports SKIP where the FPGA build has no EEPROM.

**Fix.** Now four lines: `get_eeprom_dirty()` simply does not clear. The flag is cleared where the contents are taken, in `get_eeprom_data()` and `set_eeprom_data()`, which it already was. The latch I had is gone, and with it the file-static state.

**On landing it in 3.15b.** The trigger is narrow, which is why nobody reported it: a GMod2 game, a save in the game, System Info opened, then Save Cartridge. When it hits, the save is lost silently. The blast radius of the change is one accessor with two callers (`system_info.cc:88`, `c64_crt.cc:786` and `:847`), both of which only want to know. Every cartridge load calls `set_eeprom_data()` (`c64_crt.cc:351` for a file's EEPROM chunk, `:940` for a cleared one), so no state carries between cartridges.

Run against the firmware of this branch in an emulator, twice, clean. Not run on hardware.

Jondalar:
Two things carried over from the review on #929.

`6817777d` applies the same corrections to this suite: the menu button is pressed once through `menu.toggle_menu` rather than in a loop, since it is a toggle, and the polling goes through `wait.wait_until` so it paces itself from the shared values and lands in the interaction log.

And #929 no longer carries its own version of this fix. It had a file-static latch holding a change that had been seen; it now does what this PR does, which is simply not to clear the flag on read. The two branches agree about the function, so whichever order they land in, there is nothing to reconcile.

Run in an emulator, 5 checks OK.

Issue #52, opened 2018-12-16
suppert Easyflash save feature


Flexmaen:
The Easyflash specification allows to save things on the cartridge.

This is a feature more and more cartridge releases use. Since the 1541 does not support it yet, **saving a game is impossible for recent releases**.
So, it would be great to support the full Easyflash specitications.

For example, some releases that make use of the save feature:
https://csdb.dk/release/?id=156948
https://csdb.dk/release/?id=98674

Grrrolf:
From the [FAQ](https://1541u-documentation.readthedocs.io/en/latest/faq.html#does-the-u2-and-u2-support-easyflash-cartridge-images)

> Does the U2 and U2+ support Easyflash Cartridge images?
> Yes, both carts support EF. Since firmware 3.2 there is also support for saving to the EF cartridge image:
> 
> Technology Preview: EAPI Support for Easyflash, incl. but not limited to writing support. Please note that the changed crt file is NOT saved automatically. You can save it by “F5”, “Save Easyflash crt” in the filebrowser (and optionally renaming the saved file).

So, while running an EF cart, in essence, enter the Ultimate menu, press F5 (contextual menu) and then choose 'Save EasyFalsh crt'. Make sure to rename it afterwards.

While this is might not the type of solution you'd expect, it is the only (feasible) solution right now and it beats having no way of saving.

Flexmaen:
Oh, didn't know that. Sounds good, but i completely froze the Ultimate on my C128 (using Firmware 3.2 (10F)).
Should this work like a cartridge freeze?

Flexmaen:
This also happens on my SX64 and does not seem to be a C128 related problem.
Btw, I have the Ultimate II (without +).

markusC64:
USB Stick or Micro SD?

Flexmaen:
microSD

michas0602:
Agree to the starter. Even though actual way it "beats having no way of saving", saving it correct in the same file would make it 'Ultimate'.

Flexmaen:
But like I said: Even the workaround doesn't work for me yet.

marty41:
Did you wait some time, as saving the EF3 image does take a minute or even longer on the U2. This was tested working on U2 with a 250407 board and 3.2 and 3.3 firmware.

Flexmaen:
[...]

PR #929, opened 2026-09-20
Save a changed cartridge back to the file it came from

Jondalar:
Refs #52. Sits on top of #927 and #928 and carries both, so this diff loses
those two as soon as they land.

## Problem

A game that saves to its cartridge only changes the image in memory. The CRT
file keeps the old state until Save Cartridge is run by hand, a file name is
typed and an overwrite confirmed. Forget that and the save is gone the next
time the cartridge is started.

## Change

The image is hashed when it is loaded and after every save. When the menu
opens, it is hashed again; a difference means the C64 wrote to it.

A new setting, **Save Changed Cartridge** (Off / Ask / Auto), then writes it
back to the file it came from. Ask is the default and puts up "Cartridge
changed. Save it to NAME?"; Auto saves with a progress bar; Off does nothing.
Saying no sets a new baseline, so the question returns only after the next
change.

Writing goes `NAME.tmp`, then the file as first loaded is kept once as
`NAME.bak`, then the rename. Cartridges loaded from `/Temp` (REST uploads) or
`/Flash`, or from a folder that is not writable, get one notice pointing at
Save Cartridge instead.

The hash covers exactly the bytes `save_crt()` writes, EEPROM included, and
uses no multiply, since the U64 Nios II has none. It runs without stopping the
C64 — a torn read means at worst one extra save, or a change found at the next
menu open. The save itself runs stopped.

`UserInterface` gets one static hook called from `run_once()`, and the C64 side
registers it. Everything that needs the user interface sits in `c64_subsys.cc`,
so `c64_crt.cc` stays free of it and keeps building in the host test.

## Validation

`software/io/c64/tests`: 550 checks, up from 529. The new ones cover the source
a cartridge was loaded from, that one flipped bit anywhere in the banks the
file describes is seen, and that mirrored copies above them are not counted.

Builds for U64, U64-II and U2+ (application and recovery app); `app_space`
passes with 19.4 % free on U64 and 33.6 % on U64E2_100T.

On a U64: an EasyFlash game that saves through EAPI, mode Ask. The prompt came
up after the game had saved, the file was written back and `NAME.crt.bak`
appeared. In an emulator the same run end to end, and the saved file reloaded
into a byte-identical image — the game then offers "Continue" instead of a
fresh start.

Hashing 1 MiB takes 120 ms on the U64-II. Only EasyFlash, and the EEPROM of a
GMod2, are hashed at all, since in `all_carts_v5.vhd` those are the only images
the C64 can change. Every menu entry logs the figure.

## Not covered

- Loading a cartridge over REST replaces a changed image without asking.
- The telnet menu has no hook.
- A save in progress on the C64 when the menu opens is caught half-written, as
  it would be by pulling the plug.
- Types other than EasyFlash and GMod2 are hashed too, on the assumption that
  only EasyFlash writes into the ROM image. That is what `all_carts_v5.vhd`
  does; I cannot check the U64 core.
- On hardware only the EasyFlash case above. An emulator ran the same case
  through the overlay as well as through the C64 menu. GMod2 after System Info
  and the Auto mode are untested.

Design notes, the mode table and the full test matrix are in
`doc/CRT_AutoSave.md`. Happy to trim that down or move it out if you would
rather keep `doc/` thin.


chrisgleissner:
@Jondalar Thanks for this very useful improvement!

Since this change extends the user interface, would you be able to add an E2E test? The existing E2E tests provide several examples to build on.

The test could drive the menu through injected keystrokes, interacting with the device in much the same way as a user. Depending on what is most appropriate and efficient, assertions can be made against the menu state, RAM contents, or the video and audio streams. The E2E library provides abstractions for each of these.

For example, @GideonZ recently added a Jupiter Lander cartridge test that verifies the post-launch state against a reference image captured from the video stream.

An E2E test would provide durable regression coverage at the external-behaviour level. We also plan to run these tests automatically in CI once the required hardware setup is complete. For now, they can be run with:

```bash
./run-tests <host>
```

All available options are listed by:

```bash
./run-tests -h
```

If possible, please also include red/green results in the PR:

* The test fails against the unpatched firmware.
* The same test passes against the patched firmware.

This confirms that the test exercises the intended behaviour and gives reviewers greater confidence in the change.

doc/crt_auto_save.md
```diff
@@ -0,0 +1,201 @@
+# Cartridge Auto-Save
```
chrisgleissner, line 1: Could we rename this to `crt_auto_save.md` please, as per name conventions for recently added files?

GideonZ:
The AI generated report is wrong.
The U64 *does* have I-cache and a multiplier.

Jondalar:
Thank you for the correction. You're right — I read the BSP as a description of the hardware, which it isn't. Corrected: the hash avoids a multiply because `public.mk` passes `-mno-hw-mul`, not because the chip lacks one.

Jondalar:
E2E test added, and the document is renamed to `doc/crt_auto_save.md`.

`tests/e2e/io/c64/cartridge_autosave_test.py`, registered as `cartridge-autosave`. It builds an EasyFlash cartridge in the test rather than shipping a ROM: bank 0 is entered in Ultimax mode, copies a routine to $0800 and runs it there, and the routine arms one write with $65 to $DE09 and programs one byte into the ROM window whenever the harness asks for one — the way the firmware's own EAPI does it. One running cartridge therefore serves every scenario:

```
[01] a writable medium to save the cartridge to ... OK (/SD)
[02] the menu offers to save the changed cartridge, naming its file ... OK
[03] Yes writes the cartridge back to its own file ... OK
[04] the file as it was first loaded is kept once, as NAME.crt.bak ... OK
[05] the saved cartridge carries its own EAPI, not the Ultimate's ... OK
[06] an unchanged cartridge is not offered again ... OK
[07] no prompt appears while the feature is Off ... OK
[08] the file on disk is left as it was ... OK
[09] the cartridge is saved with no prompt at all ... OK
[10] the backup still holds the file as it was first loaded ... OK
[11] a cartridge uploaded over REST runs from a source it cannot write ... OK
[12] the notice is given once, not at every menu open ... OK
```

Check 5 is the half of this change that a file on disk can show: the firmware patches 768 bytes of EAPI into cartridge memory while a cartridge runs, and the saved file has to carry the cartridge's own bytes rather than those. The writing scenarios need a writable medium and report SKIP without one; the /Temp scenario runs either way.

Run against the firmware of this branch in an emulator, twice, clean, with no UI-state warning. Not run on hardware.

The document also no longer claims anything about the U64's CPU that only the BSP says.

chrisgleissner:
Thanks for having added an E2E test @Jondalar  - excellent. I will test this change against my devices as soon as they are available - currently in use for the Soft IEC PR.

In the meantime, I have done a code review using Opus 5 High. Please read the results below. Thanks for addressing them.

software/io/c64/c64_subsys.cc
```diff
@@ -75,6 +76,132 @@ static void initBootCart(void *object, void *param)
...
+    return NULL;
+}
+
+// Hooked into UserInterface::run_once(). From the overlay the C64 is still running: the hash is
+// taken without stopping it, the save runs stopped. A torn read only means an extra save now or a
+// change found at the next menu open.
+static void crt_autosave_menu_hook(UserInterface *ui)
+{
+    int mode = C64 :: getMachine()->get_cfg_value(CFG_C64_CRT_AUTOSAVE);
+    if ((mode == CRT_AUTOSAVE_OFF) || !C64_CRT :: is_valid()) {
```
chrisgleissner, line 174: A reboot makes this prompt appear for a cartridge the C64 never wrote to, and Yes or Auto then saves a corrupted file.

`Reboot C64` (menu, REST `machine:reboot`, `control_target.cc`) and `Reboot (Clr Mem)` call `start_cartridge(NULL)`, which calls `set_cartridge(NULL)` (`c64.cc:1244`). With the default empty `Cartridge` setting, that function does `cart_mem[5] = 0` to disarm the previous ROM and loads nothing. It does not call `C64_CRT::clear_crt()`, so `is_valid()` stays true and `source` still names the file.

`cart_mem[5]` is byte 5 of the first chip chunk: `$8005`, which is inside `CBM80` for a normal cartridge and bank 0 ROML for an EasyFlash. The next time the menu opens, `current_hash()` differs from the baseline. Ask offers "Cartridge changed. Save it to NAME?"; Yes, or Auto without asking, writes the file back with that byte zeroed. A normal 8K/16K cartridge saved this way no longer autostarts. The `.bak` is only written on the first save, so a later occurrence has no backup of the state before it.

Steps: load any CRT from USB/SD with the browser, choose Reboot C64, open the menu.

The E2E suite's own teardown does this (see the comment on line 493).

A possible fix: when `set_cartridge(NULL)` does not load a cartridge, call `C64_CRT::clear_crt()` so the image, source and baseline are dropped together with the running cartridge. This also stops Save Cartridge from being offered for a cartridge that is no longer running.

tests/e2e/io/c64/cartridge_autosave_test.py
```diff
@@ -0,0 +1,512 @@
...
+            device.wait_running()
+            scenario_ask_and_save(device, path)
+            scenario_off(device, path)
+            scenario_auto(device, path)
+        scenario_not_writable(device)
+    finally:
+        teardown_step("restore Save Changed Cartridge", device.restore_mode)
+        teardown_step("remove the test cartridge", device.clean_up)
+        # A reboot removes the cartridge; a reset would boot straight back into it.
+        teardown_step("restore the configured cartridge", device.api.machine.reboot)
```
chrisgleissner, line 525: This reboot leaves the cartridge loaded from `/Temp` registered with the firmware, with `$8005` zeroed (see the comment on `c64_subsys.cc:167`). The mode is restored to `Ask` just before this line. The next suite that opens the menu therefore gets "Cartridge changed, but its file cannot be written back", which it does not expect.

A check after the reboot (open the menu, `expect_quiet()`) would fail on this branch as it stands. That makes it the red half of a red/green test for the firmware fix, and it stops this suite from leaving a popup for the next one.

The teardown also reboots without closing the menu first. If a check fails while the Save prompt or the notice is on screen, `machine:reboot` reaches `client->release_host()` while the UI task is still inside the hook's popup. Please answer the prompt with `n` and close the menu before the reboot. Do not answer it with RETURN: that picks Yes, the first button, and saves the cartridge.

tests/e2e/io/c64/cartridge_autosave_test.py
```diff
@@ -0,0 +1,512 @@
...
+        self.api.machine.writemem(COMMAND, bytes([value]))
+        deadline = time.monotonic() + WRITE_TIMEOUT_SECONDS
+        while self.api.machine.readmem(ACK, 1)[0] == before:
+            if time.monotonic() >= deadline:
+                raise Failure("the cartridge routine did not acknowledge a write")
+            time.sleep(0.1)
+
+    # -- the menu --------------------------------------------------------
+
+    def set_menu(self, want_open: bool) -> None:
```
chrisgleissner, line 292: This helper presses the menu button again every `MENU_TOGGLE_SETTLE_SECONDS` (0.25 s) until the state matches. The button is a toggle. If the menu takes longer than 0.25 s to open (freeze, `appear()`, then this PR's hash, which the description measures at 120 ms for 1 MiB), the second press closes it again. The loop can then alternate between the two states until the timeout.

`menu.toggle_menu()` in `tests/e2e/lib/menu.py` covers this case: it presses once and polls the state, and its docstring explains why the press must not be repeated. `ButtonMenuOpener` and the `ui_backend` helpers are built on it. Please use one of those.

`answer()` below has the same pattern with keys. It repeats `y` or `RETURN` until the popup text disappears from `menu_screen`. If the firmware has taken the first key but not yet redrawn, the next key goes to the browser behind the popup. `api.close_menu_from_anywhere()` documents that a stray RETURN there opens the entry under the cursor. It also hides a lost key instead of reporting it. Once `wait_for_text()` has seen the popup, the UI owns input, so one key followed by a wait for the text to disappear is enough. If a key is really lost at that point, that is a defect worth reporting rather than retrying around.

tests/e2e/io/c64/cartridge_autosave_test.py
```diff
@@ -0,0 +1,512 @@
...
+ROUTINE_TIMEOUT_SECONDS = 15.0
+WRITE_TIMEOUT_SECONDS = 10.0
+PROMPT_TIMEOUT_SECONDS = 30.0       # the hash runs before the prompt appears
+SAVE_TIMEOUT_SECONDS = 120.0        # a megabyte onto an SD card
+# No prompt is a negative: long enough that a device about to raise one has.
+QUIET_SECONDS = 8.0
+
+# Roots to save to, in the order a device is likely to have them. /Temp and
+# /Flash are deliberately absent: the firmware refuses both (c64_subsys.cc).
+CANDIDATE_ROOTS = ("/SD", "/Usb0", "/Usb1", "/Usb2", "/Usb3")
```
chrisgleissner: Other duplication of existing library code in this file:

- `ftp_lib.usb_volumes()` in `tests/lib/ftp.py` lists the USB volumes the device actually serves. The device names a volume after its port, so a stick is not always on port 0 to 3, and the device lists them as `USB0`, not `Usb0`. The docstring of that helper describes a suite that skipped itself because of a hard-coded root like this one.
- `chip()` is a copy of `chip()` in `easyflash_cartridge_test.py`. `bootstrap.directory("e2e", "io", "c64")` is the documented way to import from another suite.
- The polling loops in `set_menu`, `wait_for_text`, `answer`, `wait_running`, `program` and `retrieve_saved` repeat what `wait.wait_until` / `menu.wait_until` do. Those use the shared pacing values and write to the interaction log.

chrisgleissner (commented):
Review of the firmware change and the E2E suite at 6d9cf44b. Nothing was run on hardware for this review; the findings come from reading the code.

Findings tied to a line of code are inline. The effect on the rest of the E2E run, and how to run the cartridge suites, are in a separate PR comment.

chrisgleissner:
Two more points from the review that concern the E2E run as a whole, and how to run the cartridge suites. Findings tied to a line of code are in the inline comments.

**The reboot defect also affects the existing cartridge suites.** `ultimax-cartridge`, `comal80-cartridge`, `easyflash-cartridge`, `ocean-cartridge` and `c64gs-cartridge` all upload a CRT to `/Temp` and end with `machine:reboot`, on the assumption that a reboot removes the cartridge. With this branch the reboot leaves the CRT registered with `$8005` zeroed (see the inline comment on `c64_subsys.cc:167`). After each suite, `run-tests` runs the `ui_state.py` `verify` gate, which opens the menu. The menu then shows the "cannot be written back" notice, so the runner should report each of these suites as WARN ("left the device UI outside the documented state"), although none of them changed.

**The UI-state gate can answer the Save prompt with Yes.** If the registered CRT came from a writable medium, the popup is the Yes/No prompt, not the Ok notice. `ui_state.py` only recognises a popup whose one button is Ok. Its fallback, `close_menu_from_anywhere(confirm_key="return")`, presses RETURN, and RETURN picks Yes, the first button. The test harness would then write the cartridge back to disk. Fixing the reboot defect removes the trigger that the review found.

**Running the cartridge suites.** Please run these on hardware, with an SD card or a USB stick in the device. Without one, check [01] of `cartridge-autosave` reports SKIP and the Ask, Off and Auto scenarios do not run. This runs every cartridge suite in registry order, where `cartridge-autosave` sits between the others:

```bash
./run-tests <host> --no-retry -o results/pr929 \
  -s ultimax-cartridge -s comal80-cartridge -s easyflash-cartridge \
  -s cartridge-autosave -s ocean-cartridge -s c64gs-cartridge
```

- `--no-retry` runs each suite once. By default a failing suite is run again, and a pass on the second attempt is recorded as a flake, which can hide an intermittent problem.
- `-o` keeps the report, with the screen captures and the interaction logs, so the results can be attached here.
- Leave out `--no-restore-settings`, so the runner puts the settings back at the end.
- `./run-tests -l` lists the suite names, in case the registry changes.

What to expect: every suite reports OK, not WARN. A WARN after any of the cartridge suites is the reboot defect. For the red half, run the same command against a build without the feature, and paste both summaries here. Before merging, a full `./run-tests <host> --profile standard` would also show whether anything outside the cartridge suites is affected.

software/io/c64/c64_subsys.cc
```diff
@@ -75,6 +76,132 @@ static void initBootCart(void *object, void *param)
...
+        fm->delete_file(temp.c_str());
+        C64_CRT :: set_baseline(previous);
+        printf("[CRT] save failed: %s\n", target.c_str());
+        return "Writing the cartridge failed.";
+    }
+
+    File *existing = NULL;
+    if (fm->fopen(backup.c_str(), FA_READ, &existing) == FR_OK) {
+        fm->fclose(existing);
+        res = fm->delete_file(target.c_str());
```
chrisgleissner, line 141: If the final rename below fails, a later save attempt deletes the only copy of the saved state on disk.

1. First failure: `NAME.crt.bak` exists, so `NAME.crt` is deleted here. The rename of `NAME.crt.tmp` to `NAME.crt` then fails, and the new state is left in `NAME.crt.tmp`, as the message says.
2. The baseline was restored, so the next menu open saves again. `fopen(temp, FA_CREATE_ALWAYS)` truncates `NAME.crt.tmp` and writes it again, then `delete_file(target)` returns `FR_NO_FILE`, because `NAME.crt` is already gone.
3. That error path deletes `NAME.crt.tmp` and returns. The disk now holds only `NAME.crt.bak`, the file as first loaded. Every later attempt ends at the same step, so this repeats at every menu open.

Treating `FR_NO_FILE` from this delete as success would let the retry complete.

software/io/c64/c64_crt.cc
```diff
@@ -682,10 +677,70 @@ SubsysResultCode_e C64_CRT::load_crt(const char *path, const char *filename, car
...
+}
+
+const char *C64_CRT::get_source(void)
+{
+    return get_instance()->source.c_str();
+}
+
+// Over exactly the bytes save_crt() writes. Each step is a bijection in h for a fixed word, so two
+// images that differ in one word always hash differently. No multiply: the U64 Nios II has none.
+uint32_t C64_CRT::content_hash(void)
```
chrisgleissner, line 726: This hashes the image of every cartridge type on every menu open, although only EasyFlash has its ROM image written by the C64. In `all_carts_v5.vhd`, EasyFlash is the only type with `allow_write` set while `addr_map` is ROM. The RAM of Action Replay, Retro Replay, KCS, SS5, Pagefox and the C128 cartridges is at `g_ram_base`, outside the chip chunks this function reads. GMod2 is the other type that changes, and only through its EEPROM chunk. That covers the U2 family. The U64 core is not in the repository, so it cannot be checked the same way.

Hashing only EasyFlash and GMod2 would remove the hash time on every menu open for the other types, including multi-megabyte images. It would also mean that the firmware changing another type's image, as the reboot does (see the comment on `c64_subsys.cc:167`), is never reported as a save made by the C64.

software/io/c64/c64_crt.cc
```diff
@@ -682,10 +677,70 @@ SubsysResultCode_e C64_CRT::load_crt(const char *path, const char *filename, car
...
+    crt->source += filename;
+}
+
+const char *C64_CRT::get_source(void)
+{
+    return get_instance()->source.c_str();
+}
+
+// Over exactly the bytes save_crt() writes. Each step is a bijection in h for a fixed word, so two
+// images that differ in one word always hash differently. No multiply: the U64 Nios II has none.
```
chrisgleissner: The U64 Nios II has a multiplier and an instruction cache (see the comment from @GideonZ). The build passes `-mno-hw-mul`, which is the actual reason for avoiding a multiply. The doc was corrected in one place, but the same claim is still here, in `doc/crt_auto_save.md:206` ("no cache and no multiplier") and in the Change and Validation sections of the PR description.

software/io/c64/c64_crt.h
```diff
@@ -83,6 +83,13 @@ class C64_CRT
     int          eeprom_size;
     uint8_t     *original_eapi;
 
+    // Full pathname the image was loaded from or last saved to; empty if it came from no file.
+    mstring      source;
+
+    // Hash of the image as loaded or last saved; a different hash means the C64 changed it.
+    uint32_t     baseline_hash;
+    bool         baseline_valid;
```
chrisgleissner: `baseline_valid` is set in `initialize()` and `read_crt()` but never read. Either the hook should check it, or it can be removed.

Jondalar:
From my Opus 5: 

Thank you for the review — every finding held up, and the first one was a defect I had put in.

| Finding | Commit | What changed |
|---|---|---|
| `c64_subsys.cc:167` reboot leaves the cartridge registered | `1c3799ed` | `set_cartridge(NULL)` now calls `C64_CRT::clear_crt()` when it loads nothing, so the image, its source and its baseline go with the cartridge. The branch for a cartridge with its own memory already did this. |
| `c64_subsys.cc:141` a retry deletes the only copy | `e2bd0e20` | `FR_NO_FILE` from that delete is the state the next rename puts right, so it no longer counts as a failure. |
| `c64_crt.cc:705` every type is hashed | `2b0bb3db` | EasyFlash whole, GMod2 over its 2 KiB EEPROM chunk, nothing else. |
| `c64_crt.h:91` `baseline_valid` unread | `c8c93b3d` | Removed. `read_crt()` is private to `load_crt()`, which calls `initialize()` again on failure, so `is_valid()` is already false there. |
| `c64_crt.cc:704` the multiplier claim | `b5abe604` | The comment, `doc/crt_auto_save.md` and the PR description now say the build passes `-mno-hw-mul`. |
| `:298` toggle and key loops | `5d2b66e5` | `menu.toggle_menu()`; the popup gets one key and a wait. |
| `:98` duplicated library code | `da2df7ec` | `ftp.usb_volumes()`, `chip()` imported through `bootstrap.directory()`, polling through `wait.wait_until()`. |
| `:493` teardown | `e2bdbfcc` | A scenario loads the cartridge from its medium, reboots and requires silence — the red half. The teardown answers the prompt with `n`, never RETURN, and closes the menu before the reboot. |

Your two run-level points had the same cause as the first finding and are gone with it.

On `usb_volumes()`: my list would have skipped the three writing scenarios on any device whose stick is not on one of the first four ports, and reported OK for a run that tested almost nothing. That is the finding I am most grateful for.

I have no device here I can use for 3.15 right now, so only my emulator. That is where the command you gave was run. Everything reported OK there:

```
ultimax_cartridge_test:    OK (3 checks)
comal80_cartridge_test:    OK (3 checks)
easyflash_cartridge_test:  OK (5 checks)
cartridge_autosave_test:   OK (13 checks)
ocean_cartridge_test:      OK (1 check)
c64gs_cartridge_test:      SKIP (needs the c64gs-bank-from-address fix)
     OK   e2e overlay    60.0s (6 ok)
```

The red half, the same command against a build without this branch:

```
[11] a reboot is not reported as a save made by the C64 ... FAIL
     (the menu offered 'Cartridge changed. Save it to' when it should not have)
```

GideonZ:
I am wondering how this works. Does this mean that the software is constantly calculating hashes over the memory? Isn't that extremely wasteful?

Jondalar:
> I am wondering how this works. Does this mean that the software is constantly calculating hashes over the memory? Isn't that extremely wasteful?

- Hash is  calculated when the CRT is loaded and stored
- When you return to the level of the Ultimate.App, the hook fires to calculate hash again and compare against stored value

If the hash has changed -> the Cartridge in memory was updated, so it should be saved (only EasyFlash CRTs and GMod2-EEPRom for now, until you support PTV Megabyter and C64MegaCart as well).

According to my U64 syslog it takes ~80ms to hash  for 1MiB EasyFlash:

[CRT] hash E66219BE, baseline A68B886F, 80 ms
[CRT] saved /USB1/wasteland_ef_v1.0.1.crt

